PR #70, opened 2020-01-21
fix: make expired tokens a warning 

Fox32:
If token verification fails because the token is expired, we are in a situation were we can't do anything about this in operations. Instead of an error, we should log a warning here to reduce noise in production.


pvlieres:
> If token verification fails, we are in a situation were we can't do anything about this in operations. Either the token is to old or wrong. Instead of an error, we should log a warning here to reduce noise in production. On the other hand, this could also point to misconfiguration of the service.

Either way, this should not be logged as an error, I think. So: I'd approve it.

PR #114, opened 2020-02-10
fix(server-swagger): support reading the embedded fields from an arra…

dhenneke:
…y-property

This change is useful for endpoints that return a wrapper objects e.g. for paging but still want to have the embed parameter documentation generated correctly.


sda-commons-server-swagger/src/main/java/org/sdase/commons/optional/server/swagger/parameter/embed/EmbedParameterModifier.java
```diff
@@ -88,9 +98,30 @@ private Model getModelDefinition(Map<String, Model> definitions, String name) {
     return null;
   }
 
+  private Model getResolvedModelDefinition(Map<String, Model> definitions, Model definition) {
+    // In a search endpoint, the resource that supports embedding might be in a list as property.
+    if (!definition.getProperties().containsKey(EMBEDDED_PROPERTY)) {
```
JoergSiebahn: Like to see this condition outside of this method to clarify the modelDefinition is retained if it already has embedded properties.

PR #116, opened 2020-02-11
Dead Letter Handling for Kafka

ahoeing:


christopher-cudennec:
This PR will be closed because we recommend a different strategy for handling errors:
* instead of introducing a new retry and deadletter topic we recommend resetting the offset of your consumer group, thus replaying the messages of your topic for the specific consumer group

A common retry and deadletter topic has the big disadvantage that it would replay messages for all consumer groups and not only the one that could not process the message. As it also brings more complexity because it is not Kafka standard, SDA and Edith.care decided to not merge the changes and not follow this approach. 

Please note that it is very important that your consumers can handle the same message more than once. I.e. you should implement your consumers idempotent.

PR #123, opened 2020-02-15
WIP fix(server-morphia): exclude the ping command from tracing

Fox32:
The ping command is execuded from time to time and is not related to existing traces. Therefore we should exclude it from tracing.

### DIFF
--- a/sda-commons-server-morphia/src/test/java/org/sdase/commons/server/morphia/MorphiaBundleTracingIT.java
+++ b/sda-commons-server-morphia/src/test/java/org/sdase/commons/server/morphia/MorphiaBundleTracingIT.java
@@ -1,7 +1,10 @@
 package org.sdase.commons.server.morphia;
 
 import static org.assertj.core.api.Assertions.assertThat;
+import static org.awaitility.Awaitility.await;
+import static org.sdase.commons.server.mongo.testing.MongoDbRule.Builder.DEFAULT_DATABASE;
 
+import com.mongodb.BasicDBObject;
 import dev.morphia.Datastore;
 import io.dropwizard.Application;
 import io.dropwizard.setup.Bootstrap;
@@ -11,6 +14,7 @@ import io.opentracing.mock.MockSpan;
 import io.opentracing.mock.MockTracer;
 import io.opentracing.tag.Tags;
 import java.util.Set;
+import java.util.concurrent.TimeUnit;
 import org.junit.ClassRule;
 import org.junit.Test;
 import org.junit.rules.RuleChain;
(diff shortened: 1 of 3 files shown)


sda-commons-server-morphia/src/test/java/org/sdase/commons/server/morphia/MorphiaBundleTracingIT.java
```diff
@@ -39,24 +43,43 @@
...
-    assertThat(tracer.finishedSpans())
-        .extracting(MockSpan::operationName)
-        .contains("createIndexes", "insert");
-    assertThat(
-            tracer.finishedSpans().stream()
-                .map(s -> s.tags().keySet())
-                .flatMap(Set::stream)
-                .noneMatch(Tags.DB_STATEMENT::equals))
-        .isTrue();
+    await()
```
JoergSiebahn: Can we add a timeout either to the waiting or to the test itself?

sda-commons-server-morphia/src/test/java/org/sdase/commons/server/morphia/MorphiaBundleTracingIT.java
```diff
@@ -39,24 +43,43 @@
...
-  private Datastore getDatastore() {
-    return ((MorphiaTestApp) DW.getRule().getApplication()).getMorphiaBundle().datastore();
+  @Test
+  public void shouldExcludePingCommand() {
+    BasicDBObject ping = new BasicDBObject("ping", "1");
+    getMorphiaBundle().mongoClient().getDatabase(DEFAULT_DATABASE).runCommand(ping);
+
+    MockTracer tracer = getMockTracer();
+    await()
+        .pollDelay(1, TimeUnit.SECONDS)
```
JoergSiebahn: Should have a timeout as well.

sda-commons-server-morphia/src/test/java/org/sdase/commons/server/morphia/MorphiaBundleTracingIT.java
Fox32: What do you mean with timeout? By default, await comes with a timeout of 10 seconds, I don't think that we should specify it in every test class and every usage of await.

JoergSiebahn: Good to know. It‘s hard to see that in a review on GitHub without an IDE.

Fox32:
@dhenneke hat a good comment, this change is not needed, as I already added a fix some days ago that also catches the spans for the ping command.

We need to check first if the change is really necessary.

sda-commons-server-morphia/src/test/java/org/sdase/commons/server/morphia/MorphiaBundleTracingIT.java
```diff
@@ -39,24 +43,43 @@
...
+        .untilAsserted(
+            () ->
+                assertThat(tracer.finishedSpans())
+                    .extracting(MockSpan::operationName)
+                    .doesNotContain("ping"));
```
dhenneke: (In case we need this change:) Can we also have an assertion for something that we expect to be there after the ping command? I'm not sure if the test is otherwise really testing what you want since it is hard to test that something _didn't_ happen with awaitability.

PR #168, opened 2020-03-12
Feature/maven central upload

ChristianSpiess:


build.gradle
```diff
@@ -148,6 +236,17 @@ configure(subprojects.findAll {
             rename ".*", "pom.xml"
         }
     }
+    // some shortcut tasks
+    task publishSdaIntern {
```
Fox32: We can keep this for now, but should remove it soon after we verified that the maven central upload works. We don't the additional upload.

build.gradle
```diff
@@ -123,21 +140,92 @@ configure(subprojects.findAll {it.name != 'sda-commons-bom' && it.name != 'sda-c
...
+                    url = 'https://github.com/SDA-SE/sda-dropwizard-commons'
+                    licenses {
+                        license {
+                            name = 'MIT License'
+                            url = 'https://raw.githubusercontent.com/SDA-SE/sda-dropwizard-commons/master/LICENSE'
+                        }
+                    }
+                    organization {
+                        name = 'SDA SE Open Industry Solutions'
+                        url = 'https://sdase.org'
```
Fox32: Do we have a constraint to use the same domain as our group, or can we use sda.se directly here?

Fox32:
How do we continue with the pipeline for signing?

build.gradle
```diff
@@ -123,21 +140,92 @@ configure(subprojects.findAll {it.name != 'sda-commons-bom' && it.name != 'sda-c
  * - example modules and dependency-check should not be published at all
  */
 List<String> unpublishedModules = Arrays.asList("sda-commons-bom", "sda-commons-dependencies", "sda-commons-dependency-check")
-configure(subprojects.findAll {
-    !unpublishedModules.contains(it.name) && !it.name.endsWith("-example")
-}) {
+configure(subprojects.findAll { !unpublishedModules.contains(it.name) && !it.name.endsWith("-example") }) {
```
dhenneke: The modules `sda-commons-bom` and `sda-commons-dependencies` are not uploaded. So the version in teh snapshot repository at sonatype are not usable yet.

They are configured in their own gradle files: https://github.com/SDA-SE/sda-dropwizard-commons/blob/master/sda-commons-bom/build.gradle

We can maybe check if we can also configure them in the main gradle file.

PR #171, opened 2020-03-13
feat: add support for JSON logging

Fox32:
You can activate JSON logging for normal log messages and request log via the ENABLE_JSON_LOGGING environment variable.

### DIFF
--- a/sda-commons-server-dropwizard/README.md
+++ b/sda-commons-server-dropwizard/README.md
@@ -77,3 +77,7 @@ public void initialize(Bootstrap<Configuration> bootstrap) {
     bootstrap.addBundle(DefaultLoggingConfigurationBundle.builder().build());
 }
 ```
+
+To enable [JSON logging](https://www.dropwizard.io/en/latest/manual/core.html#logging), set the environment variable `ENABLE_JSON_LOGGING` to `"true"`.
+We recommend JSON logging in production as they are better parsable by tools.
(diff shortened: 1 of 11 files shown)


dhenneke:
I tested it and I like it. Do we care about the sonar issues?

JoergSiebahn:
> I tested it and I like it. Do we care about the sonar issues?

I think we can close the "extract constant" issues as Won't Fix in Sonarcloud.
We may wan't to fix the code duplication and give the method a more generic name like `createConsoleAppender`.

JoergSiebahn:
Maybe we can involve @wurstbrot to resolve the security hotspots of sda-dropwizard-commons explicitly?

Fox32:
> I think we can close the "extract constant" issues as Won't Fix in Sonarcloud.
> We may wan't to fix the code duplication and give the method a more generic name like `createConsoleAppender`.

After looking at it, I would close all three. Refactoring the duplicated code would just bloat it up.

wurstbrot:
Thank you for considering security in your daily work!

The issue mentioned is that by using env. variables, an attacker might inject something. Therefore, env variables should be compared against a whitelist or at least sanitized.

The following condition is used:
_"true".equals(System.getenv("ENABLE_JSON_LOGGING")_
This is good from a security standpoint because a whitelist (with value "true") is defined and compared against.

PR #194, opened 2020-03-25
feat(server-auth-testing): add more flexible request matching to the OPA rule

dhenneke:
This change adds support for checking whether the JWT was correctly sent to the OPA. In the old implementation one could use `onRequest(method, path)` to configure an expected request. Now there is a builder approach `onRequest().withMethod(method).withPath(path).withJwt(jwt)` that also allows to check for explicit JWT values. This is also possible for verifying requests: `verify(1, onRequest().withMethod(method).withPath(path).withJwt(jwt))`. There was no functional change in the existing methods.

The existing `onRequest(method, path)` approach was also marked as deprecated since the new approach gives much more flexibility. Futhermore, it should be advised to always check if the JWT is if correctly forwarded to the OPA.

### DIFF
--- a/sda-commons-server-auth-testing/src/main/java/org/sdase/commons/server/opa/testing/OpaRule.java
+++ b/sda-commons-server-auth-testing/src/main/java/org/sdase/commons/server/opa/testing/OpaRule.java
@@ -29,8 +29,26 @@ public class OpaRule extends ExternalResource {
 
   private final WireMockClassRule wire = new WireMockClassRule(wireMockConfig().dynamicPort());
 
+  /**
+   * Create a builder to match a request.
+   *
+   * @return the builder
+   */
+  public static RequestMethodBuilder onRequest() {
+    return new StubBuilder();
+  }
+
+  /**
+   * Create a builder to match a request for a method and path
+   *
+   * @param method the method of the request
+   * @param path the path of the request
+   * @return the builder
+   * @deprecated use {@code onRequest().withMethod(method).withPath(path)} instead
+   */
+  @Deprecated
   public static AllowBuilder onRequest(String method, String path) {
-    return new StubBuilder(method, path);
+    return onRequest().withHttpMethod(method).withPath(path);
   }
 
   private static String getJson(String[] paths) {
@@ -42,14 +60,7 @@ public class OpaRule extends ExternalResource {
   }
 
   public static AllowBuilder onAnyRequest() {
-    return new StubBuilder();
-  }
-
-  private static String[] splitPath(String path) {
-    if (path.startsWith("/")) {
-      path = path.substring(1);
-    }
-    return path.split("/");
+    return new StubBuilder().onAnyRequest();
   }
 
   @Override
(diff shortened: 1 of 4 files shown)


sda-commons-server-auth-testing/src/main/java/org/sdase/commons/server/opa/testing/OpaRule.java
```diff
@@ -75,18 +86,59 @@ protected void after() {
...
+  public interface RequestMethodBuilder {
+    RequestPathBuilder withMethod(String method);
+  }
+
+  public interface RequestPathBuilder {
+    RequestExtraBuilder withPath(String path);
+  }
+
+  public interface RequestExtraBuilder extends AllowBuilder {
+    RequestExtraBuilder withJwt(String jwt);
```
JoergSiebahn: Why does it return the `RequestExtraBuilder`? This seems to allow multiple calls to `withJwt(…)` in the builder. Does that makes sense?

sda-commons-server-auth-testing/src/main/java/org/sdase/commons/server/opa/testing/OpaRule.java
```diff
@@ -75,18 +86,59 @@ protected void after() {
...
   public void mock(BuildBuilder builder) {
     builder.build(wire);
   }
 
-  private RequestPatternBuilder matchForVerify(String httpMethod, String[] paths) {
-    return RequestPatternBuilder.newRequestPattern(RequestMethod.POST, urlMatching("/.*"))
-        .withRequestBody(matchingJsonPath("$.input.httpMethod", equalTo(httpMethod)))
-        .withRequestBody(matchingJsonPath("$.input.path", equalToJson(getJson(paths))));
+  public interface RequestMethodBuilder {
+    RequestPathBuilder withMethod(String method);
```
JoergSiebahn: Maybe `httpMethod` is a more specific term?

JoergSiebahn:
As far as I can see, `path` in the tests contains always a single path segment. Do you have any tests that verify the implementation of the `splitPath`-Method? There may be another edge case with trailing slash. Would it be fine to split and filter for not blank? Maybe you can use the algorithm of the `OpaAuthFilter`. Following the code, it uses `decodePath` of Jerseys `UriComponent` which is static.

sda-commons-server-auth-testing/src/main/java/org/sdase/commons/server/opa/testing/OpaRule.java
dhenneke: No, that would not make sense. In future, we could also add the possibility to check for e.g. request headers. We could then extend this interface.

sda-commons-server-auth-testing/src/main/java/org/sdase/commons/server/opa/testing/OpaRule.java
dhenneke: I'll change it

dhenneke:
> As far as I can see, `path` in the tests contains always a single path segment. Do you have any tests that verify the implementation of the `splitPath`-Method? There may be another edge case with trailing slash. Would it be fine to split and filter for not blank?

Fair point. I'll check the tests for the examples.

PR #200, opened 2020-03-27
feat(server-auth): make the input extensible for different use cases

dhenneke:
This PR adds the possibility to add extensions for the `OPAInput` and moves the implementation of the headers to an extension that is enabled by default. It also adds an option to disable this default extension.

### DIFF
--- a/sda-commons-server-auth/README.md
+++ b/sda-commons-server-auth/README.md
@@ -156,7 +156,7 @@ The OPA bundle requests the policy decision providing the following inputs
  * HTTP path as Array
  * HTTP method as String
  * validated JWT (if available) 
- * all request headers 
+ * all request headers (can be disabled in the [`OpaBundle`](./src/main/java/org/sdase/commons/server/opa/OpaBundle.java) builder)
 
 _Remark to HTTP request headers:_  
 The bundle normalizes  header names to lower case to simplify handling in OPA since HTTP specification defines header names as case insensitive.
--- a/sda-commons-server-auth/README.md
+++ b/sda-commons-server-auth/README.md
@@ -157,12 +157,15 @@ The OPA bundle requests the policy decision providing the following inputs
  * HTTP method as String
  * validated JWT (if available) 
  * all request headers 
- 
+
 _Remark to HTTP request headers:_  
 The bundle normalizes  header names to lower case to simplify handling in OPA since HTTP specification defines header names as case insensitive.
 Multivalued headers are not normalized with respect to the representation as list or single string with separator char.
 They are forwarded as parsed by the framework. 
-  
+
+_Security note:_
+Please be aware that while a service might only consider one value of a specific header, the OPA is able to authorize on a array of those.
+Consider this in your policy when you want to make sure that you authorize on the same value that a service might use to evaluate the output.
 
 These inputs can be accessed inside a policy `.rego`-file in this way:
 ```
@@ -186,7 +189,7 @@ allow {
     token.payload.claim
 
     # allow if a request header 'HttpRequestHeaderName' has a certain value 
-    input.headers["httprequestheadername"][_] == "certain-value" 
+    input.headers["httprequestheadername"][_] == "certain-value"
 }
 ```
 
@@ -306,3 +309,52 @@ opa:
 
 [`sda-commons-server-auth-testing`](../sda-commons-server-auth-testing/README.md) provides support for testing
 applications with authentication.
+
+## Input Extensions
+
+The Bundle offers the option to register extensions that send custom data to the Open Policy Agent to be accessed during policy execution.
+Such extensions implement the [`OpaInputExtension`](.src/main/java/org/sdase/commons/server/opa/extension/OpaInputExtension.java) interface and are registered in a dedicated namespace.
+The extension is called in the `OpaAuthFilter` and is able to access the current `RequestContext` to for example extract additional data from the request.
+_Overriding existing input properties (`path`, `jwt`, `httpMethod`) is not possible._
+
+This extension option should only be used if the normal authorization via constraints is not powerful enough.
+In general, a custom extension should not be necessary for most use cases.
+
+_Remark on accessing the request body in an input extension:_
+The extension gets the `ContainerRequestContext` as input to access information about the request.
+Please be aware to not access the request entity since this might break your service.
+A test that shows the erroneous behavior can be found in [`OpaBundleBodyInputExtensionTest.java`](./src/test/java/org/sdase/commons/server/opa/OpaBundleBodyInputExtensionTest.java)
+
+_Security note:_
+When creating new extensions, be aware that you might access properties of a request that are not yet validated in the method interface.
+This is especially important if your service expects single values that might be accessible as array value to the OPA.
+While the service (e.g. in case of query parameters) only considers the first value, make sure to not authorize on other values in the OPA.  
+
+The following listing shows an example extension that adds a fixed boolean entry:
+```java
+public class ExampleOpaInputExtension implements OpaInputExtension<Boolean> {
+  @Override
+  public Boolean createAdditionalInputContent(ContainerRequestContext requestContext) {
+    return true;
+  }
+}
+```
+
+Register the extension during the `OpaBundle` creation:
+```java
+OpaBundle.builder()
+    .withOpaConfigProvider(YourConfiguration::getOpa)
+    .withInputExtension("myExtension", new ExampleOpaInputExtension())
+    .build();
+```
+
+Access the additional input inside a policy `.rego`-file in this way:
+```rego
+exampleExtensionWorks {
+    # check if path match '/contracts' 
+    input.path = ["contracts"]
+
+    # check if the custom input has a certain value 
+    input.myExtension == true 
+}
+```
--- a/sda-commons-server-auth/src/main/java/org/sdase/commons/server/opa/OpaBundle.java
+++ b/sda-commons-server-auth/src/main/java/org/sdase/commons/server/opa/OpaBundle.java
@@ -207,6 +207,12 @@ public class OpaBundle<T extends Configuration> implements ConfiguredBundle<T> {
     }
 
     @Override
+    public <T extends Configuration> OpaBuilder<T> withOpaConfigProvider(
+        OpaConfigProvider<T> opaConfigProvider) {
+      return new Builder<>(opaConfigProvider);
+    }
+
+    @Override
     public OpaBuilder<C> withTracer(Tracer tracer) {
       this.tracer = tracer;
       return this;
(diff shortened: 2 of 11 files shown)


sda-commons-server-auth/src/main/java/org/sdase/commons/server/opa/extension/OpaInputExtension.java
```diff
@@ -0,0 +1,84 @@
...
+   *   public JsonNode filter(ContainerRequestContext requestContext, ObjectMapper mapper) {
+   *     return mapper.valueToTree(true);
+   *   }
+   * </pre>
+   *
+   * @param requestContext the request context
+   * @param mapper the object mapper to serialize values
+   * @return the JsonNode that should be added as child of the extension's namespace property.
+   */
+  JsonNode filter(ContainerRequestContext requestContext, ObjectMapper mapper);
```
JoergSiebahn, line 83: I think the implementing class should not be aware about the technical detail that this implementation will be used in a filter. In my opinion we should find a name that clarifies what should happen in this method, maybe something like `createAdditionalPropertyContent(…)`.

I also think, the mapper should not be part of the method signature. The class can be instantiated with an appropriate mapper.

Have you considered returning either `Object` or a generic type? The interface may be implemented easier in that case. Is there a technical reason, that prevents merging the objects later?

sda-commons-server-auth/src/main/java/org/sdase/commons/server/opa/OpaBundle.java
```diff
@@ -206,6 +251,30 @@ private Builder(OpaConfigProvider<C> opaConfigProvider) {
...
+    @Override
+    public OpaExtensionsBuilder<C> withoutHeadersExtension() {
+      this.addHeadersExtension = false;
+      return this;
+    }
+
+    @Override
+    public OpaBuilder<C> withInputExtension(OpaInputExtension extension) {
+      // Check if the namespace of an extension would override any original field of the OpaInput.
+      if (Arrays.stream(OpaInput.class.getDeclaredFields())
```
JoergSiebahn, line 269: You care about the known properties but what is about already added `OpaInputExtension`s?

sda-commons-server-auth/src/main/java/org/sdase/commons/server/opa/extension/OpaInputExtension.java
dhenneke: I first used `Object` but found it too unprecise and hard to understand what this means without knowing the documentation. But I haven't though about a generic type.

sda-commons-server-auth/src/main/java/org/sdase/commons/server/opa/extension/OpaInputExtension.java
```diff
@@ -0,0 +1,84 @@
...
+ */
+public interface OpaInputExtension {
+
+  /**
+   * The namespace (i.e. property name) where this extension is included. Please stick to naming the
+   * property in snake_case or camelCase.
+   *
+   * @return the namespace to store the value in.
+   */
+  String getNamespace();
```
JoergSiebahn, line 58: If we add the namespace at time of registration the consumers have more options and this interface can be a functional interface to be implemented on demand.

sda-commons-server-auth/src/main/java/org/sdase/commons/server/opa/filter/OpaAuthFilter.java
```diff
@@ -131,7 +130,14 @@ public void filter(ContainerRequestContext requestContext) {
         // process the actual request to the open policy agent server
         String[] path =
             uriInfo.getPathSegments().stream().map(PathSegment::getPath).toArray(String[]::new);
-        OpaRequest request = OpaRequest.request(jwt, path, method, trace, headers);
+        OpaInput opaInput = new OpaInput(jwt, path, method, trace);
+        ObjectNode objectNode = om.convertValue(opaInput, ObjectNode.class);
+
+        // append the input extensions to the input object
+        inputExtensions.forEach(
+            ex -> objectNode.set(ex.getNamespace(), ex.filter(requestContext, om)));
```
JoergSiebahn, line 138: Maybe we need a test and document how implementations can deal with `RequestContext#getEntityStream()`. It may be the case that it is broken after once read. This happens regular when one is trying to implement a debug request logger that logs the request or response body.

We may have to wrap the `RequestContext` and throw an Exception or copy and reset the `InputStream` when `getEntityStream()` is called. maybe we can do the latter in a second PR. In that case we have to watch effects on memory consumption.

sda-commons-server-auth/src/main/java/org/sdase/commons/server/opa/extension/OpaInputExtension.java
JoergSiebahn: Well, `JsonNode` is as unprecise as `Object` but more complicate to handle.
The only thing you guarantee is that the returned object is serializable by the `ObjectMapper`. But I think that's not worth the complexity of usage.

sda-commons-server-auth/src/main/java/org/sdase/commons/server/opa/filter/OpaAuthFilter.java
JoergSiebahn: As discussed outside of GitHub we should also document drawbacks when query params are used, e.g. to map all occurences of a query param as List in the endpoints.

Also the header authorization has to be used carefully because services often only use the first occurrence of a header.

sda-commons-server-auth/src/main/java/org/sdase/commons/server/opa/OpaBundle.java
dhenneke: I will also check that there are no conflicts with already registered namespaces.

sda-commons-server-auth/src/main/java/org/sdase/commons/server/opa/filter/OpaAuthFilter.java
dhenneke: I'll add a test that shows this behavior but don't fix the issue in this PR.

sda-commons-server-auth/src/main/java/org/sdase/commons/server/opa/extension/OpaInputExtension.java
dhenneke: I changed it to a generic type and renamed the method.

sda-commons-server-auth/README.md
```diff
@@ -306,3 +306,47 @@ opa:
...
+The extension is called in the `OpaAuthFilter` and is able to access the current `RequestContext` to for example extract additional data from the request.
+_Overriding existing input properties (`path`, `jwt`, `httpMethod`) is not possible._
+
+This extension option should only be used if the normal authorization via constraints is not powerful enough.
+In general, a custom extension should not be necessary for most use cases.
+
+_Remark on accessing the request body in an input extension:_
+The extension gets the `ContainerRequestContext` as input to access information about the request.
+Please be aware to not access the request entity since this might break your service.
+A test that shows the erroneous behavior can be found in [`OpaBundleBodyInputExtensionTest.java`](./src/test/java/org/sdase/commons/server/opa/OpaBundleBodyInputExtensionTest.java)
```
JoergSiebahn, line 323: 👍 

Should we add a security note here that the service might only consider one value of a specific header or query param while the OPA is able to authorize on an array of those?

sda-commons-server-auth/src/main/java/org/sdase/commons/server/opa/OpaBundle.java
```diff
@@ -172,31 +196,65 @@ private boolean excludeSwagger() {
...
+
   public interface OpaBuilder<C extends Configuration> {
 
+    /**
+     * Register a custom {@link OpaInputExtension} that enriches the default {@link OpaInput} with
+     * custom properties that are sent to the Open Policy Agent. Prefer authorization based on the
+     * {@link OpaJwtPrincipal#getConstraintsAsEntity(Class) constraints} that are returned after
+     * policy execution.
+     *
+     * <p>Please note that is is prohibited to override properties that are already set in the
```
JoergSiebahn, line 231:
```suggestion
     * <p>Please note that it is prohibited to override properties that are already set in the
```

sda-commons-server-auth/src/main/java/org/sdase/commons/server/opa/OpaBundle.java
```diff
@@ -172,31 +196,65 @@ private boolean excludeSwagger() {
...
+     * original {@link org.sdase.commons.server.opa.filter.model.OpaInput}.
+     *
+     * @param namespace the namespace, i.e. the property name, that the input should be accessible
+     *     as
+     * @param extension the extension to register
+     * @param <T> the type of data that is added to the input
+     * @throws HiddenOriginalPropertyException thrown if the namespace of an input extension
+     *     interferes with a property name of the original {@link OpaInput}.
+     * @throws DuplicatePropertyException thrown if a namespace interferes with an already
+     *     registered extension.
```
JoergSiebahn, line 241: I like that we have distinct exceptions here 👍

sda-commons-server-auth/src/main/java/org/sdase/commons/server/opa/OpaBundle.java
```diff
@@ -172,31 +196,65 @@ private boolean excludeSwagger() {
...
+     * Register a custom {@link OpaInputExtension} that enriches the default {@link OpaInput} with
+     * custom properties that are sent to the Open Policy Agent. Prefer authorization based on the
+     * {@link OpaJwtPrincipal#getConstraintsAsEntity(Class) constraints} that are returned after
+     * policy execution.
+     *
+     * <p>Please note that is is prohibited to override properties that are already set in the
+     * original {@link org.sdase.commons.server.opa.filter.model.OpaInput}.
+     *
+     * @param namespace the namespace, i.e. the property name, that the input should be accessible
+     *     as
```
JoergSiebahn, line 235:
```suggestion
     * @param namespace the namespace is used as property name, that the input should be 
     *     accessible as in the OPA policy
```

sda-commons-server-auth/src/test/java/org/sdase/commons/server/opa/OpaBundleBodyInputExtensionTest.java
```diff
@@ -0,0 +1,212 @@
...
+import org.junit.Before;
+import org.junit.ClassRule;
+import org.junit.Test;
+import org.junit.rules.RuleChain;
+import org.sdase.commons.server.opa.config.OpaConfig;
+import org.sdase.commons.server.opa.extension.OpaInputExtension;
+import org.sdase.commons.server.testing.DropwizardRuleHelper;
+import org.sdase.commons.server.testing.LazyRule;
+
+public class OpaBundleBodyInputExtensionTest {
```
JoergSiebahn, line 39: 👍  I like how this test shows the difference and how it is documented. Quite a lot of code. But that's how it is when you have to bootstrap two Applications.

PR #209, opened 2020-03-31
Introduce builder methods to mock OPA with all JWT outputs available in AuthRule

JoergSiebahn:
I lost some time in figuring out why my OPA mocks did not match until I realized that I you can not use the same value in the request and the OPA. This PR adds methods for all data types the AuthRule is able to return the JWT. So the OPA mock builder can be used in any case easily.

### DIFF
--- a/sda-commons-server-auth-testing/src/main/java/org/sdase/commons/server/opa/testing/OpaRule.java
+++ b/sda-commons-server-auth-testing/src/main/java/org/sdase/commons/server/opa/testing/OpaRule.java
@@ -16,14 +16,21 @@ import com.github.tomakehurst.wiremock.client.ResponseDefinitionBuilder;
 import com.github.tomakehurst.wiremock.http.RequestMethod;
 import com.github.tomakehurst.wiremock.junit.WireMockClassRule;
 import com.github.tomakehurst.wiremock.matching.RequestPatternBuilder;
+import java.util.List;
+import java.util.Objects;
+import javax.ws.rs.core.MultivaluedMap;
 import org.junit.rules.ExternalResource;
 import org.junit.rules.RuleChain;
 import org.junit.runner.Description;
 import org.junit.runners.model.Statement;
 import org.sdase.commons.server.opa.filter.model.OpaResponse;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 @SuppressWarnings("WeakerAccess")
 public class OpaRule extends ExternalResource {
+
+  private static final Logger LOG = LoggerFactory.getLogger(OpaRule.class);
 
   private static final ObjectMapper OM = new ObjectMapper();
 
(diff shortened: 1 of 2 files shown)


sda-commons-server-auth-testing/src/main/java/org/sdase/commons/server/opa/testing/OpaRule.java
```diff
@@ -139,6 +146,37 @@ public void mock(BuildBuilder builder) {
...
+      if (jwtWithBearerPrefix == null || !jwtWithBearerPrefix.toLowerCase().startsWith("bearer ")) {
+        LOG.warn(
+            "Requested to mock OPA request from header value but no Bearer prefix found in {}",
+            jwtWithBearerPrefix);
+        return this;
+      }
+      return withJwt(jwtWithBearerPrefix.substring("bearer".length()).trim());
+    }
+
+    default RequestExtraBuilder withJwtFromHeaders(MultivaluedMap<String, Object> headers) {
```
Fox32, line 160: This (or both) new methods feels a bit strange, doesn't the AuthRule provide the matching methods to directly generate the right values? Now you pack them first and then unpack them here again?

JoergSiebahn: It depends on how you write your tests. I came up to this extension when I had to modify test cases like this:

```
     String jwt = AUTH.auth().buildToken();

    OPA.mock(
        onRequest()
            // …
            .withJwt(jwt)
            // …

  Response response =
        createTarget()
            // …
            .request(MediaType.APPLICATION_JSON_TYPE)
            .header("Authorization", "Bearer " + jwt) // this should not be needed
            .get()) {
```

Where the previous version without JWT assertion in the OPA mock used simply:

```
            // …
            .header("Authorization", AUTH.auth().buildHeaderValue())
            // …
```

In other cases we use the simplest method `.headers(AUTH.auth().buildAuthHeader())`.

In my opinion it is much easier to enhance existing tests with the new JWT assertion in OPA if you provide methods for all variants the `AuthRule` offers.

Actually it took me some (too much) time to realize that I'm passing the full header value to `withJwt(String)` and „Bearer“ does not match in the OPA mock. I like to avoid that mistake a second time. Guidance with some explicit methods is appreciated.

Fox32: Ah, ok. Calling `buildXXX` twice might cause different tokens, that can be a problem.

dhenneke: But you still need to always store the jwt value (either as jwt, header, or map) in a variable. Why not store the token builder and then use the corresponding methods to get the correct value?

```java
// Configure the token once. Depending on the
// test class maybe as instance variable.
AuthBuilder jwt = AUTH.auth();

OPA.mock(
    onRequest()
        // …
        .withJwt(jwt.buildToken()) // get the signed token
        // …

Response response =
    createTarget()
        // …
        .request(MediaType.APPLICATION_JSON_TYPE)
        .headers(jwt.buildAuthHeader()) // get the signed token as header
        .get()) {
```

The signature algorithms that we use all produce reproducible signatures for the same input.

JoergSiebahn: Although the current implementation (in the test builder) generates the same token every time for the same content, you should not rely on that. Usually a token will contain a timestamp of creation and at least with other algorithms the signature should be salted. If one of these changes, e.g. supporting different algorithms or adding `iat` to the generated test token for any reason I would prefer that my test code does not break for reasons that are hard to find.

PR #263, opened 2020-05-17
docs: add javadoc badge

Fox32:


README.md
```diff
@@ -1,4 +1,5 @@
 [![Latest Release](https://img.shields.io/github/v/release/sda-se/sda-dropwizard-commons?label=latest)](https://github.com/SDA-SE/sda-dropwizard-commons/releases/latest)
+[![javadoc](https://javadoc.io/badge2/org.sdase.commons/sda-commons-bom/javadoc.svg)](https://javadoc.io/doc/org.sdase.commons/sda-commons-bom)
```
dhenneke, line 2: I'm not sure if the bom is the right one to choose here. At least it doesn't provide a javadoc so the linked resource will not work :-) Maybe sda-commons-server-starter?

Fox32: How about keeping the badge image as-is and changing the link to: https://javadoc.io/doc/org.sdase.commons
However that content looks wrong?

Fox32: Looks like the contents of that page are filled ones they are generated. Maybe we should provide a link in every sub-readme to initialize them ;)

JoergSiebahn:
There are some modules that do not have Javadoc, e.g. sda-commons-shared-forms. I think it is misleading to put a link in that modules. Maybe it is enough to add it in the main readme.

Fox32:
> Maybe it is enough to add it in the main readme. This keeps it more simple as well, I think.

I did this in the first try, but then we don't have any documentation as it is only generated automatically when it is referenced. By adding a badge for everything we get the documentation generated automatically and can add a badge linking to everything in the main readme.

JoergSiebahn:
Thanks for the explanation. I expected that it is generated/loaded at the first call of a module if it is not empty. That would be suitable with the main link only. But if that is not the case, this PR is fine for me.

PR #264, opened 2020-05-17
feat: publish the artifacts to maven central

### LINKED ISSUE #60: Provide artifacts at Maven central
Our users would like to use the library without having access to our internal Maven repository. We currently provide two kinds of builds internally: releases and PR snapshots.
Therefore we should provide at least release snapshots in a public repository (Maven central), but better both (PR snapshots somewhere else).

To implement the publish process we can use these resources
* General Sonatyp Guide: https://central.sonatype.org/pages/ossrh-guide.html
* Gradle specifc: https://central.sonatype.org/pages/gradle.html
* Promote releases to Maven central: https://github.com/Codearte/gradle-nexus-staging-plugin/
* JavaDoc hosting: https://www.javadoc.io/

dhenneke:
The release are also published to our internal nexus. The snapshots are still only uploaded to our internal nexus.

Closes #60
Closes #168


dhenneke:
We should also mention that the user needs the following repository configuration:
`maven { url "https://packages.confluent.io/maven/" }`.

dhenneke:
I removed the dependency to confluent in the `sda-commons-server-kafka` bundle. The tests are green so I guess it was a leftover from a former dependency conflict.

.releaserc.yml
```diff
@@ -9,4 +9,4 @@ plugins:
 
   # use the publish hook of semantic-release to publish the modules to the nexus
   - - "@semantic-release/exec"
-    - publishCmd: "export SEMANTIC_VERSION=${nextRelease.version} && ./gradlew publish"
+    - publishCmd: "export SEMANTIC_VERSION=${nextRelease.version} && ./gradlew publishMavenPublicationToMavenCentralRepository closeAndReleaseRepository publishMavenPublicationToSdaInternRepository"
```
JoergSiebahn, line 12: Do we have some kind of dry run? Or a manual test without affecting our master branch?

dhenneke: https://search.maven.org/search?q=org.sdase.commons

build.gradle
```diff
@@ -262,6 +289,16 @@ task codeCoverageReport(type: JacocoReport) {
...
 }
 
+// Automatically close and release the staging repository that gets created
+// during the upload to maven central.
+nexusStaging {
+  username System.getenv('SDA_SONATYPE_USER')
+  password System.getenv('SDA_SONATYPE_PASSWORD')
+  packageGroup "org.sdase.commons"
+  // wait for max. 10 minutes
+  numberOfRetries 300
```
JoergSiebahn, line 299: How do number of retries map on minutes? Are 10 minutes enough? I already encountered quite long times of the checks in Sonatype staging with much smaller projects.

dhenneke: The delay between the checks is [2 seconds by default](https://github.com/Codearte/gradle-nexus-staging-plugin#configuration). By default it only waited for 40 seconds but that didn't worked since it took about 1-2 Minutes. I guess we will see if it is enough.

.releaserc.yml
JoergSiebahn: Ok, let's give it a try.

JoergSiebahn:
```
> Task :sda-commons-bom:signMavenPublication FAILED

FAILURE: Build failed with an exception.

* What went wrong:
Execution failed for task ':sda-commons-bom:signMavenPublication'.
> Cannot perform signing task ':sda-commons-bom:signMavenPublication' because it has no configured signatory
```
Can we exclude the signing for local publication?
Maybe [this](https://gitlab.com/codefetti/proxy/-/blob/master/build.gradle#L115) can help.

JoergSiebahn:
🎉

PR #287, opened 2020-06-03
Add AsyncAPI spec helpers

Fox32:


sda-commons-dependencies/build.gradle
```diff
@@ -83,6 +84,7 @@ dependencies {
     api "io.opentracing.contrib:opentracing-web-servlet-filter:0.4.1"
     api "org.jetbrains.kotlin:kotlin-stdlib:1.3.72" // override version from dropwizard-bom
 
+    api "javax.validation:validation-api:2.0.1.Final"
```
JoergSiebahn, line 87: This clashes with the old version of hibernate validator which we can't upgrade in Dropwizard, as far as I know. Therefore we always added this version as a test dependency only in the project. There is a high risk that someone uses the new validation-api annotations in production code. That would only have an effect on the documentation but won't trigger a validation with hibernate-validator. There are some annotations that the old hibernate version provides itself which do not exist in 1.x of the validation-api. Right now I do not have a good solution for this. It may be needed to register a validation class for the new annotations in hibernate validator manually.

Fox32: I don't think I can downgrade it as it will run into an `java.lang.NoClassDefFoundError: javax/validation/constraints/NotBlank` error 😢 

Do you know about this? Do we have tests for this?

Fox32: Looks like this would be fixed with Dropwizard two. Maybe we make this an experimental feature, document it so that users have to override the dependency in the consuming package and know about the risks. With a Dropwizard 2 upgrade we can make it the default.

sda-commons-shared-asyncapi/README.md
```diff
@@ -0,0 +1,92 @@
...
+        try (AutoCloseableSoftAssertions softly = new AutoCloseableSoftAssertions()) {
+            softly.assertThat(actual)
+                .as("The current asyncapi.yaml file is not up-to-date. If this happens "
+                    + "locally, just run the test again. The asyncapi.yaml file is updated "
+                    + "automatically after running this test. If this happens in the CI, make sure "
+                    + "that you have committed the latest asyncapi.yaml file!")
+                .isEqualTo(expected);
+
+            Files.write(asyncApiPath, expected.getBytes());
+        }
```
JoergSiebahn, line 80: We should ensure that the expected file is written initially when the test is run for the first time and on error.

```suggestion
    try (AutoCloseableSoftAssertions softly = new AutoCloseableSoftAssertions()) {
      String actual = new String(Files.readAllBytes(asyncApiPath));
      softly.assertThat(actual)
          .as("The current asyncapi.yaml file is not up-to-date. If this happens "
              + "locally, just run the test again. The asyncapi.yaml file is updated "
              + "automatically after running this test. If this happens in the CI, make sure "
              + "that you have committed the latest asyncapi.yaml file!")
          .isEqualTo(expected);
    }
    finally {
      Files.write(asyncApiPath, expected.getBytes());
    }
```
Sonar lint locally claims that there is no assertion in the try block. This is a false positive of course. But maybe we find a way to write the example so that it does not raise a warning after copy paste.

sda-commons-dependencies/build.gradle
Fox32: We fixed this by reverting to the old version and include the missing classes in our code.

Fox32:
Thanks for the feedback:

> **Tested against our model**

This approach is using `mbknor-jackson-jsonSchema` because it was recommended by the Eventing-Team (see our [messaging guide](https://draft.sda.dev/core-concepts/communication/messaging/#Messaging-Documentation) and the [example](https://github.com/SDA-SE/ecosystem-eventing-asyncapi-experiments/tree/master/dropwizard)). But it also recommends the `jsonschema-generator`, however it doesn't support annotations for example values, which we think is very valuable.
For most use cases `mbknor-jackson-jsonSchema` should be sufficient, the tests include an example with multiple levels of inheritance, which works well. Your use case is very complex and should not be something we should recommend to new users. However, this implementation allows to use both, while the `jsonschema-generator` might need some additional work. You can pass a schema from another generator [into the builder](https://github.com/SDA-SE/sda-dropwizard-commons/blob/feat/asyncapi/sda-commons-shared-asyncapi/src/main/java/org/sdase/commons/shared/asyncapi/AsyncApiGenerator.java#L70).


> **Automation**

The main idea of the approach is to have a valid template file, that also works with other tools and also work without embedding the schema files. This can be useful if you plan to use a schema first approach and what to create a self contained file. I'm not sure if you can achieve this without any references at all.

The current implementation has no opinion on whether one references only the base class of the events or every event on its own. You can decide that in your template file. I tried both and we decided reference them separately in our solution as it allows to provide additional documentation.


> **Providing JSON Schema**

Creating a self contained file is important to use it with other tools, like the playground or a later catalog. It isn't mandatory in our eventing guide yet, but once it evolves we should add a similar [rule as we have for OpenAPI files](
https://sda.dev/core-concepts/communication/restful-api-guide/#RESTfulAPIGuide-MUST:SpecifytheAPIusingOpenAPI).

For standalone JSON schemas this PR provides the `JsonSchemaGenerator` which is mentioned in the readme. I would left it to the reader, but if a full example is required please create a follow-up PR.


I don't see any requested changes and would proceed otherwise

sda-commons-shared-asyncapi/README.md
Fox32: Also changed to compare the structure instead of the string representation (has problems on Windows)

sda-commons-shared-asyncapi/README.md
```diff
@@ -0,0 +1,128 @@
...
+    .forClass(BaseEvent.class)
+    .generateYaml();
+```
+
+
+## Document Models
+
+You can document the models using annotations like `JsonPropertyDescription` from Jackson or
+`JsonSchemaExamples` from [`mbknor-jackson-jsonSchema`](https://github.com/mbknor/mbknor-jackson-jsonSchema).
+See the tests of this module for [example model classes](./src/test/java/org/sdase/commons/shared/asyncapi/models).
```
JoergSiebahn, line 128:
```suggestion
See the tests of this module for [example model classes](./src/test/java/org/sdase/commons/shared/asyncapi/models).
Note that this requires to add the module as compile time dependency.
```

sda-commons-shared-asyncapi/README.md
```diff
@@ -0,0 +1,131 @@
...
+                YamlUtil.load(expected, new TypeReference<Map<String, Object>>() {});
+
+            softly.assertThat(actualJson)
+                .as("The current asyncapi.yaml file is not up-to-date. If this happens "
+                    + "locally, just run the test again. The asyncapi.yaml file is updated "
+                    + "automatically after running this test. If this happens in the CI, make sure "
+                    + "that you have committed the latest asyncapi.yaml file!")
+                .isEqualToComparingFieldByFieldRecursively(expectedJson);
+        } finally {
+            Files.write(asyncApiPath, expected.getBytes());
```
christopher-cudennec, line 91: You should add the encoding to "getBytes".

sda-commons-shared-asyncapi/src/main/java/org/sdase/commons/shared/asyncapi/internal/JsonReference.java
```diff
@@ -0,0 +1,72 @@
...
+   */
+  static JsonReference parse(String ref) {
+    int urlEndIndex = ref.indexOf('#');
+
+    if (urlEndIndex < 0) {
+      throw new IllegalArgumentException("Invalid reference format: " + ref);
+    }
+
+    String url = urlEndIndex == 0 ? null : ref.substring(0, urlEndIndex);
+    String pointerText = ref.substring(urlEndIndex + 1);
```
christopher-cudennec, line 67: Is "foo#" a possible value for "ref"? This might cause an exception here.

sda-commons-shared-asyncapi/src/main/java/org/sdase/commons/shared/asyncapi/internal/JsonSchemaEmbedder.java
```diff
@@ -0,0 +1,138 @@
...
+      current.forEach(
+          jsonNode ->
+              resolveInPlace(target, root, jsonNode, resolvedReferences, shouldEmbedInternal));
+    } else if (current.isObject()) {
+      List<Entry<String, JsonNode>> fields = new ArrayList<>();
+      current.fields().forEachRemaining(fields::add);
+
+      fields.forEach(
+          e -> {
+            if (REF.equals(e.getKey()) && e.getValue().isTextual()) {
```
christopher-cudennec, line 66: ignore case?

sda-commons-shared-asyncapi/src/main/java/org/sdase/commons/shared/asyncapi/internal/JsonSchemaEmbedder.java
```diff
@@ -0,0 +1,138 @@
...
+      ObjectNode parentNode = (ObjectNode) ensureParentNode(target, pointer.head());
+      parentNode.set(pointer.last().getMatchingProperty(), JsonNodeFactory.instance.objectNode());
+      node = target.at(embedPointer);
+    }
+
+    return node;
+  }
+
+  private String replaceSpecialCharacters(String value) {
+    return value.replaceAll("[^a-zA-Z0-9]", "");
```
christopher-cudennec, line 136: too strict for URLs? is it important to represent the URL properly?

sda-commons-shared-asyncapi/src/main/java/org/sdase/commons/shared/asyncapi/JsonSchemaGenerator.java
```diff
@@ -0,0 +1,101 @@
...
+    }
+
+    @Override
+    public JsonNode generate() {
+      ObjectMapper objectMapper = ObjectMapperConfigurationUtil.configureMapper().build();
+      com.kjetland.jackson.jsonSchema.JsonSchemaGenerator jsonSchemaGenerator =
+          new com.kjetland.jackson.jsonSchema.JsonSchemaGenerator(
+              objectMapper,
+              JsonSchemaConfig.vanillaJsonSchemaDraft4()
+                  .withJsonSchemaDraft(JsonSchemaDraft.DRAFT_07)
```
christopher-cudennec, line 84: According to https://json-schema.org/specification.html#draft-numbering this is not the latest draft. Why not use DRAFT_2019_09?

sda-commons-shared-asyncapi/src/main/java/org/sdase/commons/shared/asyncapi/internal/JsonReference.java
Fox32: Yes it is, and we have a test case that covers it and I don't see an exception here.

sda-commons-shared-asyncapi/README.md
Fox32: Done

sda-commons-shared-asyncapi/src/main/java/org/sdase/commons/shared/asyncapi/internal/JsonSchemaEmbedder.java
Fox32: The idea is to generate a temporary and unique class name, I don't really care about how it looks, just that it is safe. 
Url will be a relative path most of the time, so I wouldn't consider urls here right now.

sda-commons-shared-asyncapi/src/main/java/org/sdase/commons/shared/asyncapi/JsonSchemaGenerator.java
Fox32: @JoergSiebahn what do you think?

dhenneke: AsyncAPI uses DRAFT 07: https://www.asyncapi.com/docs/specifications/2.0.0/#a-name-messageobjectschemaformattable-a-schema-formats-table

But I am not sure what the differences between the drafts are and whether AsyncAPI plans to change the used draft version.

sda-commons-shared-asyncapi/src/main/java/org/sdase/commons/shared/asyncapi/internal/JsonSchemaEmbedder.java
Fox32: I can't find a statement that keywords in json schema are case insensitive nor that they are case sensitive. However the underlining JSON standard is case sensitive (you can have both the properties `$ref` and `$REF` in the same document), so I think the keywords should be case sensitive.

JoergSiebahn:
_Finally we can generate Async API with a quite dirty workaround from our streaming model with this tooling. As our model is quite complex such a workaround seems to be acceptable although it is not the best solution._

Declaring this module as experimental is a good way to collect further experience in creating JSON Schemas and Async API documentation across teams. The used tooling, specifically the JSON Schema generator, has a few drawbacks. But it also has the the advantage to declare examples.

I'd like to keep the drawbacks in memory so we can solve them before this module becomes mature. This may result in a change of the used JSON Schema generator. There is [another one](https://github.com/victools/jsonschema-generator) that works better in some of the mentioned cases when properly configured and is more easily extendable or customizable. But it has no support for examples yet.

Lets collect some drawbacks of the experimental module now:

- To produce well documented schemas it is required to add a compile time dependency which not only adds annotations but also the schema generation services and transitive dependencies from the Scala world. These will also influence foreign services in case a streaming model is provided as library. Users of the new module may be required to work carefully with something like the provided scope as used in Maven.
- Enums can not be documented at class level. Multiple use of the same enum requires to repeat the documentation of enum values at every location where it is used as type of a property. If enums are the items of a list it renders nicely with Async API when the meaning of the array field is documented individually while the items get their documentation from the enum itself. This is not possible with the current tooling.
- `boolean` fields and also `Boolean`s are always added as required although this is unnecessary for `boolean`s with our object mapper and wrong for Java `Boolean`s in general. I'm not sure if this applies to other primitives as well.
- The property used to distinguish sub types in inheritance can not be documented. This is not that critical if there is exactly one value per sub type which can be properly documented at type level.
- In complex cases inheritance is not identified correctly without workarounds. This can be frustrating for new users. But there are some cases in all tools where developers need specific structure to satisfy Jackson and the documentation at the same time.
- It seems, that the generator is actually creating Draft 4 and labels it as Draft 7 ( `JsonSchemaConfig.vanillaJsonSchemaDraft4().withJsonSchemaDraft(JsonSchemaDraft.DRAFT_07)` ). We should check that and learn more about the differences of the versions if this assumption is correct.
- The generator is not as customizable and extendable by SDA or at project level as the alternatives. It may be difficult to fix the drawbacks or address project specific problems without working in the raw generated schema or creating a [PR in the source project](https://github.com/mbknor/mbknor-jackson-jsonSchema/pull/124).

Users of this module should be aware that the implementation details may change in the future. This may affect the result of the generated schema and the used annotations in production code. To have less trouble after changes, users should use `org.sdase.commons.shared.asyncapi.JsonSchemaGenerator` and `org.sdase.commons.shared.asyncapi.AsyncApiGenerator` as an abstraction layer in their code.

christopher-cudennec:
Werbetrommel für den Dev-Blog 🥁  😄

Issue #259, opened 2020-05-13
sda-commons-dependencies - Upgrade to Dropwizard 2.x needed


mattias88:
... for Dropwizard JSON Logging

We are using Splunk as log-server and fluentd as log-shipper. A shortcoming of fluentd is the maximum size of a single log statement - 16 KB. Stacktraces easily break this limit. Therefore I want to configure the stacktrace- / exception-handling in Dropwizard-JSON-Logging. This feature 
was added with PR #2501 to Dropwizard 2.x but is missing in Dropwizard 1.x.
Please consider upgrading to Dropwizard 2.x

Fox32:
Hey Mathias, thanks for your suggestion. That is by far not the only reason to update to Dropwizard 2.

- Security updates, but right now 1.3.x is still maintained
- Dependency trouble around Kafka and other libraries
- Compatibility to other Dropwizard libraries

We already started to work on a migration in the branch https://github.com/SDA-SE/sda-dropwizard-commons/tree/feat/dropwizard2.
However we run into some trouble around Jersey and there are some breaking changes were we need a plan how we migrate to the new version.

Right now our roadmap for this not completely clear, but we will work out how we will solve this soon and post an update here.

dhenneke:
We started to actively work on the 2.0 Update in #303 and succeeded to find all issues that need a fix. We are also planning to do some deprecations and removals during the update, and also need to test this internally before we can release the update.

JoergSiebahn:
Some information about the progress on Dropwizard 2:

We created an [alpha release](https://github.com/SDA-SE/sda-dropwizard-commons/releases/tag/2.0.0-alpha-2), that is available at [Maven Central](https://search.maven.org/search?q=g:org.sdase.commons%20v:2.0.0-alhpa-2). The release notes of [1.0.0](https://github.com/SDA-SE/sda-dropwizard-commons/releases/tag/1.0.0) document the deprecations that are removed with the next major version. There is also a [blog post](https://sda.dev/blog/2020/07/22/dropwizard-2-upgrade-guide/) (SDA SE or SDA customer account required) with information about the changes and tasks to do when upgrading.

@mattias88 it would be very helpful, if you can try the alpha release in your service and check for compatibility. We appreciate your feedback.

mattias88:
> @mattias88 it would be very helpful, if you can try the alpha release in your service and check for compatibility. We appreciate your feedback.

[...]

PR #303, opened 2020-06-11
Dropwizard 2

dhenneke:


sda-commons-server-swagger-example/src/main/java/org/sdase/commons/server/swagger/example/people/rest/PersonEndpoint.java
```diff
@@ -88,4 +90,10 @@ private URI getPersonUri(int id) {
         .resolveTemplate("personId", id)
         .build();
   }
+
+  @Override
+  public boolean configure(FeatureContext context) {
+    // TODO: Add comment here
+    return true;
+  }
```
JoergSiebahn, line 98: Maybe we can add an interface like `SdaEndpoint` or `ContextAware` or ??? for convenience that extends `Feature` and implements `configure` as default method?

sda-commons-server-swagger/src/test/java/org/sdase/commons/server/swagger/SwaggerBundleTestApp.java
```diff
@@ -62,7 +60,7 @@ public void run(Configuration configuration, Environment environment) {
   @Produces(APPLICATION_JSON)
   @ApiOperation(value = "get", response = PartnerResource.class)
   @ApiResponses(@ApiResponse(code = 200, message = "get", response = PartnerResource.class))
-  public PartnerResource getJohnDoe() {
+  public PartnerResource getJohnDoe(@Context UriInfo uriInfo) {
```
JoergSiebahn, line 63: Is it intentional not to implement `Feature` here?

sda-commons-server-swagger-example/src/main/java/org/sdase/commons/server/swagger/example/people/rest/PersonEndpoint.java
dhenneke: How about `ContextAwareEndpoint`? Though this would also be usable by any other class that is no JAX-RS resource.

JoergSiebahn: There is one thing that came to my mind this morning. Does it matter if the endpoint is context aware? From my understanding, Dropwizard 2 will reject all instances that do not implement _Feature_ (or similar) at the time of registration? So maybe _Endpoint_ is enough.

Although any interface for this purpose can be used by any service, I would recommend it for endpoints only. In some cases it may be useful for services that are only used by endpoints for the purpose of handling context information. But they may also be implemented with methods that convert a given context specific object into the desired value. If we keep all service context unaware and give the required information when calling a method, testing of the services should be much easier and we only have to take care about the context (mock it or build integration tests) when testing endpoints.

dhenneke: It's not that Dropwizard 2 _rejects_ instances. It's just how they register it to jersey.

Some Context: When calling `environment.jersey().register(new Test())`, and `Test` is not identified as any kind of "Provider" by jersey, they create a new `AbstractBinder` that does `bind(object).to(clazz)`. But jersey will never inject any fields or contexts into `object`. A trivial solution would be to call `injectionManager.inject(object)` manually in the Dropwizard code, however, there is no trivial way to get access to Jersey's `InjectionManager` in an `AbstractBinder`.

JoergSiebahn: Thanks for the details. In this case `ContextAwareEndpoint` sounds most valuable as you suggested. It clearly names the purpose of the interface.

JoergSiebahn:
We have some very informational commit-messages in this PR which will not make it in the release notes automatically, e.g. https://github.com/SDA-SE/sda-dropwizard-commons/pull/303/commits/bde5c78d37d5d32d6dbc82a3ca00dcd375958743 

Should we collect the change information we want publish in the PR description?

dhenneke:
> Should we collect the change information we want publish in the PR description?

I am not sure if these are information that are interesting for anyone else than the reviewer of this specific commit since they are often internal changes that don't affect users of the library. Furthermore this commits still lacks a BREAKING CHANGE notice that I didn't include yet because I am not sure what we actually want to write there.

JoergSiebahn:
Note: We need to merge this branch locally with FF because there are tags that reference commits in this branch.

JoergSiebahn (approved):

JoergSiebahn:
closes #259

christopher-cudennec (approved):

PR #320, opened 2020-06-23
Provide `OpaJwtPrincipal` as injectable bean

JoergSiebahn:
In many of our services we add a dedicated class or method that extracts and casts the principal from the injected `SecurityContext`.

To avoid the repeated block of code that always has to be tested, we can inject the `OpaJwtPrincipal` directly to read the constraints defined by the OPA.

Unfortunately there seems to be no option to provide constraints in the context as they depend on the project.

### DIFF
--- a/sda-commons-server-auth-testing/src/test/java/org/sdase/commons/server/opa/testing/test/AuthAndOpaBundleTestApp.java
+++ b/sda-commons-server-auth-testing/src/test/java/org/sdase/commons/server/opa/testing/test/AuthAndOpaBundleTestApp.java
@@ -3,7 +3,6 @@ package org.sdase.commons.server.opa.testing.test;
 import io.dropwizard.Application;
 import io.dropwizard.setup.Bootstrap;
 import io.dropwizard.setup.Environment;
-import java.io.IOException;
 import javax.ws.rs.GET;
 import javax.ws.rs.Path;
 import javax.ws.rs.core.Context;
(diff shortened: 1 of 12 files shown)


JoergSiebahn:
The [used scope](https://github.com/SDA-SE/sda-dropwizard-commons/pull/320/files#diff-33ba1e5c09a67641a3cd74e83d558bc9R28) comes from an internal Jersey package. I found a few examples that had similar needs like providing something in request scope and all used this annotation. So I think, it's the way to go.

sda-commons-server-auth-testing/src/test/java/org/sdase/commons/server/opa/testing/test/AuthAndOpaBundleTestApp.java
```diff
@@ -43,14 +42,19 @@ public void run(AuthAndOpaBundeTestAppConfiguration configuration, Environment e
...
 
       PrincipalInfo result =
           new PrincipalInfo()
               .setName(principal.getName())
               .setJwt(principal.getJwt())
-              .setConstraints(principal.getConstraintsAsEntity(ConstraintModel.class));
+              .setSub(
+                  principal.getClaims() == null
+                      ? null
+                      : principal.getClaims().get("sub").asString())
```
christopher-cudennec, line 55: Also check the claim value for `null`?

JoergSiebahn: In a test where `sub` has a default value from the `AuthRule`? I think that is too much boilerplate.

christopher-cudennec: Alright. Did not see that this is just a test! :)

JoergSiebahn:
> In future, we should also consider if we want to provide commonly used classes from other bundles to the DI.

We should do that very carefully. The `OpaJwtPrincipal` is ideally used in an endpoint implementation to extract the constraints and pass them to the next service. If we keep endpoints minimal to collect context information it helps to test the services without full REST integration tests. If we add injected fields to other services, it makes them harder to test. I think the desired location of use should be considered when adding other classes to the context.

PR #356, opened 2020-07-08
fix(server-auth): OPA's policy exists health check handles null respo…

christopher-cudennec:
…nses in case of misconfiguration

### DIFF
--- a/sda-commons-server-auth/src/main/java/org/sdase/commons/server/opa/health/PolicyExistsHealthCheck.java
+++ b/sda-commons-server-auth/src/main/java/org/sdase/commons/server/opa/health/PolicyExistsHealthCheck.java
@@ -1,7 +1,6 @@
 package org.sdase.commons.server.opa.health;
 
 import com.codahale.metrics.health.HealthCheck;
-import com.fasterxml.jackson.databind.node.NullNode;
 import javax.ws.rs.client.Entity;
 import javax.ws.rs.client.WebTarget;
 import org.sdase.commons.server.opa.filter.model.OpaResponse;
@@ -29,7 +28,9 @@ public class PolicyExistsHealthCheck extends HealthCheck {
       // If there is an exception, the check will be unhealthy
       OpaResponse opaResponse = client.request().post(Entity.json(null), OpaResponse.class);
 
-      if (opaResponse == null || opaResponse.getResult() instanceof NullNode) {
+      if (opaResponse == null
+          || opaResponse.getResult() == null
+          || opaResponse.getResult().isNull()) {
         LOGGER.warn("The policy response seems not to be SDA guideline compliant");
         return Result.unhealthy("The policy response seems not to be SDA guideline compliant");
       }
--- a/sda-commons-server-auth-testing/src/test/java/org/sdase/commons/server/opa/testing/PolicyExistsHealthCheckIT.java
+++ b/sda-commons-server-auth-testing/src/test/java/org/sdase/commons/server/opa/testing/PolicyExistsHealthCheckIT.java
@@ -1,8 +1,12 @@
 package org.sdase.commons.server.opa.testing;
 
+import static com.github.tomakehurst.wiremock.client.WireMock.anyUrl;
+import static com.github.tomakehurst.wiremock.client.WireMock.okJson;
+import static com.github.tomakehurst.wiremock.client.WireMock.post;
 import static org.assertj.core.api.Assertions.assertThat;
 import static org.sdase.commons.server.opa.testing.OpaRule.onAnyRequest;
 
+import com.codahale.metrics.health.HealthCheck.Result;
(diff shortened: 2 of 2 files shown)


sda-commons-server-auth/src/main/java/org/sdase/commons/server/opa/health/PolicyExistsHealthCheck.java
```diff
@@ -29,7 +29,9 @@ public Result check() {
       // If there is an exception, the check will be unhealthy
       OpaResponse opaResponse = client.request().post(Entity.json(null), OpaResponse.class);
 
-      if (opaResponse == null || opaResponse.getResult() instanceof NullNode) {
+      if (opaResponse == null
+          || opaResponse.getResult() == null
+          || opaResponse.getResult() instanceof NullNode) {
```
JoergSiebahn, line 34:
```suggestion
          || opaResponse.getResult().isNull()) {
```

sda-commons-server-auth/src/test/java/org/sdase/commons/server/opa/health/PolicyExistsHealthCheckTest.java
```diff
@@ -0,0 +1,35 @@
+package org.sdase.commons.server.opa.health;
```
dhenneke, line 1: There are already tests in [the sda-commons-server-auth-testing module](https://github.com/SDA-SE/sda-dropwizard-commons/blob/fda0809f4394306aa618012bf72d6d0a4f65bd5c/sda-commons-server-auth-testing/src/test/java/org/sdase/commons/server/opa/testing/HealthCheckTest.java).

The main problem with those tests is that they don't check the message and they are missing a test for this specific issue (There is a test that checks if `{"result": null}` is working but not if `{}` also works).

christopher-cudennec: You mean OpaResponsesIT, right?

Personally I prefer:
* fast unit tests over slow integration tests
* having tests close to the classes to be tested / didn't expect the tests to be in another module

sda-commons-server-auth/src/main/java/org/sdase/commons/server/opa/health/PolicyExistsHealthCheck.java
christopher-cudennec: Applied as suggested.

sda-commons-server-auth/src/test/java/org/sdase/commons/server/opa/health/PolicyExistsHealthCheckTest.java
dhenneke: No, I mean HealthCheckTest (which should be named PolicyExistsHealthCheckIT but that's another story). I too think that this test should not be part of the `*-testing` module but this is a wider problem across the whole `*-server-auth` module (that would be solvable, but not in this PR).

Personally I more like the tests that show me what I can expect from the Health Check when a "real" OPA responds.

christopher-cudennec: I tried to extend `HealthCheckTest` but can't make the result `null`. The test always gives me a `NullNode`. I.e. it's good to have both tests because `null` results really happen in practice. Is this due to WireMock?

christopher-cudennec: I moved the unit test to the same package as the integration test and renamed it as suggested.

dhenneke: The default `OpaRule` will only send a `{"result": null}` due to the JSON serialization settings. Something like that might work: `OPA_RULE.mock(wire -> wire.stubFor(post(anyUrl()).willReturn(okJson("{}"))));`

dhenneke: It should also be possible to move the IT to the `*server-auth` module. But I guess that you then need a test-dependency to the `*-testing` module. I guess that this wasn't done originally to avoid circular dependencies. But I would expect that `server-auth:test -> server-auth-testing:compile -> server-auth:compile` should work.

christopher-cudennec: > OPA_RULE.mock(wire -> wire.stubFor(post(anyUrl()).willReturn(okJson("{}"))));

Yeah, that worked. Thanks! 👍 
I leave the test in "testing" module for now.

christopher-cudennec:
Read for Review :)

PR #363, opened 2020-07-10
fix(deps): upgrade hamcrest

JoergSiebahn:
hamcrest (2.x) and hamcrest-core (1.x) have mostly similar classes 
but are not compatible in all details the later version of hamcrest
comes with Awaitility, so we use that in favour of the older one
served with junit. It is assumed that junit does not use it itself
and therefore has no conflicts due to the changes.

As junit by default provides hamcrest, we still ship it. But it is
not used any more here in favour of AssertJ. All usages of hamcrest 
are replaced with AssertJ.

We need to manage the deprecated dependency hamcrest-core to 
overwrite transitive dependencies. Version 2.2 of hamcrest-core
contains deprecation notices only and has no conflicts with the new
artifact.

### DIFF
--- a/sda-commons-dependencies/build.gradle
+++ b/sda-commons-dependencies/build.gradle
@@ -135,6 +135,12 @@ dependencies {
 
     // sda-commons-server-testing
     api "junit:junit:4.13"
+    api "org.hamcrest:hamcrest:2.2"
+    // hamcrest-core is deprecated and overlaps with hamcrest, in hamcrest-core 2.2 is only one
+    // class named HamcrestCoreIsDeprecated. We force this version to avoid complex exclude and
+    // dependency substitution configuration as there is no conflict with hamcrest in the code base.
+    // Please do not upgrade hamcrest-core.
+    api "org.hamcrest:hamcrest-core:2.2"
     api "org.assertj:assertj-core:3.16.1"
     api "org.mockito:mockito-core:3.3.3"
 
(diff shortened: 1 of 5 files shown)


sda-commons-dependencies/build.gradle
```diff
@@ -135,6 +135,7 @@ dependencies {
 
     // sda-commons-server-testing
     api "junit:junit:4.13"
```
dhenneke, line 137: I wonder if we still need to set this here since it is also part of the dropwizard-bom(?)

dhenneke: Also applies to assertj-core and mockito-core

JoergSiebahn: I'll give it a try and removed all test dependencies that are managed by Dropwizard.

sda-commons-dependencies/build.gradle
```diff
@@ -134,9 +134,12 @@ dependencies {
...
     // sda-commons-server-testing
-    api "junit:junit:4.13"
-    api "org.assertj:assertj-core:3.16.1"
-    api "org.mockito:mockito-core:3.3.3"
+    api "org.hamcrest:hamcrest:2.2"
+    // hamcrest-core is deprecated and overlaps with hamcrest, in hamcrest-core 2.2 is only one
+    // class named HamcrestCoreIsDeprecated. We force this version to avoid complex exclude and
+    // dependency substitution configuration as there is no conflict with hamcrest in the code base.
+    // Please do not upgrade hamcrest-core.
+    api "org.hamcrest:hamcrest-core:2.2"
```
christopher-cudennec, line 142: I used Gradle's "because" syntax to explain certain versions and like that pattern.

PR #366, opened 2020-07-13
Clean up dependency management

JoergSiebahn:


sda-commons-dependencies/build.gradle
```diff
@@ -88,7 +81,12 @@ dependencies {
     api "org.flywaydb:flyway-core:6.5.0"
 
     // sda-commons-server-hibernate-testing
-    api "com.github.database-rider:rider-core:1.2.9" // latest version that supports Parameterized tests: https://github.com/database-rider/database-rider/issues/104
+    api "com.github.database-rider:rider-core:1.2.9", {
+      because "latest version that supports Parameterized tests: https://github.com/database-rider/database-rider/issues/104"
+    }
+    api "org.dbunit:dbunit:2.7.0", {
```
dhenneke, line 87: Does it include security updates? If not I am not sure if this version works well with rider-core (https://github.com/database-rider/database-rider/issues/111#issuecomment-452425777).

sda-commons-dependencies/build.gradle
```diff
@@ -125,6 +123,11 @@ dependencies {
     api "io.prometheus:simpleclient_dropwizard:$prometheusVersion"
     api "io.prometheus:simpleclient_servlet:$prometheusVersion"
 
+    // sda-commons-server-s3
+    api "commons-logging:commons-logging:1.2", {
```
dhenneke, line 127: Does it include security updates or needed features? I am not sure if we want to upgrade a dependency just for the sake of upgrading it.

sda-commons-dependencies/build.gradle
JoergSiebahn: I only checked to provide the same versions as in [master](https://github.com/SDA-SE/sda-dropwizard-commons/blob/master/sda-commons-dependencies/build.gradle#L110).

dhenneke: But that doesn't necessarily mean it made sense before. But I'm fine with leaving it on 2.7.

sda-commons-dependencies/build.gradle
JoergSiebahn: I selected the version that is in master at the moment. I'm not sure if it has a reason. Maybe some compatibility issues with SLF4J. But we can try the version that comes with the aws lib. The only part from the changelog that may be useful is „Improve performance of LogFactory#getFactory()“. But that depends on how and how often the aws lib uses the LogFactory.

PR #397, opened 2020-07-14
fix(deps): prefer using jakarta.* over javax.* dependencies that are provided by dropwizard-dependencies

dhenneke:
There are still some `javax.*` dependencies that are brought in by dropwizard. I guess we want to ignore those for now?
* javax.transaction:javax.transaction-api (by dropwizard-hibernate)
* javax.servlet:javax.servlet-api (by jetty ➡️ by nearly alle dropwizard modules)

### DIFF
--- a/sda-commons-client-jersey-wiremock-testing/build.gradle
+++ b/sda-commons-client-jersey-wiremock-testing/build.gradle
@@ -1,6 +1,10 @@
 dependencies {
 
   compile "com.github.tomakehurst:wiremock-jre8", {
+    /**
+     * Dropwizard comes with jakarta.servlet instead of javax.servlet.
+     * Both contain the same classes.
+     */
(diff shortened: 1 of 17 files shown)


JoergSiebahn:
> There are still some `javax.*` dependencies that are brought in by dropwizard. I guess we want to ignore those for now?

We should keep the dependencies like they come from Dropwizard and avoid managing them ourselves. But we should be aware that Dropwizard may change other javax.* dependencies to jakarta.* to add exclusions to other dependencies that bring javax.* into our dependency tree. I don't know how we get notice of that automatically. Hopefully it is obvious in the Dropwizard changelog.

dhenneke:
👍

christopher-cudennec:
Is this change worth a feat commit?
Consumers of sda-commons might have to adapt their own dependency management.

PR #421, opened 2020-07-22
test(server-morphia): fix test to verify that no personal data is traced

JoergSiebahn:
IntelliJ claimed that `noneMatch(Tags.DB_STATEMENT::equals)` is comparing incompatible types, thanks for that.
Actually it should have been `noneMatch(Tags.DB_STATEMENT.getKey()::equals)`.
But that is the wrong assertion because we do not remove the statement from the trace but replace it with a sanitized version.

This change asserts that we want the statement without the personal information in the trace.

### DIFF
--- a/sda-commons-server-morphia/src/test/java/org/sdase/commons/server/morphia/MorphiaBundleTracingIT.java
+++ b/sda-commons-server-morphia/src/test/java/org/sdase/commons/server/morphia/MorphiaBundleTracingIT.java
@@ -1,6 +1,7 @@
 package org.sdase.commons.server.morphia;
 
 import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.SoftAssertions.assertSoftly;
 
 import dev.morphia.Datastore;
 import io.dropwizard.Application;
@@ -10,7 +11,10 @@ import io.dropwizard.testing.junit.DropwizardAppRule;
 import io.opentracing.mock.MockSpan;
 import io.opentracing.mock.MockTracer;
 import io.opentracing.tag.Tags;
+import java.util.List;
+import java.util.Map;
 import java.util.Set;
+import java.util.stream.Collectors;
 import org.junit.ClassRule;
 import org.junit.Test;
 import org.junit.rules.RuleChain;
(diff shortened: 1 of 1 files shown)


sda-commons-server-morphia/src/test/java/org/sdase/commons/server/morphia/MorphiaBundleTracingIT.java
```diff
@@ -50,12 +54,23 @@ public void shouldHaveInstrumentation() {
...
-        .isTrue();
+    List<Object> dbStatementTags =
+        tracer.finishedSpans().stream()
+            .map(MockSpan::tags)
+            .map(Map::entrySet)
+            .flatMap(Set::stream)
+            .filter(e -> Tags.DB_STATEMENT.getKey().equals(e.getKey()))
+            .map(Map.Entry::getValue)
+            .collect(Collectors.toList());
+    assertSoftly(
```
Fox32, line 65: This changes what the test is testing, this test is about the instrumentation, not about removing personal data. It's sufficient to check whether spans are emitted. See https://github.com/SDA-SE/sda-dropwizard-commons/blob/master/sda-commons-server-morphia/src/test/java/org/sdase/commons/server/morphia/internal/CommandSanitizerTest.java

JoergSiebahn: But I see no integration test that ensures that the sanitizer is applied to the produced spans. I think we have to split the tracing integration test into multiple test methods to cover both, the tracing itself and the absence of personal data.

My assumption from the previous assertion is, that there should be no information about the db statement in the trace. But when I corrected the type error in the assertion the test failed. So the next best guess is, that there should be no personal information in the traced db statement. Another option would be to remove the previous assertion about the db statement. But then we do no test to verify that the sanitizer is applied, do we?

Fox32: I think the condition should be something like:
```assertThat(
            tracer.finishedSpans().stream()
                .map(s -> s.tags().keySet())
                .flatMap(Set::stream)
                .anyMatch(Tags.DB_STATEMENT.getKey()::equals))
        .isTrue();```

But I think a `isNotEmpty` matcher would be better to show that we expect `DB_STATEMENT`s:

```
    assertThat(
            tracer.finishedSpans().stream()
                .map(s -> s.tags().keySet())
                .flatMap(Set::stream)
                .filter(Tags.DB_STATEMENT.getKey()::equals))
        .isNotEmpty();```

JoergSiebahn:
> This changes what the test is testing.

Actually the test was wrong and tested nothing. The assertion itself tried to verify that no instrumentation was made but fails due to incompatible types. Thats why the test was green.
Now the test ensures that the instrumentation works. Additionally it ensures, that personal data is removed in the instrumentation. Why shouldn't we merge an improvement?

JoergSiebahn:
Closing as the effort investing here is not appropriate.

PR #452, opened 2020-08-12
fix(server-hibernate): do not register the HibernateBundle at Jersey

JoergSiebahn:
Registering the bundle causes a warning in the log and has no effect.
The bundle is not context aware.

### DIFF
--- a/sda-commons-server-hibernate/src/main/java/org/sdase/commons/server/hibernate/HibernateBundle.java
+++ b/sda-commons-server-hibernate/src/main/java/org/sdase/commons/server/hibernate/HibernateBundle.java
@@ -78,7 +78,7 @@ public class HibernateBundle<C extends Configuration> implements ConfiguredBundl
   }
 
   public static InitialBuilder builder() {
-    return new Builder();
+    return new Builder<>();
   }
 
   public SessionFactory sessionFactory() {
@@ -93,7 +93,6 @@ public class HibernateBundle<C extends Configuration> implements ConfiguredBundl
   @Override
   public void run(C configuration, Environment environment) throws Exception {
     delegate.run(configuration, environment);
-    environment.jersey().register(this);
   }
 
   //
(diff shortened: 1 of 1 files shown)


sda-commons-server-hibernate/src/main/java/org/sdase/commons/server/hibernate/HibernateBundle.java
```diff
@@ -93,7 +93,6 @@ public void initialize(Bootstrap<?> bootstrap) {
   @Override
   public void run(C configuration, Environment environment) throws Exception {
     delegate.run(configuration, environment);
-    environment.jersey().register(this);
```
christopher-cudennec, line 96: Why was it registered before? Is it necessary for "UnitOfWork", i.e. transactions to work properly?
Might be a change for users of the Hibernate bundle if they now need to register the bundle themselves.

JoergSiebahn: It was there from the beginning without further explanation. We register components at Jersey to be able to use `@Context` in the class or expose endpoints. Is there any other reason why a class or an instance should be registered? We use the `io.dropwizard.hibernate.HibernateBundle` which will register health checks and a `UnitOfWorkApplicationListener` for transaction management.

christopher-cudennec: > Is there any other reason why a class or an instance should be registered? 

Can't say. We hardly use that bundle.

christopher-cudennec: https://www.dropwizard.io/en/latest/manual/hibernate.html

Why do they register DAOs with Jersey?

dhenneke: They register the resource there. Not the DAO.

christopher-cudennec: True. 🤦

PR #453, opened 2020-08-12
feat(server-auth): use HttpUrlConnectorProvider for loading public ke…

falcowehber:
Die SI implementiert zurzeit einen neuen OIDC Provider, welcher außerhalb des internes Netzes zur Verfügung steht.
Die SI Infrastruktur sieht vor, dass Ressourcen außerhalb des internen Netzes über den Proxy angefragt werden müssen.

Zur Verwendung der Java Standard Networking Properties (https.proxyHost, etc. -> https://docs.oracle.com/javase/7/docs/api/java/net/doc-files/net-properties.html) muss eine kleine Anpassung vorgenommen.

Generell ist diese Lösung auch interessant für andere Services (Firebase, Scanner-Service) die von se erstellt werden

### DIFF
--- a/sda-commons-server-auth/src/main/java/org/sdase/commons/server/auth/AuthBundle.java
+++ b/sda-commons-server-auth/src/main/java/org/sdase/commons/server/auth/AuthBundle.java
@@ -11,6 +11,7 @@ import io.dropwizard.setup.Environment;
 import io.opentracing.Tracer;
 import io.opentracing.util.GlobalTracer;
 import javax.ws.rs.client.Client;
+import org.glassfish.jersey.client.HttpUrlConnectorProvider;
 import org.sdase.commons.server.auth.config.AuthConfig;
 import org.sdase.commons.server.auth.config.AuthConfigProvider;
 import org.sdase.commons.server.auth.config.KeyLocation;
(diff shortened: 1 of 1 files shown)


Fox32:
Independent from the change itself: A test verifying that the new solution behaves as expected would be good. Otherwise we can guarantee that the implementation stays this way in the future (because somebody "deletes" the line). Maybe also a comment why it is important.

JoergSiebahn:
A link to the configuration options you mentioned in the PR would be appreciated in the README of the module. Especially for the desired purpose of using a proxy. A test, like @Fox32 asked for, would also help to show how a proxy must be configured.

Unfortunately I tried the same in the [`ClientFactory`](https://github.com/SDA-SE/sda-dropwizard-commons/blob/master/sda-commons-client-jersey/src/main/java/org/sdase/commons/client/jersey/ClientFactory.java#L112) and got failing tests. It would be nice, if proxy configuration works the same for all (HTTP) clients.

dhenneke:
I had a deeper look into the topic and came to this conclusion: It is not wise to use the `HttpUrlConnectorProvider` because it [disables](https://github.com/dropwizard/dropwizard/blob/3a10e6c345caa4948fc7fe6b0221130283e75bad/dropwizard-client/src/main/java/io/dropwizard/client/JerseyClientBuilder.java#L417) all features that Dropwizard introduces with its own `DropwizardApacheConnector`.

But there is another way: We can register the `new SystemDefaultRoutePlanner(ProxySelector.getDefault())`. It does basically the same as the current solution of this PR, but it only adds the proxy functionality. Without overriding any other settings.

To make this more complete: As I tried in #455, we can make the client configurable such that is can be overridden by System Properties. If this is also added to the `JerseyClientBuilder`, we can have proxy settings for each individual client (if the configuration is part of the Application Configuration class). In addition, we can support the `SystemDefaultRoutePlanner` so that the users can decide that they want to globally activate a proxy server. (Note that using the full JerseyClientConfiguration class instead of our limited own one, we can solve other problems that we had in the future, such as making timeouts configurable etc.).

falcowehber:
Request withdrawn

PR #454, opened 2020-08-12
fix(kafka): log a warning if the kafka health check fails

huebner-j:


sda-commons-server-kafka/src/main/java/org/sdase/commons/server/kafka/health/KafkaHealthCheck.java
```diff
@@ -31,6 +31,7 @@ protected Result check() {
     try (AdminClient adminClient = AdminClient.create(KafkaProperties.forAdminClient(config))) {
       adminClient.listTopics().names().get(2, TimeUnit.SECONDS);
     } catch (Exception e) {
+      logger.warn("Kafka health check failed", e);
```
JoergSiebahn, line 34: We should not use the logger from `AdminClientConfig` but an own logger for this class.

huebner-j: oh yes, you are absolutely right. Thanks for your review.

JoergSiebahn:
@SDA-SE/sda-commons-maintainer should we postpone this PR until #303 is merged?

Fox32:
I'm not exactly sure if we want to log that. I remember that we already removed such log messages because they are to chatty. Maybe just put the exception into the message itself?

christopher-cudennec:
> @SDA-SE/sda-commons-maintainer should we postpone this PR until #303 is merged?

+1 

New features should be added after DW2 migration was successful. Looks like #303 will be merged soon, doesn't it?

dhenneke:
I think we could already merge this. But for that, please change the base to `feat/dropwizard2` branch and rebase the commit. The master branch is read-only until #303 will be merged (which is very soon hopefully)!

dhenneke:
We will release the fix when we release DW2 since we currently don’t plan to do a further pre-release. I hope it is not too urgent.

christopher-cudennec:
Thanks @dhenneke . I think that's totally fine! :)

PR #455, opened 2020-08-12
feat(server-auth): add configuration options to the jersey client that is used to load keys

dhenneke:
Adds a property to the `AuthConfig` in order to use all properties of Dropwizard's original [`HttpClientConfiguration`](https://github.com/dropwizard/dropwizard/blob/master/dropwizard-client/src/main/java/io/dropwizard/client/HttpClientConfiguration.java), including Proxy settings.

Dropwizard also provides the option to override configurations via [system properties](https://www.dropwizard.io/en/latest/manual/core.html#configuration). Thus something like
```sh
java -Ddw.auth.keyLoaderClient.proxyConfiguration.host=asdf server my-config.json
```
 should work even if the service doesn't specify the options in the original `config.yml`. This might be an alternative to #453 that is more aligned with the features of Dropwizard.

---

In addition I added an experimental commit that makes our custom `HttpClientConfiguration` extend the `JerseyClientConfiguration`. This would be a breaking change (builder-pattern setters switch to void-setters, though I didn't find any use in our organization). It will be possible to also configure all other clients (as long as the configuration has been added to the application's configuration class). I think this a way better pattern than circumventing our limiting configuration class by custom configurations whenever we need to make timeouts configurable. It also adds out-of-the-box proxy support that can be individually controlled for every HTTP client.

---

Both the `AuthConfig` as well as the `JerseyClientBuilder` can now read the [global java proxy settings](https://docs.oracle.com/javase/7/docs/api/java/net/doc-files/net-properties.html#Proxies). So the application operator can decide whether they want to globally activate a proxy or if they only want to configure it on the client level.

### DIFF
--- a/sda-commons-client-jersey/README.md
+++ b/sda-commons-client-jersey/README.md
@@ -168,10 +168,50 @@ executorService.submit(transferRequestContext(() -> {
 })).get();
 ```
 
+## HTTP Client Configuration and Proxy Support
 
-## Proxy Settings
+Each client can be configured with the standard
+[Dropwizard configuration](https://www.dropwizard.io/en/latest/manual/configuration.html#man-configuration-clients-http).
+Please note that this requires that there is a property in your Application's `Configuration` class. 
 
-All clients consume the standard [proxy system properties](https://docs.oracle.com/javase/7/docs/api/java/net/doc-files/net-properties.html#Proxies).
+```java
+import org.sdase.commons.client.jersey.HttpClientConfiguration;
+
+public class MyConfiguration extends Configuration {
+   private HttpClientConfiguration myClient = new HttpClientConfiguration();
+
+   public HttpClientConfiguration getMyClient() {
+     return myClient;
+   }
+
+   public void setMyClient(HttpClientConfiguration myClient) {
+     this.myClient = myClient;
+   }
+}
+```
+
+```java
+Client client = clientFactory.platformClient(configuration.getMyClient()).buildGenericClient("test");
+```
+
+```yaml
+myClient:
+  timeout: 500ms
+  proxy:
+    host: 192.168.52.11
+    port: 8080
+    scheme : http
+```
+
+> Tip: There is no need to make all configuration properties available as environment variables.
+> Seldomly used properties can always be configured using [System Properties](https://www.dropwizard.io/en/latest/manual/core.html#man-core-configuration).
+
+This configuration can be used to configure a proxy server if needed.
+Use this if all clients should use individual proxy configurations.
+
+In addition, each client can consume the standard [proxy system properties](https://docs.oracle.com/javase/7/docs/api/java/net/doc-files/net-properties.html#Proxies).
+Please note that a specific proxy configuration in the `HttpClientConfiguration` disables the proxy system properties for the client using that configuration.
+This can be helpful when all clients in an Application should use the same proxy configuration (this includes the clients that are used by the [`sda-commons-server-auth` bundle](../sda-commons-server-auth).
 
 ## Tips and Tricks
 
--- a/sda-commons-client-jersey/README.md
+++ b/sda-commons-client-jersey/README.md
@@ -169,6 +169,10 @@ executorService.submit(transferRequestContext(() -> {
 ```
 
 
+## Proxy Settings
+
+All clients consume the standard [proxy system properties](https://docs.oracle.com/javase/7/docs/api/java/net/doc-files/net-properties.html#Proxies).
+
 ## Tips and Tricks
 
 ### 3rd Party `javax.ws.rs-api` Client Implementations in Classpath
--- a/sda-commons-client-jersey/src/main/java/org/sdase/commons/client/jersey/builder/AbstractBaseClientBuilder.java
+++ b/sda-commons-client-jersey/src/main/java/org/sdase/commons/client/jersey/builder/AbstractBaseClientBuilder.java
@@ -2,14 +2,19 @@ package org.sdase.commons.client.jersey.builder;
 
 import static org.sdase.commons.server.opentracing.client.ClientTracingUtil.registerTracing;
 
+import com.fasterxml.jackson.databind.ObjectMapper;
 import io.dropwizard.client.JerseyClientBuilder;
+import io.dropwizard.setup.Environment;
 import io.opentracing.Tracer;
+import java.net.ProxySelector;
 import java.time.Duration;
 import java.util.ArrayList;
 import java.util.List;
 import javax.ws.rs.client.Client;
 import javax.ws.rs.client.ClientRequestFilter;
+import org.apache.http.impl.conn.SystemDefaultRoutePlanner;
 import org.glassfish.jersey.client.ClientProperties;
+import org.sdase.commons.client.jersey.HttpClientConfiguration;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -21,46 +26,39 @@ import org.slf4j.LoggerFactory;
 abstract class AbstractBaseClientBuilder<T extends AbstractBaseClientBuilder> {
 
   private static final Logger LOG = LoggerFactory.getLogger(AbstractBaseClientBuilder.class);
-
-  /**
-   * The default timeout to wait for data in an established connection. 2 seconds is used as a trade
-   * between "fail fast" and "better return late than no result". The timeout may be changed
-   * according to the use case considering how long a user is willing to wait and how long backend
-   * operations need.
-   */
-  private static final int DEFAULT_READ_TIMEOUT_MS = 2_000;
-
-  /**
-   * The default timeout to wait until a connection is established. 500ms should be suitable for all
-   * communication in the platform. Clients that request information from external services may
-   * extend this timeout if foreign services are usually slow.
-   */
-  private static final int DEFAULT_CONNECTION_TIMEOUT_MS = 500;
-
   /**
    * As default, the client will follow redirects so that redirect status codes are automatically
    * resolved by the client.
    */
   private static final boolean DEFAULT_FOLLOW_REDIRECTS = true;
 
+  private HttpClientConfiguration httpClientConfiguration;
   private JerseyClientBuilder jerseyClientBuilder;
   private final Tracer tracer;
+  private final ObjectMapper objectMapper;
 
   private List<ClientRequestFilter> filters;
 
-  private int connectionTimeoutMillis;
+  private Integer connectionTimeoutMillis;
 
-  private int readTimeoutMillis;
+  private Integer readTimeoutMillis;
 
   private boolean followRedirects;
 
-  AbstractBaseClientBuilder(JerseyClientBuilder jerseyClientBuilder, Tracer tracer) {
-    this.jerseyClientBuilder = jerseyClientBuilder;
+  AbstractBaseClientBuilder(
+      Environment environment, HttpClientConfiguration httpClientConfiguration, Tracer tracer) {
+    this.httpClientConfiguration = httpClientConfiguration;
+    this.jerseyClientBuilder = new JerseyClientBuilder(environment);
     this.tracer = tracer;
+    this.objectMapper = environment.getObjectMapper();
     this.filters = new ArrayList<>();
-    this.readTimeoutMillis = DEFAULT_READ_TIMEOUT_MS;
-    this.connectionTimeoutMillis = DEFAULT_CONNECTION_TIMEOUT_MS;
     this.followRedirects = DEFAULT_FOLLOW_REDIRECTS;
+
+    // a specific proxy configuration always overrides the system proxy
+    if (httpClientConfiguration.getProxyConfiguration() == null) {
+      // register a route planner that uses the default proxy variables (e.g. http.proxyHost)
+      this.jerseyClientBuilder.using(new SystemDefaultRoutePlanner(ProxySelector.getDefault()));
+    }
   }
 
   /**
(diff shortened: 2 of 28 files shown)


christopher-cudennec:
Can you provide a test or do tests already exist?

JoergSiebahn:
Just a background information: When the clients where introduced we did not use the full featured configuration from Dropwizard to reduce complexity. At that time we were not aware of the need to configure proxies within the service.

I think we still should avoid to require every service developer to repeat the whole configuration class(es) in the `config.yml`. Setting absent parameters with Java system properties seems an unusual setup in a container environment as you have to overwrite the entrypoint or the command. But it may be suitable to allow configuration of less common settings.

I think it should be possible to introduce the `JerseyClientConfiguration` in a backward compatible manner by overloading the builder methods where we pass the `HttpClientConfiguration`.

dhenneke:
I added more features regarding proxies. We should discuss if we want to add all the functionality as proposed, or use them individually.

Thanks to @falcowehber for the discussions and for bringing up a working solution-prototype.

falcowehber:
it would help us if the auth implementation is already available

sda-commons-client-jersey/src/test/java/org/sdase/commons/client/jersey/ClientProxyTest.java
```diff
@@ -0,0 +1,112 @@
...
+  public static final WireMockClassRule PROXY_WIRE =
+      new WireMockClassRule(wireMockConfig().dynamicPort());
+
+  @Rule
+  public final DropwizardAppRule<ClientTestConfig> dw =
+      new DropwizardAppRule<>(ClientTestApp.class, resourceFilePath("test-config.yaml"));
+
+  @BeforeClass
+  public static void beforeClass() {
+    System.setProperty("http.proxyHost", "localhost");
```
Fox32, line 43: We could extract a `SystemPropertyRule` here

sda-commons-server-auth-testing/src/test/java/org/sdase/commons/server/auth/testing/CustomKeyLoaderConfigIT.java
```diff
@@ -0,0 +1,83 @@
...
+  public static void beforeClass() {
+    WIRE.resetAll();
+    WIRE.stubFor(
+        get(anyUrl())
+            .willReturn(
+                ok().withHeader(CONTENT_TYPE, APPLICATION_JSON).withBody("{\"keys\": []}")));
+  }
+
+  @Test
+  public void shouldBeUnauthorizedIfAccessedWithTokenFromWrongIdp() {
```
Fox32, line 64: Please rename to give context what is tested here!

sda-commons-server-auth/src/main/java/org/sdase/commons/server/auth/config/AuthConfig.java
```diff
@@ -1,11 +1,15 @@
...
 
+import io.dropwizard.client.JerseyClientConfiguration;
 import java.util.ArrayList;
 import java.util.List;
 
 /** Configuration for authentication using JWT. */
 public class AuthConfig {
 
+  /** The client configuration of the HTTP client that is used for the key loader. */
+  private JerseyClientConfiguration keyLoaderClientConfig;
```
Fox32, line 11: I would name it just `keyLoaderClient`

sda-commons-client-jersey/README.md
```diff
@@ -168,6 +168,49 @@ executorService.submit(transferRequestContext(() -> {
...
+```
+
+> Tip: There is no need to make all configuration properties available as environment variables.
+> Seldomly used properties can always be configured using [System Properties](https://www.dropwizard.io/en/latest/manual/core.html#man-core-configuration).
+
+This configuration can be used to configure a proxy server if needed.
+Use this if all clients should use individual proxy configurations.
+
+In addition, each client consumes the standard [proxy system properties](https://docs.oracle.com/javase/7/docs/api/java/net/doc-files/net-properties.html#Proxies).
+This can be helpful when all clients in an Application should use the same proxy configuration (this includes the client that is used by the [`sda-commons-server-auth` bundle](../sda-commons-server-auth).
```
Fox32, line 213: the client => the clients

sda-commons-client-jersey/src/test/java/org/sdase/commons/client/jersey/ClientProxyTest.java
```diff
@@ -0,0 +1,101 @@
...
+  public static final WireMockClassRule PROXY_WIRE =
+      new WireMockClassRule(wireMockConfig().dynamicPort());
+
+  private final SystemPropertyRule systemProperty =
+      new SystemPropertyRule()
+          .setProperty("http.proxyHost", "localhost")
+          .setProperty("http.proxyPort", "" + PROXY_WIRE.port())
+          .setProperty("http.nonProxyHosts", "localhost");
+
+  private final DropwizardAppRule<ClientTestConfig> dw =
```
Fox32, line 44: We could use a classrule here, as we have multiple tests

sda-commons-server-auth/README.md
```diff
@@ -302,8 +327,10 @@ opa:
   # Package name of the policy file that should be evaluated for authorization decision
   # The package name is used to resolve the full path
   policyPackage: http.authz
-  # readTimeout for OPA requests in millis, default 500
-  readTimeout: 500
+  # Advanced configuration of the HTTP client that is used to call the Open Policy Agent
+  opaClient:
+    # timeout for OPA requests, default 500ms
+    timeout: 500ms
```
Fox32, line 333: This?

sda-commons-server-auth/README.md
```diff
@@ -317,6 +344,26 @@ opa:
...
+[Dropwizard configuration](https://www.dropwizard.io/en/latest/manual/configuration.html#man-configuration-clients-http).
+
+> Tip: There is no need to make all configuration properties available as environment variables.
+> Seldomly used properties can always be configured using [System Properties](https://www.dropwizard.io/en/latest/manual/core.html#man-core-configuration).
+> 
+```yaml
+opa:
+  opaClient:
+    httpClient:
+      timeout: 500ms
```
Fox32, line 359: Or This?

sda-commons-server-testing/src/main/java/org/sdase/commons/server/testing/SystemPropertyRule.java
```diff
@@ -0,0 +1,61 @@
...
+
+/**
+ * This {@link TestRule} allows to set system properties for JUnit tests. All system properties will
+ * be reset after the test has run.
+ *
+ * <p>Example:
+ *
+ * <pre>
+ *   class MyTest {
+ *     &#64;ClassRule public static final EnvironmentRule PROP = new SystemPropertyRule().setProperty("DISABLE_JWT", "true");
```
Fox32, line 17: `EnvironmentRule `

sda-commons-server-auth/README.md
dhenneke: this

sda-commons-client-jersey/src/main/java/org/sdase/commons/client/jersey/builder/AbstractBaseClientBuilder.java
```diff
@@ -21,45 +22,32 @@
...
+  private Integer connectionTimeoutMillis;
 
-  private int readTimeoutMillis;
+  private Integer readTimeoutMillis;
 
   private boolean followRedirects;
 
-  AbstractBaseClientBuilder(JerseyClientBuilder jerseyClientBuilder, Tracer tracer) {
+  AbstractBaseClientBuilder(
+      HttpClientConfiguration httpClientConfiguration,
```
JoergSiebahn, line 44: I think we should document that this configuration must be the same instance that is used by the given `JerseyClientBuilder`. Another option would be to apply the configuration in this class after it has been modified. In the current implementation of the `io.dropwizard.client.JerseyClientBuilder` and the `io.dropwizard.client.HttpClientBuilder` that is used internally by Dropwizard the configuration is read and applied when `build(String)` is called. But we should not rely on the use of mutable configuration objects by the framework. If we do because it is working right now, we should also document that [where we modify the instance](https://github.com/SDA-SE/sda-dropwizard-commons/pull/455/files#diff-d74fe01ce7e7a048fc30666449d1bd56R130-R142) (see below).

As we are talking about code that is only used for backward compatibility, the documentation option seems to be the most simple one in the long term.

We may also document the parameter as deprecated. It won't be needed any more when the deprecated timeout setters are removed.

```suggestion
      @Deprecated HttpClientConfiguration httpClientConfiguration,
```

sda-commons-client-jersey/README.md
```diff
@@ -168,6 +168,49 @@ executorService.submit(transferRequestContext(() -> {
...
+```
+
+> Tip: There is no need to make all configuration properties available as environment variables.
+> Seldomly used properties can always be configured using [System Properties](https://www.dropwizard.io/en/latest/manual/core.html#man-core-configuration).
+
+This configuration can be used to configure a proxy server if needed.
+Use this if all clients should use individual proxy configurations.
+
+In addition, each client consumes the standard [proxy system properties](https://docs.oracle.com/javase/7/docs/api/java/net/doc-files/net-properties.html#Proxies).
+This can be helpful when all clients in an Application should use the same proxy configuration (this includes the clients that are used by the [`sda-commons-server-auth` bundle](../sda-commons-server-auth).
```
JoergSiebahn, line 213: Do the specific settings of the `JerseyClientConfiguration` override the global _proxy system settings_? Can we document more clearly, if individual settings and global settings are exclusive or can be used at the same time? Can global settings be used when the proxy setting of a client is defined in the `config.yaml` or is it required to use the individual settings in this case? Do things get complicated with this many options?

sda-commons-server-testing/src/main/java/org/sdase/commons/server/testing/SystemPropertyRule.java
```diff
@@ -0,0 +1,61 @@
...
+ *
+ * <p>Example:
+ *
+ * <pre>
+ *   class MyTest {
+ *     &#64;ClassRule public static final SystemPropertyRule PROP = new SystemPropertyRule().setProperty("DISABLE_JWT", "true");
+ *   }
+ * </pre>
+ */
+public class SystemPropertyRule implements TestRule {
```
JoergSiebahn, line 21: NIT: System Properties can be set in tests using [System Rules](https://stefanbirkner.github.io/system-rules/). We already use them [in some services](https://github.com/search?q=org%3ASDA-SE+stefanbirkner&type=Code) but not here yet. They may also replace our [`EnvironmentRule`](https://github.com/SDA-SE/sda-dropwizard-commons/blob/66e7d99a2bc69e4a39197fa2746f78580677dc15/sda-commons-server-testing/src/main/java/org/sdase/commons/server/testing/EnvironmentRule.java) in the long term.

JoergSiebahn:
Apart from some documentation details I'm happy with the changes. I think we should tackle the new configuration complexity with some clear statements in the documentation and a „best practice“ advice that gives the users all options and keeps the implementation and the default use as simple as possible.

You may also consider to document the deprecations with their alternatives, like [Sonar suggests](https://sonarcloud.io/project/issues?id=SDA-SE_sda-dropwizard-commons&pullRequest=455&resolved=false&severities=MAJOR&types=CODE_SMELL).

sda-commons-client-jersey/src/main/java/org/sdase/commons/client/jersey/builder/AbstractBaseClientBuilder.java
dhenneke: > […] Another option would be to apply the configuration in this class after it has been modified. […]

Yes. It is not necessary anymore that we create the `JerseyClientBuilder` in the `ClientFactory`. I can try to refactor it.

> […] we should not rely on the use of mutable configuration objects by the framework […]

I can also reset it to the previous value in case someone else wants to use the same configuration instance in different builders (could be problematic in terms of thread-safety).

sda-commons-client-jersey/README.md
dhenneke: Global settings override specific settings. (In fact, it seems that the module for global settings _always_ overrides the specific settings). I'll think about how to handle this.

sda-commons-server-testing/src/main/java/org/sdase/commons/server/testing/SystemPropertyRule.java
Fox32: They also provide an `EnvironmentRule`, is it time do deprecate our rule and use this one instead?

Fox32: Nevermind :wink

JoergSiebahn: As I said, we should do that in the long term. But that can be a separate PR and should not block this one in my opinion.

Fox32: On the other hand it looks like this would tie us to junit4, so we should probably not deprecated it and stay with our own implementation

sda-commons-client-jersey/README.md
dhenneke: We could change this to: Specific settings override global settings. We can use the same check logic that Dropwizard uses to check whether the `NonProxyListProxyRoutePlanner` should be used to not add the `SystemDefaultRoutePlanner` in that case.

However, I get problems when one wants to set the non-proxy-hosts via the system property `dw.client.proxy.nonProxyHosts`. Since it is an array, it only works in case the property exists as an array before in the config file, which isn't the wanted behavior. We won't provide the config settings as part of the default `config.yml` of all our services. But I could update the documentation to highlight that it is preferred to use the global proxy settings.

dhenneke: I changed the logic: If there is no specific proxy configuration for a client, it defaults to the default java properties. Service creators should not add any proxy related configuration to their `config.yml`. If needed, this should be done via system properties (would be nice if dropwizard would also support environment variables).

sda-commons-client-jersey/src/main/java/org/sdase/commons/client/jersey/builder/AbstractBaseClientBuilder.java
dhenneke: I refactored the `AbstractBaseClientBuilder` so it creates the `JerseyClientBuilder`. This removes the need to highlight how the builder should have been configured.

JoergSiebahn (approved):
🏃‍♂️

PR #463, opened 2020-08-17
fix(server-auth): don't include the exception class when a token decoding failed

dhenneke:
This PR also adds tests that checks if the error messages are actually the ones that are expected to be returned.


sda-commons-server-auth/src/main/java/org/sdase/commons/server/auth/error/JwtAuthException.java
```diff
@@ -18,6 +18,6 @@ public JwtAuthException(String message, Throwable cause) {
   }
 
   public JwtAuthException(Throwable cause) {
-    super(cause);
+    this(cause.getMessage(), cause);
```
christopher-cudennec, line 21: From my experience the message is often useless. E.g. the message for NullpointerException always is `null`.

PR #519, opened 2020-09-10
chore(deps): add another label to dependabot PRs

JoergSiebahn:
This way it's easier to distinguish updates for the current release and the maintenance branch.

See https://dependabot.com/docs/config-file/#default_labels


christopher-cudennec (approved):
If the default label is "dependencies" as stated in the documentation, I'm fine with the changes.

PR #527, opened 2020-09-16
fix(server-kafka): Fix waiting for topic creation

christopher-cudennec:
The new test `AutocommitStrategyTest.java` mainly consists of the old `MessageListenerTest` of sda-commons 1. I'm not sure why the complete test was deleted because I think many tests still apply to the new `MessageListenerStrategy`.

The code change handles "null" as a result of `KafkaConsumer#partitionsFor` which caused a `NullPointerException`.


JoergSiebahn (approved):
Thx for restoring.

PR #535, opened 2020-09-21
test(server-morphia): resolve warning about comparing incompatible types

Fox32:
Resolves #421

### DIFF
--- a/sda-commons-server-morphia/src/test/java/org/sdase/commons/server/morphia/MorphiaBundleTracingIT.java
+++ b/sda-commons-server-morphia/src/test/java/org/sdase/commons/server/morphia/MorphiaBundleTracingIT.java
@@ -54,8 +54,8 @@ public class MorphiaBundleTracingIT {
             tracer.finishedSpans().stream()
                 .map(s -> s.tags().keySet())
                 .flatMap(Set::stream)
-                .noneMatch(Tags.DB_STATEMENT::equals))
-        .isTrue();
+                .filter(Tags.DB_STATEMENT.getKey()::equals))
+        .isNotEmpty();
   }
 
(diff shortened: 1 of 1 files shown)


sda-commons-server-morphia/src/test/java/org/sdase/commons/server/morphia/MorphiaBundleTracingIT.java
```diff
@@ -54,8 +54,8 @@ public void shouldHaveInstrumentation() {
             tracer.finishedSpans().stream()
                 .map(s -> s.tags().keySet())
                 .flatMap(Set::stream)
-                .noneMatch(Tags.DB_STATEMENT::equals))
-        .isTrue();
+                .filter(Tags.DB_STATEMENT.getKey()::equals))
+        .isNotEmpty();
```
christopher-cudennec, line 58: Before your change the test asserted that none matched the filter expression.
Now it assert that we have at least one match.
Just to confirm: The test was wrong before?
(I did not follow your discussion in the previous PR)

christopher-cudennec: As @JoergSiebahn stated:
> But that is the wrong assertion because we do not remove the statement from the trace but replace it with a sanitized version.

Seems to be ok then! 👍

Fox32: The test was wrong before, even though it still asserted correctly when removing the tracing support 😆

christopher-cudennec (approved):

PR #537, opened 2020-09-21
fix(server-jackson): support nullable query params in the HalLinkProvider

jeskosda:
In some cases it could be required that the QueryParam is nullable. Maybe you want to link to a specific set of entities which are prefiltered?

```
    URI uri = linkTo(methodOn(MedicalHistoryService.class).getEvents("12345", null, null, null, null, null, null)).asUri();
    assertThat(uri).hasPath("/medicalEvents").hasParameter("insuredPersonId", "12345");
    assertThat(uri.toString()).isEqualTo("/medicalEvents?insuredPersonId=12345");
```


dhenneke:
`fix(server-jackson): support nullable query params in the HALLinkProvider`?

jeskosda:
Changed the commit message. I was not aware of the right message pattern.

dhenneke:
We usually use the last part of the module name that is changed as scope. In this case sda-commons-**server-jackson**.

Fox32:
> I was not aware of the right message pattern.

See the [contributing guidelines](https://github.com/SDA-SE/sda-dropwizard-commons/blob/master/CONTRIBUTING.md) with some examples.

dhenneke (approved):

PR #548, opened 2020-09-28
feat(server-kafka): allow the configuration of the sasl.mechanism

dhenneke:
This PR adds support for the `SCRAM-SHA-512` SASL-mechanism.

The changes:
1. Automatically select the correct login module for the SASL mechanisms (`PLAIN`-> `PlainLoginModule`, `SCRAM-SHA-*` -> `ScramLoginModule`). Other mechanisms lead to a failing bundle. Those should be used with the `config` overrides or an sda-dropwizard-commons PR.
2. The `AdminConfig` lacked a `config` property to provide custom configurations if needed.
3. Don't require username or password to configure the security protocol. `SSL` doesn't need credentials, but it wasn't obvious that one still need to fill the username and password fields before.
4. A new Readme-section that describes the different auth-configurations.
5. A new test that checks if the bundle can connect to a `SASL_PLAINTEXT` / `PLAIN` Kafka broker.
6. A new test that checks if the bundle can connect to a `SASL_PLAINTEXT` / `SCRAM-SHA-512` Kafka broker.

Plus: I increased the log level for the tests because I hope that it helps to find some strange errors in flaky tests in the future.

### DIFF
--- a/sda-commons-server-kafka/README.md
+++ b/sda-commons-server-kafka/README.md
@@ -11,19 +11,18 @@ implements a polling loop for Kafka consumers. The user of this bundle must only
 
 ## Usage
 Add the following dependency:
-```
+```groovy
 compile 'org.sdase.commons:sda-commons-server-kafka:<current-version>'
 ```
 
 **Bootstrap**
 
-The bundle got enhanced to allow more control and flexibility how Kafka messages are consumed and which commit strategy is used. How to use
-the old and now deprecated `KafkaBundle::registerMessageHandler` approach is documented [here](docs/deprecated.md).
+The bundle got enhanced to allow more control and flexibility how Kafka messages are consumed and which commit strategy is used.
      
 The bundle should be added as field to the application since it provides methods for the creation of `MessageProducer` and `MessageListener`.
 The Builders for `MessageListenerRegistration` and `ProducerRegistration` supports the user in the creation of these complex configurable objects. 
  
-```
+```java
 public class DemoApplication {
    private final KafkaBundle<AppConfiguration> kafkaBundle = KafkaBundle.builder().withConfigurationProvider(AppConfiguration::getKafka).build();
    private final MessageProducer<String, ProductBundle> producer;
@@ -133,7 +132,7 @@ But be aware that the `key` or `value` can be `null` in this case in both `Messa
 Another alternative is to implement your own Deserializer to have even more control and where you can potentially apply any fallback deserialization
 strategy. 
 
-```
+```java
 public class DemoApplication {
    private final KafkaBundle<AppConfiguration> kafkaBundle = KafkaBundle.builder().withConfigurationProvider(AppConfiguration::getKafka).build();
    private final MessageProducer<String, ProductBundle> producer;
@@ -184,7 +183,7 @@ kafka:
       user: user
       password: password
       protocol: SASL_SSL
-    
+
   # List of brokers to bootstrap consumer and provider connection
   brokers:
     - kafka-broker-1:9092 
@@ -195,6 +194,7 @@ kafka:
     user: user
     password: password
     protocol: SASL_SSL
+    saslMechanism: PLAIN
 
   # Additional configuration properties that are added to all consumers, producers, and the admin client
   # configuration key -> values as defined in the kafka documentation
@@ -255,6 +255,95 @@ kafka:
       topicMissingRetryMs: 60000
       # Milliseconds to sleep between two poll intervals if no messages are available
       pollInterval: 200
+```
+
+### Security Settings
+
+There are different configuration options to connect to a Kafka Broker.
+
+#### PLAINTEXT
+
+The server uses an unencrypted connection with no authentication. 
+
+```yaml
+  security :
+    protocol: PLAINTEXT # can be omitted, default value
+```
+
+#### SSL
+
+The server uses an encrypted connection with no authentication. 
+
+```yaml
+  security :
+    protocol: SSL
+```
+
+#### SASL_PLAINTEXT
+
+The server uses an unencrypted connection with `PLAIN` authentication. 
+
+```yaml
+  security :
+    protocol: SASL_PLAINTEXT
+    saslMechanism: PLAIN # can be omitted, default value when username and password are specified
+    user: user
+    password: password
+```
+
+#### SASL_SSL
+
+The server uses an encrypted connection with `PLAIN` authentication. 
+
+```yaml
+  security :
+    protocol: SASL_SSL
+    saslMechanism: PLAIN # can be omitted, default value when username and password are specified
+    user: user
+    password: password
+```
+
+#### Other SASL mechanisms
+
+Beside `sasl.mechanism: PLAIN`, the bundle also supports `SCRAM-SHA-256` and `SCRAM-SHA-512`.
+All mechanisms can be used with both `SASL_PLAINTEXT` and `SASL_SSL`.
+
+```yaml
+  security :
+    protocol: SASL_PLAINTEXT # or SASL_SSL
+    saslMechanism: SCRAM-SHA-512 # or SCRAM-SHA-256
+    user: user
+    password: password
+```
+
+Other mechanisms can be configured by overriding the config properties.
+Note that the properties can also be configured individually for each consumer, each producer, and the admin client.
+
+```yaml
+  config:
+    sasl.mechanism: OTHER-VALUE
+    sasl.jaas.config: "org.apache.kafka.common.security.scram.ScramLoginModule required username='youruser' password='yourpassword';"
+```
+
+OR
+
+```yaml
+  consumers:
+    yourConsumer:
+      config:
+        sasl.mechanism: OTHER-VALUE
+        sasl.jaas.config: "org.apache.kafka.common.security.scram.ScramLoginModule required username='youruser' password='yourpassword';"
+
+  producers:
+    yourProducer:
+      config:
+        sasl.mechanism: OTHER-VALUE
+        sasl.jaas.config: "org.apache.kafka.common.security.scram.ScramLoginModule required username='youruser' password='yourpassword';"
+
+  adminConfig:
+    config:
+      sasl.mechanism: OTHER-VALUE
+      sasl.jaas.config: "org.apache.kafka.common.security.scram.ScramLoginModule required username='youruser' password='yourpassword';"
 ```
 
 ### Configuration value defaults (extending/changing the Kafka defaults)
--- a/sda-commons-server-kafka/README.md
+++ b/sda-commons-server-kafka/README.md
@@ -195,6 +195,12 @@ kafka:
     user: user
     password: password
     protocol: SASL_SSL
+
+  # Additional configuration properties that are added to all consumers, producers, and the admin client
+  # configuration key -> values as defined in the kafka documentation
+  config:
+    ssl.truststore.location: /my/truststore/location.jks
+
   # Map with consumer configurations. The key is used as name/id to address the configuration within the code. 
   consumers:
     # id/name of the consumer configuration
--- a/sda-commons-server-kafka/src/main/java/org/sdase/commons/server/kafka/KafkaProperties.java
+++ b/sda-commons-server-kafka/src/main/java/org/sdase/commons/server/kafka/KafkaProperties.java
@@ -86,6 +86,7 @@ public class KafkaProperties extends Properties {
     props.put(
         AdminClientConfig.REQUEST_TIMEOUT_MS_CONFIG,
         configuration.getAdminConfig().getAdminClientRequestTimeoutMs());
+    props.putAll(configuration.getAdminConfig().getConfig());
     return props;
   }
 
--- a/sda-commons-server-kafka/src/main/java/org/sdase/commons/server/kafka/KafkaProperties.java
+++ b/sda-commons-server-kafka/src/main/java/org/sdase/commons/server/kafka/KafkaProperties.java
@@ -6,8 +6,11 @@ import org.apache.kafka.clients.CommonClientConfigs;
 import org.apache.kafka.clients.admin.AdminClientConfig;
 import org.apache.kafka.clients.consumer.ConsumerConfig;
 import org.apache.kafka.clients.producer.ProducerConfig;
+import org.apache.kafka.common.security.plain.PlainLoginModule;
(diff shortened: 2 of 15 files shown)


sda-commons-server-kafka/src/main/java/org/sdase/commons/server/kafka/KafkaProperties.java
```diff
@@ -112,4 +118,16 @@ public static KafkaProperties forProducer(KafkaConfiguration configuration) {
 
     return props;
   }
+
+  private static Class<?> getLoginModule(String saslMechanism) {
+    switch (saslMechanism) {
```
christopher-cudennec, line 123: Might be nicer to add `toUpperCase` here.

sda-commons-server-kafka/src/test/java/com/salesforce/kafka/test/SharedKafkaTestResourceScram.java
```diff
@@ -0,0 +1,100 @@
...
+      try {
+        Field f = KafkaTestCluster.class.getDeclaredField("zkTestServer");
+
+        // the field is private, so we need to make it accessible
+        f.setAccessible(true);
+
+        // the field is final, so we need to remove the modifier
+        Field modifiersField = Field.class.getDeclaredField("modifiers");
+        modifiersField.setAccessible(true);
+        modifiersField.setInt(f, f.getModifiers() & ~Modifier.FINAL);
```
christopher-cudennec, line 37: Do we have a test helper class for these kind of modifications? I imagine we need this trick in a couple of places.

sda-commons-server-kafka/src/test/java/com/salesforce/kafka/test/SharedKafkaTestResourceScram.java
```diff
@@ -0,0 +1,100 @@
...
+import kafka.admin.ConfigCommand;
+import org.junit.runner.Description;
+import org.junit.runners.model.Statement;
+
+/**
+ * Creates and starts up a {@link SharedKafkaTestResource} that uses a custom {@link
+ * SharedZookeeperTestResource} with installed SCRAM-SHA-* credentials.
+ *
+ * <p>See {@link SharedKafkaTestResource} for usage.
+ */
```
christopher-cudennec, line 18: Have you tried contributing the class to the original test library?

sda-commons-server-kafka/src/test/java/org/sdase/commons/server/kafka/config/KafkaPropertiesSpec.java
```diff
@@ -26,4 +27,82 @@ public void itShouldBuildSaslStringCorrectly() {
         "org.apache.kafka.common.security.plain.PlainLoginModule required username='user' password='password';";
     assertThat(props.getProperty("sasl.jaas.config")).isEqualTo(saslJaasConfig);
   }
+
+  @Test
+  public void itShouldConfigureSecurityForAdminConfig() {
```
christopher-cudennec, line 32: Will classes ending with "Spec" be picked up by Junit at all?

sda-commons-server-kafka/src/test/java/com/salesforce/kafka/test/SharedKafkaTestResourceScram.java
dhenneke: Not yet. It would be nice if we integrate this into the original library since we could then get rid of the quite hacky code in SharedKafkaTestResourceScram.java. But we should do this independent from this PR.

sda-commons-server-kafka/src/test/java/org/sdase/commons/server/kafka/config/KafkaPropertiesSpec.java
dhenneke: Yes. JUnit is only interested in the `@Test` Though I agree that the name of the file could be improved to match our usual pattern.

sda-commons-server-kafka/src/test/java/com/salesforce/kafka/test/SharedKafkaTestResourceScram.java
dhenneke: I'm not sure. I guess we shouldn't make it too easy to do these kinds of things.

sda-commons-server-kafka/src/main/java/org/sdase/commons/server/kafka/KafkaProperties.java
dhenneke: Yes. That might prevent potential frustrations.

.github/workflows/gradle.yml
```diff
@@ -32,7 +32,7 @@ jobs:
           java-version: ${{ matrix.java_version }}
 
       - name: Test with Gradle
-        run: ./gradlew --parallel --continue javadoc spotlessCheck check
+        run: ./gradlew --parallel --continue javadoc spotlessCheck check --info
```
JoergSiebahn, line 35: Is it intended to keep the `--info` flag? The build log is quite large right now.

sda-commons-server-kafka/README.md
```diff
@@ -251,6 +252,87 @@ kafka:
...
+
+```yaml
+  consumers:
+    yourConsumer:
+      config:
+        sasl.mechanism: OTHER-VALUE
+        sasl.jaas.config: "org.apache.kafka.common.security.scram.ScramLoginModule required username='youruser' password='yourpassword';"
+
+  producers:
+    yourPrpducer:
```
JoergSiebahn, line 325: Typo
```suggestion
    yourProducer:
```

sda-commons-server-kafka/README.md
```diff
@@ -251,6 +252,87 @@ kafka:
       pollInterval: 200
 ```
 
+### Security Settings
```
JoergSiebahn, line 255: I like this comprehensive documentation about the authentication settings 👍

sda-commons-server-kafka/src/main/java/org/sdase/commons/server/kafka/KafkaProperties.java
```diff
@@ -18,6 +21,33 @@ private KafkaProperties() {
...
+      props.put("sasl.mechanism", security.getSaslMechanism());
+      props.put(
+          "sasl.jaas.config",
+          getLoginModule(security.getSaslMechanism())
+              .getName()
+              .concat(" required username='")
+              .concat(security.getUser())
+              .concat("' password='")
+              .concat(security.getPassword())
+              .concat("';"));
```
JoergSiebahn, line 45: NIT: Although this is the same as before, I think this is one of the rare cases where even I think, `String.format` is the most readable option. Personally I would prefer `String.format` over the `+` operator over `concat` in this specific case.

sda-commons-server-kafka/src/main/java/org/sdase/commons/server/kafka/KafkaProperties.java
```diff
@@ -18,6 +21,33 @@ private KafkaProperties() {
...
+    }
+
+    if (security.getPassword() != null && security.getUser() != null) {
+      props.put("sasl.mechanism", security.getSaslMechanism());
+      props.put(
+          "sasl.jaas.config",
+          getLoginModule(security.getSaslMechanism())
+              .getName()
+              .concat(" required username='")
+              .concat(security.getUser())
```
JoergSiebahn, line 42: NIT: Same as before, but don't we have to escape username and password here?

sda-commons-server-kafka/src/main/java/org/sdase/commons/server/kafka/KafkaProperties.java
dhenneke: Yes. `String.format` would be perfect here.

sda-commons-server-kafka/src/test/java/com/salesforce/kafka/test/SharedKafkaTestResourceScram.java
```diff
@@ -0,0 +1,100 @@
...
+
+        // the field is final, so we need to remove the modifier
+        Field modifiersField = Field.class.getDeclaredField("modifiers");
+        modifiersField.setAccessible(true);
+        modifiersField.setInt(f, f.getModifiers() & ~Modifier.FINAL);
+
+        // use the already started (and configured) zookeeper server.
+        f.set(kafkaCluster, sharedZookeeperTestResource.getZookeeperTestServer());
+      } catch (NoSuchFieldException | IllegalAccessException e) {
+        // empty
```
JoergSiebahn, line 42: I think we should rethrow the exceptions as RuntimeException here with an informational message so that we know when the implementation of the modified `KafkaTestCluster` changes.

sda-commons-server-kafka/src/main/java/org/sdase/commons/server/kafka/KafkaProperties.java
dhenneke: I'm not sure what you mean with "escape" here.

sda-commons-server-kafka/src/test/java/com/salesforce/kafka/test/SharedKafkaTestResourceScram.java
dhenneke: I guess when they change the implementation the Kafka wouldn't start anymore since the Zookeeper is configured invalid. But we can rethrow them here.

In the long run it would also be nice to get this contributed to kafka-junit, so this workaround would be unnecessary. But for now it is easier to keep it that way.

sda-commons-server-kafka/src/test/java/com/salesforce/kafka/test/listeners/SaslScramListener.java
```diff
@@ -0,0 +1,123 @@
...
+    return "SASL_PLAINTEXT";
+  }
+
+  @Override
+  public Properties getBrokerProperties() {
+    final Properties properties = new Properties();
+    properties.put("sasl.enabled.mechanisms", "SCRAM-SHA-256,SCRAM-SHA-512");
+    properties.put("sasl.mechanism.inter.broker.protocol", "SCRAM-SHA-512");
+
+    // properties.put("inter.broker.listener.name", "SASL_PLAINTEXT");
```
JoergSiebahn, line 100: Can we remove the commented code or add an advice when to use it?

dhenneke: No sure what this is for. It came from here and I didn't removed it (https://github.com/salesforce/kafka-junit/blob/9675c18b5221a156821d534129b9f03e6b6779b3/kafka-junit-core/src/main/java/com/salesforce/kafka/test/listeners/SaslPlainListener.java#L90).

I'll remove it.

sda-commons-server-kafka/src/main/java/org/sdase/commons/server/kafka/KafkaProperties.java
JoergSiebahn: I'm not sure what happens if the password (or less likely the username) contains `'`. So it may be required to convert a password like `my'secret` into `my\'secret` or anything else how strings have to be escaped here. But we can postpone that discussion and wait until a problem is reported.

sda-commons-server-kafka/src/main/java/org/sdase/commons/server/kafka/config/AdminConfig.java
```diff
@@ -11,6 +13,8 @@
 
   private Security adminSecurity = new Security();
 
+  private Map<String, String> config = new HashMap<>();
```
JoergSiebahn, line 16: Does the general `KafkaConfiguration` need a flexible `config` map as well?

dhenneke: Yes. This could also obsolete this addition in the `AdminConfig`. We will need for the SSL stuff anyway (#589) so it might be a good idea to already add it here. This also simplifies the part with the custom SASL mechanism in the Readme.

sda-commons-server-kafka/src/test/java/org/sdase/commons/server/kafka/config/KafkaPropertiesTest.java
```diff
@@ -0,0 +1,133 @@
...
+public class KafkaPropertiesTest {
+  @Test
+  public void itShouldUseGlobalConfig() {
+    KafkaConfiguration config = new KafkaConfiguration();
+
+    config.getConfig().put("setting", "from.global");
+
+    KafkaProperties.forProducer(config).get("from.global");
+    KafkaProperties.forConsumer(config).get("from.global");
+    KafkaProperties.forAdminClient(config).get("from.global");
```
JoergSiebahn, line 20: To avoid the Sonar complaint:
```suggestion
    assertThat(KafkaProperties.forProducer(config).get("setting")).isEqualTo("from.global");
    assertThat(KafkaProperties.forConsumer(config).get("setting")).isEqualTo("from.global");
    assertThat(KafkaProperties.forAdminClient(config).get("setting")).isEqualTo("from.global");
```

sda-commons-server-kafka/src/test/java/org/sdase/commons/server/kafka/config/KafkaPropertiesTest.java
```diff
@@ -0,0 +1,133 @@
...
+
+    AdminConfig adminConfig = new AdminConfig();
+    adminConfig.getConfig().put("setting", "from.admin");
+    config.setAdminConfig(adminConfig);
+
+    config.getConfig().put("setting", "from.global");
+
+    KafkaProperties.forProducer(config).get("from.global");
+    KafkaProperties.forConsumer(config).get("from.global");
+    KafkaProperties.forAdminClient(config).get("from.admin");
```
JoergSiebahn, line 35: To avoid the Sonar complaint:

```suggestion
    assertThat(KafkaProperties.forProducer(config).get("setting")).isEqualTo("from.global");
    assertThat(KafkaProperties.forConsumer(config).get("setting")).isEqualTo("from.global");
    assertThat(KafkaProperties.forAdminClient(config).get("setting")).isEqualTo("from.admin");
```

JoergSiebahn (approved):
Thanks for the new auth options and the refactoring of the Kafka configuration.

PR #589, opened 2020-10-20
docs(server-kafka): document the use of custom SSL certificates

dhenneke:


sda-commons-server-kafka/src/test/resources/ssl/create-ca.txt
```diff
@@ -0,0 +1,20 @@
+# create CA
```
Fox32, line 1: Could this be an `.sh` file?

sda-commons-server-kafka/README.md
```diff
@@ -346,6 +346,39 @@ OR
       sasl.jaas.config: "org.apache.kafka.common.security.scram.ScramLoginModule required username='youruser' password='yourpassword';"
 ```
 
+#### Custom Certificate Authority and Client Certificates
+
+`SSL` or `SASL_SSL` can also use Kafka brokers that have a self-signed or private-CA certificate. 
+Use the Java-default system properties `javax.net.ssl.trustStore` and `javax.net.ssl.trustStorePassword` to provide the certificates (see [KafkaBundleWithSslTruststoreIT](./src/test/java/org/sdase/commons/server/kafka/KafkaBundleWithSslTruststoreIT.java)).
+
+For more control, configure the truststore only for the Kafka bundle and not for the complete JVM (see [KafkaBundleWithSslIT](./src/test/java/org/sdase/commons/server/kafka/KafkaBundleWithSslIT.java)):
```
Fox32, line 354: `KafkaBundleWithSslTruststoreIT` and `KafkaBundleWithSslIT`

sda-commons-server-kafka/src/test/resources/ssl/create-ca.txt
Fox32: Maybe also a short comment that this file was used to generate the example files and is here for documentation?

sda-commons-server-kafka/README.md
```diff
@@ -346,6 +346,39 @@ OR
...
+
+This configuration option also supports providing client certificates in a custom keystore:
+
+```yaml
+  security :
+    protocol: SSL
+
+  config:
+    # configure the truststore
+    ssl.truststore.location: /your/truststore/location.jks
```
christopher-cudennec, line 373: Will we need an environment variable for this value in practice? If yes: Would be nice to change the example accordingly.

Fox32: I think we should make every team think about environment variables here. They won't do it anyway... If such a complex setup is required if would go for system properties

dhenneke: No. I would expect that these kinds of settings are always configured via system properties (`-Ddw.kafka.config.ssl\.truststore\.location`). But I don't want to repeat this option for each and every settings. We need a central readme section that tells that this is an option and where we describe how we expect the `config.yml` to look like.

sda-commons-server-kafka/src/test/resources/ssl/create-ca.txt
dhenneke: 👍

Fox32 (approved):
👍

PR #590, opened 2020-10-20
feat(server-dropwizard): support configuration overrides from environment variables

dhenneke:
Dropwizard supports configuration overrides by system properties (`-D dw.my.property=new-value`). This change supports the same name schema in environment variables (`export dw.my.property=new-value`).

### DIFF
--- a/sda-commons-server-dropwizard/README.md
+++ b/sda-commons-server-dropwizard/README.md
@@ -74,7 +74,7 @@ The bundle sets the log threshold for the console appender to `INFO` and uses th
 Make sure to add the bundle **after the `ConfigurationSubstitutionBundle`** if it's present.
 Logging related configuration is not required by this bundle. 
 
-```
+```java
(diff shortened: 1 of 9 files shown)


christopher-cudennec:
I would close this PR since env variables are not really recommended from Java 16+. Okay for you @JoergSiebahn?

christopher-cudennec:
As discussed: Maybe we can support a scheme like `DW_MY_PROPERTY` to support magically overriding properties instead of using a `.`.

christopher-cudennec:
Leaving one more comment / thought here: Might be a good idea to bring our Dropwizard Commons closer to the configuration possibilities of Spring applications that are based on [sda-spring-boot-commons](https://github.com/SDA-SE/sda-spring-boot-commons). This would probably mean:
* overrides would not need a prefix
* environment variables are translated from "screaming snake case" (`MY_VAR`) to "dot notation" (`my.var`)
* overrides are possible either using system properties or environment variables
* ...

PR #601, opened 2020-10-26
Make Jaeger support optional

dheid:
We experience ICMP port problems in our environments and currently don't want our applications to use Jaeger. So I added a functionality to disable Jaeger.

### DIFF
--- a/sda-commons-server-starter/src/main/java/org/sdase/commons/server/starter/SdaPlatformBundle.java
+++ b/sda-commons-server-starter/src/main/java/org/sdase/commons/server/starter/SdaPlatformBundle.java
@@ -55,6 +55,7 @@ public class SdaPlatformBundle<C extends Configuration> implements ConfiguredBun
   private CorsBundle.FinalBuilder<C> corsBundleBuilder;
   private ConsumerTokenBundle.FinalBuilder<C> consumerTokenBundleBuilder;
   private SwaggerBundle.FinalBuilder swaggerBundleBuilder;
+  private JaegerBundle.Builder jaegerBundleBuilder;
 
   private SdaPlatformBundle(
       SecurityBundle.Builder securityBundleBuilder,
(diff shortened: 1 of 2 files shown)


Fox32:
Hello Daniel, thanks for your PR!

We believe that configuring Jaeger is a run/deployment time configuration and should be available using environment variables. Disabling it in the `SdaPlatformBundle` feels wrong, as that bundle should serve as a recommend default / golden path for service development at the SDA SE. While it is often easy to disable it with one line of code, it can be pretty hard for a operations team to revert that changes later on as they aren't the code owner. All our services should have similar behavior when it comes to operations. So we believe that we keep it activated as a default and configure it instead if required. If teams don't or can't want to support, for whatever reason, Jaeger at all, we recommend not to use the `SdaPlatformBundle` as it would mean to leave the golden path anyway.

The first idea that comes into mind here, would be a `JAEGER_DISABLED` environment variable that can disable it at run time. We already have similar variables for other components. Jaeger implementations of other libraries already have such a environment variable, but the Java one doesn't. But Jaeger has a strong opinion on that:

> JAEGER_DISABLED (not recommended) when set to true, instructs the Configuration to return a no-op tracer. Neither 
> trace/span IDs nor baggage will be propagated with a no-op tracer. Instead of disabling the tracer completely, use const  
> sampler with 0 param, which will minimize the overhead but keep the propagation going.
> https://www.jaegertracing.io/docs/1.19/client-features/

As it is already possible, I don't see a need for another implementation. Please use the [sampler configuration](https://www.jaegertracing.io/docs/1.17/sampling/#client-sampling-configuration) `JAEGER_SAMPLER_TYPE=const` and `JAEGER_SAMPLER_PARAM=0`. However, we have an issue here with lacking documentation. Therefore I created #602 to improve the documentation and give teams a hint how to solve the warning messages.

For future contributions please take a look at [CONTRIBUTING](https://github.com/SDA-SE/sda-dropwizard-commons/blob/master/CONTRIBUTING.md) first.

dheid:
@Fox32 Danke für die sehr gute Antwort. Das hat für uns funktioniert!

PR #604, opened 2020-10-27
feat: allow disabling of kafka health check

eikenilsknopp:
An alternativ solution could be a field in the kafka configuration yml instead of `withoutHealthCheck()` in the builder.

### DIFF
--- a/sda-commons-server-kafka/README.md
+++ b/sda-commons-server-kafka/README.md
@@ -363,6 +363,23 @@ kafka:
       pollInterval: 200
 ```
 
+You can disable the  health check manually if Kafka is not essential for the functionality of your service,
+e.g. it's only used to invalidate a cache, notify about updates or other minor tasks that could fail without
+affecting the service so much it's no longer providing a viable functionality.
+
+This way your service can stay healthy even if the connection to Kafka is disrupted.
+
+```java
+private KafkaBundle<KafkaTestConfiguration> bundle =
+  KafkaBundle.builder()
+    .withConfigurationProvider(KafkaTestConfiguration::getKafka)
+    .withoutHealthCheck() // disable kafkas health check
+    .build();
+```
+Keep in mind that in this case a producer might fail if the broker is not available, so depending on the
+use case the producer should do appropriate error handling e.g. storing unprocessed messages in a queue until the broker is available again.
+If no error handling is done you might be better off not disabling the health check.
+
 ### Security Settings
 
 There are different configuration options to connect to a Kafka Broker.
--- a/sda-commons-server-kafka/src/main/java/org/sdase/commons/server/kafka/KafkaBundle.java
+++ b/sda-commons-server-kafka/src/main/java/org/sdase/commons/server/kafka/KafkaBundle.java
@@ -59,8 +59,11 @@ public class KafkaBundle<C extends Configuration> implements ConfiguredBundle<C>
 
   private static final Logger LOGGER = LoggerFactory.getLogger(KafkaBundle.class);
 
+  public static final String HEALTHCHECK_NAME = "kafkaConnection";
+
   private final Function<C, KafkaConfiguration> configurationProvider;
   private KafkaConfiguration kafkaConfiguration;
+  private boolean healthCheckDisabled;
 
   private ProducerTopicMessageCounter topicProducerCounterSpec;
   private ConsumerTopicMessageHistogram topicConsumerHistogram;
@@ -71,8 +74,10 @@ public class KafkaBundle<C extends Configuration> implements ConfiguredBundle<C>
 
   private Map<String, ExpectedTopicConfiguration> topics = new HashMap<>();
 
-  private KafkaBundle(KafkaConfigurationProvider<C> configurationProvider) {
+  private KafkaBundle(
+      KafkaConfigurationProvider<C> configurationProvider, boolean healthCheckDisabled) {
     this.configurationProvider = configurationProvider;
+    this.healthCheckDisabled = healthCheckDisabled;
   }
 
   public static InitialBuilder builder() {
@@ -89,10 +94,10 @@ public class KafkaBundle<C extends Configuration> implements ConfiguredBundle<C>
   public void run(C configuration, Environment environment) {
     kafkaConfiguration = configurationProvider.apply(configuration);
     kafkaConfiguration.getTopics().forEach((k, v) -> topics.put(k, createTopicDescription(v)));
-    if (!kafkaConfiguration.isDisabled()) {
+    if (!kafkaConfiguration.isDisabled() && !healthCheckDisabled) {
       environment
           .healthChecks()
-          .register("kafkaConnection", new KafkaHealthCheck(kafkaConfiguration));
+          .register(HEALTHCHECK_NAME, new KafkaHealthCheck(kafkaConfiguration));
     }
     topicProducerCounterSpec = new ProducerTopicMessageCounter();
     topicConsumerHistogram = new ConsumerTopicMessageHistogram();
@@ -593,12 +598,23 @@ public class KafkaBundle<C extends Configuration> implements ConfiguredBundle<C>
   }
 
   public interface FinalBuilder<T extends Configuration> {
+
+    /**
+     * Disables the health check for Kafka. By disabling the health check the service can stay
+     * healthy even if the connection to Kafka is disrupted, if Kafka is not essential to the
+     * functionality.
+     *
+     * @return the same builder instance
+     */
+    FinalBuilder<T> withoutHealthCheck();
+
     KafkaBundle<T> build();
   }
 
   public static class Builder<T extends Configuration> implements InitialBuilder, FinalBuilder<T> {
 
     private KafkaConfigurationProvider<T> configurationProvider;
+    private boolean healthCheckDisabled = false;
 
     private Builder() {}
 
@@ -607,8 +623,14 @@ public class KafkaBundle<C extends Configuration> implements ConfiguredBundle<C>
     }
 
     @Override
+    public FinalBuilder<T> withoutHealthCheck() {
+      healthCheckDisabled = true;
+      return this;
+    }
+
+    @Override
     public KafkaBundle<T> build() {
-      return new KafkaBundle<>(configurationProvider);
+      return new KafkaBundle<>(configurationProvider, healthCheckDisabled);
     }
 
     @Override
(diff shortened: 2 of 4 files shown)


sda-commons-server-kafka/src/test/java/org/sdase/commons/server/kafka/dropwizard/KafkaWithoutHealthCheckTestApplication.java
```diff
@@ -0,0 +1,38 @@
...
+import org.sdase.commons.server.dropwizard.bundles.ConfigurationSubstitutionBundle;
+import org.sdase.commons.server.kafka.KafkaBundle;
+
+public class KafkaWithoutHealthCheckTestApplication extends Application<KafkaTestConfiguration> {
+
+  private KafkaBundle<KafkaTestConfiguration> bundle =
+      KafkaBundle.builder()
+          .withConfigurationProvider(KafkaTestConfiguration::getKafka)
+          .withoutHealthCheck()
+          .build();
```
eikenilsknopp, line 16: Example of how the method is used.

sda-commons-server-kafka/src/test/java/org/sdase/commons/server/kafka/health/KafkaWithoutHealthCheckIT.java
```diff
@@ -0,0 +1,120 @@
...
+
+  @Before
+  public void before() {
+    app = DROPWIZARD_APP_RULE.getRule().getApplication();
+  }
+
+  @Test
+  public void healthCheckShouldNotContainKafka() {
+    SortedSet<String> checks = app.healthCheckRegistry().getNames();
+    assertThat(checks).isNotEmpty().doesNotContain("kafkaConnection");
```
christopher-cudennec, line 118: I'd prefer using a constant defined in `KafkaBundle`.

sda-commons-server-kafka/src/test/java/org/sdase/commons/server/kafka/health/KafkaWithoutHealthCheckIT.java
```diff
@@ -0,0 +1,120 @@
...
+              DropwizardRuleHelper.dropwizardTestAppFrom(
+                      KafkaWithoutHealthCheckTestApplication.class)
+                  .withConfigFrom(KafkaTestConfiguration::new)
+                  .withRandomPorts()
+                  .withConfigurationModifier(
+                      c -> {
+                        KafkaConfiguration kafka = c.getKafka();
+
+                        kafka.setBrokers(
+                            KAFKA.getKafkaBrokers().stream()
```
christopher-cudennec, line 49: Well this programmatic configuration is pretty long. Probably copied from somewhere else? Don't think it's worth a refactoring.

sda-commons-server-kafka/src/main/java/org/sdase/commons/server/kafka/KafkaBundle.java
```diff
@@ -593,22 +596,37 @@ private void shutdownConsumerThreads() {
   }
 
   public interface FinalBuilder<T extends Configuration> {
+
+    /**
+     * Disables the health check for Kafka
+     *
+     * @return the same builder instance
+     */
+    FinalBuilder<T> withoutHealthCheck();
```
christopher-cudennec, line 605: Should we add some documentation in the module's README?

JoergSiebahn: Should we extend the Javadoc when this should be used and when not?

sda-commons-server-kafka/src/test/java/org/sdase/commons/server/kafka/health/KafkaWithoutHealthCheckIT.java
eikenilsknopp: Moved to constant `HEALTHCHECK_NAME` in KafkaBundle.

sda-commons-server-kafka/src/main/java/org/sdase/commons/server/kafka/KafkaBundle.java
eikenilsknopp: Added documentation to both README and Javadoc.

sda-commons-server-kafka/src/test/java/org/sdase/commons/server/kafka/health/KafkaWithoutHealthCheckIT.java
eikenilsknopp: Yes its copied from another test. I already stripped some configurations but I guess it could be reduced even further if thats necessary.

sda-commons-server-kafka/README.md
```diff
@@ -257,6 +257,17 @@ kafka:
       pollInterval: 200
 ```
 
+If the KafkaBundle is not essential for the functionality of your service you can disable the health check manually.
+This way your service can stay healthy even if the connection to Kafka is disrupted.
```
Fox32, line 261: It is possible to describe "not essential" a bit further? Can we give some usage example were a service should have a health check and where not?

eikenilsknopp: A specific example in our world is the document-access-service. Its cache is updated via kafka, but it could still fulfill its main task (handling requests from the kundenapp and delivering documents to it) without kafka, just with a potentially outdated cache.

Do you already have some general usage examples in mind? Im not sure if we can give usage examples in the documentation as its probably always very specific to the use case.

Fox32: Yeah such a example, a bit more generic. But also an example when the health check is useful, maybe a use-case were the data has to be up-2-date.

Maybe also an example for a producer: A producer might do error handling if the kafka broker is unavailable, e.g. by queuing stuff or telling the user that he can't process data right now. If the producer isn't doing that it might be useful if the health check fails.

sda-commons-server-kafka/src/test/java/org/sdase/commons/server/kafka/health/KafkaWithoutHealthCheckIT.java
Fox32: Maybe that helps as an example what is really required right now: https://github.com/SDA-SE/partner-consent-stack/blob/feat/events/consent-store-service/src/test/java/com/sdase/service/consent/events/EventManagerIT.java#L40-L49

Fox32: But I would keep the withBrokerProperty, that seems to be some kind of performance optimization for the tests

sda-commons-server-kafka/README.md
```diff
@@ -363,6 +363,23 @@ kafka:
...
+private KafkaBundle<KafkaTestConfiguration> bundle =
+  KafkaBundle.builder()
+    .withConfigurationProvider(KafkaTestConfiguration::getKafka)
+    .withoutHealthCheck() // disable kafkas health check
+    .build();
+```
+Keep in mind that in this case a producer might fail if the broker is not available, so depending on the
+use case the producer should do appropriate error handling e.g. storing unprocessed messages in a queue until the broker is available again.
+If no error handling is done you might be better off not disabling the health check.
+
```
christopher-cudennec, line 382: Good explanation 👍

christopher-cudennec (approved):
Ok for me.
What do you think @Fox32 ?

sda-commons-server-kafka/README.md
```diff
@@ -363,6 +363,23 @@ kafka:
       pollInterval: 200
 ```
 
+You can disable the  health check manually if kafka is not essential for the functionality of your service,
+e.g. it's only used to invalidate a cache, notify about updates or other minor tasks that could fail without
+affecting the service so much it's no longer providing a viable functionality.
```
Fox32, line 368: kafka or Kafka?

eikenilsknopp: Fixed the inconsistent spelling, should be Kafka everywhere.

sda-commons-server-kafka/src/test/java/org/sdase/commons/server/kafka/health/KafkaWithoutHealthCheckIT.java
```diff
@@ -0,0 +1,56 @@
...
+  private static final String CONSUMER = "consumer";
+  private static final String PRODUCER = "producer";
+
+  private static final String LISTENER_CONFIG_ASYNC = "async";
+
+  private static final DropwizardAppRule<KafkaTestConfiguration> DW =
+      new DropwizardAppRule<>(
+          KafkaWithoutHealthCheckTestApplication.class,
+          resourceFilePath("test-config-default.yml"),
+          config("kafka.disabled", "false"),
```
Fox32, line 39: you don't need this line, your kafka isn't disabled by the base config file

sda-commons-server-kafka/src/test/java/org/sdase/commons/server/kafka/dropwizard/KafkaWithoutHealthCheckTestApplication.java
```diff
@@ -0,0 +1,38 @@
...
+
+  private HealthCheckRegistry healthCheckRegistry;
+
+  @Override
+  public void initialize(Bootstrap<KafkaTestConfiguration> bootstrap) {
+    bootstrap.addBundle(ConfigurationSubstitutionBundle.builder().build());
```
Fox32, line 22: You shouldn't need this bundle, as you aren't using env substitution

sda-commons-server-kafka/src/test/java/org/sdase/commons/server/kafka/health/KafkaWithoutHealthCheckIT.java
eikenilsknopp: Right, removed

sda-commons-server-kafka/src/test/java/org/sdase/commons/server/kafka/dropwizard/KafkaWithoutHealthCheckTestApplication.java
eikenilsknopp: Hm, without it the deserialisation of the config.yml failes

Fox32: Yeah you are right. But removing the env variable in the config would break other things

Fox32 (approved):
👍

PR #605, opened 2020-10-27
feat(shared-asyncapi) sort definitions in generated json schema and asyncapi files

Fox32:
Sorting the schema makes it easier to find entities in a definition but also allows to compare them easier.
This might break assertions on generated files as the order of the content changes, but beside that there are no changes to
content.

Btw. changing the order in the `_expected` files has no influence on the tests, but I did it for fun 😉


christopher-cudennec (approved):

christopher-cudennec:
👍

dhenneke (approved):

PR #608, opened 2020-10-28
docs(server-kafka): document a production-ready configuration for the kafka bundle

dhenneke:
I added a configuration file example that can be copied and pasted to a new service.

What do you think about that? If this is a good idea, we should also add this to all the other bundles that require configurations.


christopher-cudennec (approved):
Nice 😃  👑

Fox32:
Looks good, quite similar what we did today. We need a follow up for the rest of the readme, most of it isn't right anymore.

PR #617, opened 2020-10-30
fix(server-mongo-testing): java.lang.IllegalArgumentException at LocalArtifactStore.java:48

RappC:


sda-commons-server-mongo-testing/src/main/java/org/sdase/commons/server/mongo/testing/MongoDbRule.java
```diff
@@ -156,6 +157,7 @@ private static IDownloadConfig createDownloadConfig() {
         downloadConfigBuilder.downloadPath(embeddedMongoDownloadPath);
       }
 
+      downloadConfigBuilder.artifactStorePath(new UUIDDir());
```
Fox32, line 160: Just as a warning and because I know you are not working on Mac: If the artifact store is the location where the executable is stored, that we should change that per test. Because that would mean that the Mac users have to agree to the firewall settings on every class. That's the reason for having only a single location for it.

We already have a fix for that, but that one wasn't merged because the dead letter handling branch was closed. Maybe that helps you: https://github.com/SDA-SE/sda-dropwizard-commons/commit/4c0757ce91a0d3062d08452af6db1361a84fec60

Fox32:
Btw, I would be fine with going with a lower coverage here. We can deactivate the check and reset in sonarcloud later.

cpsdase (dismissed):

sda-commons-server-mongo-testing/src/main/java/org/sdase/commons/server/mongo/testing/MongoDbRule.java
```diff
@@ -68,6 +70,12 @@
  */
 public class MongoDbRule extends ExternalResource {
   private static final Logger LOG = LoggerFactory.getLogger(MethodHandles.lookup().lookupClass());
+  private static final RetryRegistry downloadMongoRetryRegistry =
+      RetryRegistry.of(
+          RetryConfig.custom()
+              .maxAttempts(10)
+              .intervalFunction(IntervalFunction.ofRandomized())
+              .build());
```
christopher-cudennec, line 78: Oh, that's a nice approach! Much better than our self-implemented retry 👍

Fox32: Has the advantage that it doesn't count into the coverage 😆

christopher-cudennec:
Could you reproduce / try it out locally?

christopher-cudennec:
Agreed on closing the PR / branch.

PR #619, opened 2020-11-02
fix(server-jackson): use a custom hibernate node name provider to resolve the paths in the validation exceptions

### LINKED ISSUE #611: ExceptionHandler throws on invalid nested objects
Hey there,
we observed a strange behaviour in one of our projects with your default [JerseyValidationExceptionMapper](https://github.com/SDA-SE/sda-dropwizard-commons/blob/1f12329e447d49b76592e3fd32cbd4878e48ab31/sda-commons-server-jackson/src/main/java/org/sdase/commons/server/jackson/errors/JerseyValidationExceptionMapper.java#L114) throwing an exception when using invalid nested model classes. 
I created a minimal example at [https://github.com/rschwedasi/sda-exception-fail](https://github.com/rschwedasi/sda-exception-fail) serving two endpoints showing the expected behaviour and the erroneous one. 

I hope the example and its [README](https://github.com/rschwedasi/sda-exception-fail/blob/main/README.md) are self-explaining. Please don't hesitate to ask further questions. 

Best regards,
Robert S.

dhenneke:
Previously, we used custom code to correctly resolve the property names (if used with `@JsonProperty("name")`). Hibernate now supports providing a custom property node name provider (see https://github.com/hibernate/hibernate-validator/pull/1029). They also provide an example implementation for Jackson in their documentation. After applying this custom `propertyNodeNameProvider`, we can use an existing function from Dropwizard (`ConstraintMessage.isRequestEntity(...)`) that computes the correct property path.

Closes #611

### DIFF
--- a/sda-commons-server-jackson/src/test/java/org/sdase/commons/server/jackson/test/NameNotRepeatedValidator.java
+++ b/sda-commons-server-jackson/src/test/java/org/sdase/commons/server/jackson/test/NameNotRepeatedValidator.java
@@ -19,13 +19,12 @@ public class NameNotRepeatedValidator
       context.disableDefaultConstraintViolation();
       context
           .buildConstraintViolationWithTemplate("First name and last name must be different.")
-          .addPropertyNode("firstName")
-          .addBeanNode()
+          // The property firstName is annotated with @JsonProperty("name")
+          .addPropertyNode("name")
           .addConstraintViolation();
       context
           .buildConstraintViolationWithTemplate("First name and last name must be different.")
           .addPropertyNode("lastName")
-          .addBeanNode()
           .addConstraintViolation();
       return false;
     }
--- a/sda-commons-server-jackson/src/test/java/org/sdase/commons/server/jackson/test/NestedNestedResource.java
+++ b/sda-commons-server-jackson/src/test/java/org/sdase/commons/server/jackson/test/NestedNestedResource.java
@@ -1,0 +1,14 @@
+package org.sdase.commons.server.jackson.test;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+import javax.validation.constraints.NotEmpty;
+
+public class NestedNestedResource {
+
+  @NotEmpty()
+  @JsonProperty("anotherNestedField")
+  private String anotherNested;
+
+  @JsonProperty("someNumber")
+  private int anotherNumber;
+}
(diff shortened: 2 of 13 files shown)


sda-commons-server-jackson/src/test/java/org/sdase/commons/server/jackson/test/NameNotRepeatedValidator.java
```diff
@@ -19,13 +19,11 @@ public boolean isValid(ValidationResource value, ConstraintValidatorContext cont
       context.disableDefaultConstraintViolation();
       context
           .buildConstraintViolationWithTemplate("First name and last name must be different.")
-          .addPropertyNode("firstName")
-          .addBeanNode()
+          .addPropertyNode("name")
```
dhenneke, line 22: This custom validator must set the correct (=resolved) name of the property since this is not forwarded to the `PropertyNodeNameProvider`. Since this code is only used for testing, I guess we can just change it.

In addition, the call `addBeanNode()` seems to be wrong here, since it added an empty path entry to the list that led to the generation of the path `name.` instead of `name`.

JoergSiebahn: Can we add a comment to avoid confusion?
```suggestion
          // The property firstName is annotated with @JsonProperty("name")
          .addPropertyNode("name")
```

sda-commons-server-jackson/src/test/java/org/sdase/commons/server/jackson/test/NestedNestedResource.java
```diff
@@ -0,0 +1,14 @@
+package org.sdase.commons.server.jackson.test;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+import org.hibernate.validator.constraints.NotEmpty;
```
JoergSiebahn, line 4: Is it intended to use the deprecated Hibernate version? Should we move to `javax.validation.constraints.NotEmpty`?

dhenneke: Correct. I added a commit that changes all the deprecated usages from `org.hibernate.validator.constraints.NotEmpty` to `javax.validation.constraints.NotEmpty`. That reduces the risk to inherit deprecated code via copy-and-pasting in the future.

JoergSiebahn (approved):

PR #624, opened 2020-11-03
fix(server-kafka): only configure the sasl settings if a sasl protocol is configured

dhenneke:
The service emits a warning in the log if the `sasl.jaas.config` property is set even if the security protocol doesn't support SASL:

> The configuration 'sasl.jaas.config' was supplied but isn't a known config.

This behavior was introduced in #548 that allowed the configuration of `SSL` without configured (and unneeded) user credentials.


christopher-cudennec (commented):
Nice improvement. I'll approve after rebase :).

christopher-cudennec (approved):
👍

Fox32 (approved):
👍

PR #628, opened 2020-11-04
Feat/test updates

dhenneke:
I did change three things:
1. Deprecate the `LazyRule` and remove all usages in this repository.
2. Deprecate the `DropwizardRuleHelper` and remove all usages in this repository.
3. Try to improve the test stability.

### `LazyRule`

We often have the following pattern in our services. This is necessary because `WIRE.baseUrl()` can only be accessed after WireMock has been started.

```java
class MyTest {
  private static final WireMockClassRule WIRE =
      new WireMockClassRule(wireMockConfig().dynamicPort());

  private static final LazyRule<DropwizardAppRule<AppConfiguration>> DW =
      new LazyRule<>(
          () ->
              new DropwizardAppRule<>(
                  TestApplication.class,
                  ResourceHelpers.resourceFilePath("test-config.yml"),
                  ConfigOverride.config("url", WIRE.baseUrl())));

  @ClassRule public static final RuleChain CHAIN = RuleChain.outerRule(WIRE).around(DW);
}
```

But the `DropwizardAppRule` already provides a solution for this: there is a `ConfigOverride.config(String key, Supplier<String> value)` variant, so it can be written without the `LazyRule`:

```diff
  class MyTest {
    private static final WireMockClassRule WIRE =
        new WireMockClassRule(wireMockConfig().dynamicPort());

    private static final DropwizardAppRule<AppConfiguration> DW =
-       new LazyRule<>(
-           () ->
        new DropwizardAppRule<>(
            TestApplication.class,
            ResourceHelpers.resourceFilePath("test-config.yml"),
-           ConfigOverride.config("url", WIRE.baseUrl())));
+           ConfigOverride.config("url", WIRE::baseUrl));

  @ClassRule public static final RuleChain CHAIN = RuleChain.outerRule(WIRE).around(DW);
}
```

This also allows more complex values like `config("url", () -> WIRE.path("anything"))`. We also have a test that requires such a feature from the `SystemPropertyRule`, so it has also been added there.

### `DropwizardRuleHelper`

This one is a bit trickier. This helper should make it easier to configure tests, because you don't need a config-file and can more easily setup a `withRandomPorts` behavior. But I would argue that we should try to stick to Dropwizard as near as possible since I hope that this would also make it easier to switch to JUnit 5 in the future. There were also cases where the creation of configuration objects in java was more verbose than the other approach (e.g. `KafkaBundleWithConfigIT`). Also, this is quite far from how one would configure the service for production usage.

I prefer the config-file + config-overrides way over the rule helper. But this might be up to taste.

With `DropwizardRuleHelper`:
```java
  private static final LazyRule<DropwizardAppRule<Config>> DW =
      new LazyRule<>(
          () ->
              DropwizardRuleHelper.dropwizardTestAppFrom(MorphiaTestApp.class)
                  .withConfigFrom(Config::new)
                  .withRandomPorts()
                  .withConfigurationModifier(
                      c ->
                          c.getMongo()
                              .setHosts(MONGODB.getHost())
                              .setDatabase(MONGODB.getDatabase()))
                  .build());
```

Without:
```yaml
server:
  applicationConnectors:
  - type: http
    port: 0
  adminConnectors:
  - type: http
    port: 0

```

```java
  private static final DropwizardAppRule<Config> DW =
      new DropwizardAppRule<>(
          MorphiaTestApp.class,
          resourceFilePath("test-config.yaml"),
          config("mongo.hosts", MONGODB::getHost),
          config("mongo.database", MONGODB::getDatabase));
```

### Tests

We have some flaky tests and I tried to make them more stable. There were some issues:

* Some tests used `@Rule` instead of `@ClassRule` without a good reason, which led to unnecessary application restarts. If used together with `@Retry`, tests always called cold-services and timeout issues can't be resolved on a highly-utilized build runner.
* There are some strange assertions regarding async operations (see `ApiClientTimeoutTest`).
* Reduce calls to things like `KAFKA.getKafkaTestUtils().consumeAllRecordsFromTopic(TOPIC_NAME)` since they can lead to timeout-errors and can delay the test-execution needlessly.

This doesn't mean that the tests are super stable from now on, but I hope this improves it a bit. There are still some issues around e.g. `createTopics`.

### DIFF
--- a/sda-commons-client-jersey-example/src/test/java/org/sdase/commons/client/jersey/JerseyClientExampleIT.java
+++ b/sda-commons-client-jersey-example/src/test/java/org/sdase/commons/client/jersey/JerseyClientExampleIT.java
@@ -6,6 +6,7 @@ import static com.github.tomakehurst.wiremock.client.WireMock.get;
 import static com.github.tomakehurst.wiremock.client.WireMock.urlEqualTo;
 import static com.github.tomakehurst.wiremock.client.WireMock.urlMatching;
 import static com.github.tomakehurst.wiremock.core.WireMockConfiguration.wireMockConfig;
+import static io.dropwizard.testing.ConfigOverride.config;
 import static io.dropwizard.testing.ResourceHelpers.resourceFilePath;
 import static java.util.Arrays.asList;
 import static org.assertj.core.api.Assertions.assertThat;
@@ -20,28 +21,22 @@ import org.apache.http.HttpHeaders;
 import org.junit.Before;
 import org.junit.BeforeClass;
 import org.junit.ClassRule;
-import org.junit.Rule;
 import org.junit.Test;
 import org.junit.rules.RuleChain;
 import org.sdase.commons.client.jersey.clients.apia.Car;
-import org.sdase.commons.server.testing.EnvironmentRule;
 
 public class JerseyClientExampleIT {
 
-  @ClassRule
   public static final WireMockClassRule WIRE =
       new WireMockClassRule(wireMockConfig().dynamicPort());
 
   private static final DropwizardAppRule<JerseyClientExampleConfiguration> DW =
       new DropwizardAppRule<>(
-          JerseyClientExampleApplication.class, resourceFilePath("test-config.yaml"));
+          JerseyClientExampleApplication.class,
+          resourceFilePath("test-config.yaml"),
+          config("servicea", () -> WIRE.url("api")));
 
-  @Rule
-  public final RuleChain rule =
-      RuleChain.outerRule(
-              new EnvironmentRule()
-                  .setEnv("SERVICE_A_URL", String.format("%s/%s", WIRE.baseUrl(), "api")))
-          .around(DW);
+  @ClassRule public static final RuleChain rule = RuleChain.outerRule(WIRE).around(DW);
 
   private JerseyClientExampleApplication app;
   private static final ObjectMapper OM = new ObjectMapper();
@@ -49,11 +44,6 @@ public class JerseyClientExampleIT {
       new Car().setSign("HH XX 1234").setColor("bright blue"); // NOSONAR
   private static final Car LIGHT_BLUE_CAR =
       new Car().setSign("HH XY 4321").setColor("light blue"); // NOSONAR
-
-  @BeforeClass
-  public static void start() {
-    WIRE.start();
-  }
 
   @Before
   public void resetRequests() {
--- a/sda-commons-client-jersey/src/test/java/org/sdase/commons/client/jersey/ClientProxyTest.java
+++ b/sda-commons-client-jersey/src/test/java/org/sdase/commons/client/jersey/ClientProxyTest.java
@@ -23,7 +23,6 @@ import org.junit.Test;
 import org.junit.rules.RuleChain;
 import org.sdase.commons.client.jersey.test.ClientTestApp;
 import org.sdase.commons.client.jersey.test.ClientTestConfig;
-import org.sdase.commons.server.testing.LazyRule;
 import org.sdase.commons.server.testing.SystemPropertyRule;
 
 /** A test that checks if the proxy can be configured via system properties */
@@ -34,13 +33,11 @@ public class ClientProxyTest {
   private static final WireMockClassRule PROXY_WIRE =
       new WireMockClassRule(wireMockConfig().dynamicPort());
 
-  private static final LazyRule<SystemPropertyRule> PROP =
-      new LazyRule<>(
-          () ->
-              new SystemPropertyRule()
-                  .setProperty("http.proxyHost", "localhost")
-                  .setProperty("http.proxyPort", "" + PROXY_WIRE.port())
-                  .setProperty("http.nonProxyHosts", "localhost"));
+  private static final SystemPropertyRule PROP =
+      new SystemPropertyRule()
+          .setProperty("http.proxyHost", "localhost")
+          .setProperty("http.proxyPort", () -> "" + PROXY_WIRE.port())
+          .setProperty("http.nonProxyHosts", "localhost");
 
   private static final DropwizardAppRule<ClientTestConfig> DW =
       new DropwizardAppRule<>(ClientTestApp.class, resourceFilePath("test-config.yaml"));
--- a/sda-commons-server-auth-testing/src/test/java/org/sdase/commons/server/auth/testing/KeyLoaderProxyIT.java
+++ b/sda-commons-server-auth-testing/src/test/java/org/sdase/commons/server/auth/testing/KeyLoaderProxyIT.java
@@ -21,41 +21,37 @@ import javax.ws.rs.core.Response;
 import org.apache.http.HttpHeaders;
 import org.junit.BeforeClass;
 import org.junit.ClassRule;
-import org.junit.Rule;
 import org.junit.Test;
 import org.junit.rules.RuleChain;
 import org.sdase.commons.server.auth.testing.test.AuthTestApp;
 import org.sdase.commons.server.auth.testing.test.AuthTestConfig;
 import org.sdase.commons.server.testing.EnvironmentRule;
-import org.sdase.commons.server.testing.LazyRule;
 import org.sdase.commons.server.testing.SystemPropertyRule;
 
 /** A test that checks if the jersey client that is used to load keys can use a proxy server */
 public class KeyLoaderProxyIT {
-  @ClassRule
-  public static final WireMockClassRule PROXY_WIRE =
+  private static final WireMockClassRule PROXY_WIRE =
       new WireMockClassRule(wireMockConfig().dynamicPort());
 
-  @ClassRule
-  public static final EnvironmentRule ENVIRONMENT_RULE =
+  private static final EnvironmentRule ENVIRONMENT_RULE =
       new EnvironmentRule().setEnv("AUTH_RULE", "{\"keys\": [{}]}");
 
-  private final SystemPropertyRule systemProperty =
+  private static final SystemPropertyRule SYSTEM_PROPERTY =
       new SystemPropertyRule()
           .setProperty("http.proxyHost", "localhost")
-          .setProperty("http.proxyPort", "" + PROXY_WIRE.port())
+          .setProperty("http.proxyPort", () -> "" + PROXY_WIRE.port())
           .setProperty("http.nonProxyHosts", "localhost");
 
-  private final LazyRule<DropwizardAppRule<AuthTestConfig>> dw =
-      new LazyRule<>(
-          () ->
-              new DropwizardAppRule<>(
-                  AuthTestApp.class,
-                  ResourceHelpers.resourceFilePath("test-config.yaml"),
-                  config("auth.keys[0].type", "JWKS"),
-                  config("auth.keys[0].location", "http://sda.se/jwks")));
+  private static final DropwizardAppRule<AuthTestConfig> DW =
+      new DropwizardAppRule<>(
+          AuthTestApp.class,
+          ResourceHelpers.resourceFilePath("test-config.yaml"),
+          config("auth.keys[0].type", "JWKS"),
+          config("auth.keys[0].location", "http://sda.se/jwks"));
 
-  @Rule public final RuleChain rule = RuleChain.outerRule(systemProperty).around(dw);
+  @ClassRule
+  public static final RuleChain rule =
+      RuleChain.outerRule(PROXY_WIRE).around(ENVIRONMENT_RULE).around(SYSTEM_PROPERTY).around(DW);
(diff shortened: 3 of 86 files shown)


sda-commons-client-jersey-example/src/test/java/org/sdase/commons/client/jersey/JerseyClientExampleIT.java
```diff
@@ -50,11 +45,6 @@
   private static final Car LIGHT_BLUE_CAR =
       new Car().setSign("HH XY 4321").setColor("light blue"); // NOSONAR
 
-  @BeforeClass
-  public static void start() {
-    WIRE.start();
-  }
```
Fox32, line 56: This is also a pattern that I see a lot. As a explanation why we should delete it: If we use Wiremock as a rule (with `@Rule` or via the `RuleChain`), it is automatically started and stopped. There is no need to do it manually.

sda-commons-client-jersey/src/test/java/org/sdase/commons/client/jersey/ClientProxyTest.java
```diff
@@ -34,13 +33,11 @@
...
-      new LazyRule<>(
-          () ->
-              new SystemPropertyRule()
-                  .setProperty("http.proxyHost", "localhost")
-                  .setProperty("http.proxyPort", "" + PROXY_WIRE.port())
-                  .setProperty("http.nonProxyHosts", "localhost"));
+  private static final SystemPropertyRule PROP =
+      new SystemPropertyRule()
+          .setProperty("http.proxyHost", "localhost")
+          .setProperty("http.proxyPort", () -> "" + PROXY_WIRE.port())
```
Fox32, line 39: Can we extend this to pass `Object` instead of `String`. Later call `.toString()`.

sda-commons-server-auth-testing/src/test/java/org/sdase/commons/server/auth/testing/KeyLoaderProxyIT.java
```diff
@@ -21,41 +21,37 @@
...
 public class KeyLoaderProxyIT {
-  @ClassRule
-  public static final WireMockClassRule PROXY_WIRE =
+  private static final WireMockClassRule PROXY_WIRE =
       new WireMockClassRule(wireMockConfig().dynamicPort());
 
-  @ClassRule
-  public static final EnvironmentRule ENVIRONMENT_RULE =
+  private static final EnvironmentRule ENVIRONMENT_RULE =
       new EnvironmentRule().setEnv("AUTH_RULE", "{\"keys\": [{}]}");
```
Fox32, line 37: Can't we set this via config()?

dhenneke: This is a tricky one. With today's knowledge, I guess we would have build the AuthRule in a different way. This `auth: ${AUTH_RULE}` entry in the config.yml is not very nice and can make the life hard. Better would have been to just set config overrides (`ConfigOverride.config("auth.keys", "whatever").addToSystemProperties()`) if possible (didn't tested it).

But I can try if it is really necessary here.

sda-commons-client-jersey/src/test/java/org/sdase/commons/client/jersey/ClientProxyTest.java
dhenneke: This is the Dropwizard style and we have the same problems there: `public static ConfigOverride config(String key, Supplier<String> value) {`. But we could change it to Object.

sda-commons-server-kafka/src/test/java/org/sdase/commons/server/kafka/KafkaBundleWithConfigIT.java
```diff
@@ -860,16 +780,6 @@ public void shouldSetConsumerNameCorrectlyWithConsumerConfig() {
...
-  @Test
-  public void shouldSetConsumerNameCorrectlyWithConsumerConfigFromYaml() {
-    try (KafkaConsumer<String, String> c1 =
-        kafkaBundle.createConsumer(
-            new StringDeserializer(), new StringDeserializer(), CONSUMER_1)) {
-
-      assertThat(KafkaHelper.getClientId(c1)).isEqualTo(CONSUMER_1 + "-0");
-    }
-  }
-
```
Fox32, line 872: Why is this gone?

sda-commons-client-jersey/src/test/java/org/sdase/commons/client/jersey/ClientProxyTest.java
Fox32: Ah ok, thought that Dropwizard is using an object here. No keep it like Dropwizard.

sda-commons-server-auth-testing/src/test/java/org/sdase/commons/server/auth/testing/KeyLoaderProxyIT.java
Fox32: Ah, right if we have an env variable there, we can't later reset it without having a custom config.yaml

sda-commons-server-kafka/src/test/java/org/sdase/commons/server/kafka/KafkaBundleWithConfigIT.java
dhenneke: https://github.com/SDA-SE/sda-dropwizard-commons/blob/f35d324874fd14657ac0e7c897d54c0ce83016af/sda-commons-server-kafka/src/test/java/org/sdase/commons/server/kafka/KafkaBundleWithConfigIT.java#L275 and https://github.com/SDA-SE/sda-dropwizard-commons/blob/f35d324874fd14657ac0e7c897d54c0ce83016af/sda-commons-server-kafka/src/test/java/org/sdase/commons/server/kafka/KafkaBundleWithConfigIT.java#L864 are identical. One test less to execute is one test less which can fail.

sda-commons-server-kafka/src/test/java/org/sdase/commons/server/kafka/KafkaBundleWithConfigIT.java
```diff
@@ -531,9 +455,6 @@ public void kafkaConsumerReceivesMessages() {
                     new IgnoreAndProceedErrorHandler<>()))
             .build());
 
-    // empty topic before test
-    KAFKA.getKafkaTestUtils().consumeAllRecordsFromTopic(topic);
```
dhenneke, line 535: This is a new topic, so we can skip this. Saves us >3 seconds in execution and a potential failure-opportunity.

JoergSiebahn:
While we are on to move closer to Dropwizard and closer to production configuration: Is there a way to configure the random ports without a dedicated `test-config.yaml` and use the real defaults of the production application from `config.yaml`?
One reason for the rule helper was to get rid of a second config file and use defaults from the configuration where possible. It also has it's downsides because it did not consider the `config.yaml` but defaults in the code. It avoided confusion with two yaml files that had to be kept as similar as possible manually while introducing other confusion.

Fox32:
It would be interesting if there is a way to build the random ports directly into `DropwizardAppRule`, but I don't think that we would succeed with it, as it's deprecated in DW2.

Fox32:
> While we are on to move closer to Dropwizard and closer to production configuration: Is there a way to configure the random ports without a dedicated `test-config.yaml` and use the real defaults of the production application from `config.yaml`?
> One reason for the rule helper was to get rid of a second config file and use defaults from the configuration where possible. It also has it's downsides because it did not consider the `config.yaml` but defaults in the code. It avoided confusion with two yaml files that had to be kept as similar as possible manually while introducing other confusion.

Most of the time the main `config.yaml` is not useful for tests as it requires env substitution for parameters that one doesn't want to bother with in tests. Not sure if thats a good way.

JoergSiebahn:
Hopefully the env replacements have defaults in most cases. If not, the developers should feel the same pain in their tests, the operators feel with the deployment. Better they find wisely chosen defaults. I think the env properties can be overwritten with `ConfigOverride` as well, so it should not require to create additional environment variables in the tests.
It may be an option to promote the setting of random ports with `ConfigOverride` in the sda-dropwizard-commons tests. But we can discuss and evaluate that in a follow up.

JoergSiebahn:
> But sadly, Java can't add arrays to a varargs function :-( Looks like Kotlin supports it, but that might not be what we want to head to.

An option  could be something like

```java
new DropwizardAppRule<>(
    TestApplication.class,
    ResourceHelpers.resourceFilePath("config.yml"), // or null,
    withRandomPortsAndConfigOverrides(
        ConfigOverride.config("url", WIRE::baseUrl),
        // other overrides
    ));
```

dhenneke:
Or another option:

<details>
<summary>RandomPortsRule</summary>

```java
  public static class RandomPortsRule implements TestRule {
    ConfigOverride config = config("server.applicationConnectors[0].port", "0");
    ConfigOverride config1 = config("server.adminConnectors[0].port", "0");

    @Override
    public Statement apply(Statement base, Description description) {
      return new Statement() {
        @Override
        public void evaluate() throws Throwable {
          config.addToSystemProperties();
          config1.addToSystemProperties();

          try {
            base.evaluate();
          } finally {
            config.removeFromSystemProperties();
            config1.removeFromSystemProperties();
          }
        }
      };
    }
  }
```
</details>

```java
  private static final DropwizardAppRule<Config> DW =
      new DropwizardAppRule<>(
          MorphiaTestApp.class,
          null,
          config("mongo.hosts", MONGODB::getHost),
          config("mongo.database", MONGODB::getDatabase));

  @ClassRule
  public static final RuleChain CHAIN =
      RuleChain.outerRule(MONGODB).around(new RandomPortsRule()).around(DW);

```

dhenneke:
Or a DW change so that we could do something like:

```java
  private static final DropwizardAppRule<Config> DW =
      new DropwizardAppRule<>(
          MorphiaTestApp.class,
          null,
          config("mongo.hosts", MONGODB::getHost),
          config("mongo.database", MONGODB::getDatabase))
      .withRandomPorts();
```

dhenneke:
Or even:

```java
  public static class RandomPortsConfigOverride {
    static ConfigOverride config = config("server.applicationConnectors[0].port", "0");
    static ConfigOverride config1 = config("server.adminConnectors[0].port", "0");

    public static ConfigOverride randomPortsConfig() {
      ConfigOverride mock = Mockito.mock(ConfigOverride.class);

      Mockito.doAnswer(
              invocation -> {
                config.addToSystemProperties();
                config1.addToSystemProperties();
                return null;
              })
          .when(mock)
          .addToSystemProperties();

      Mockito.doAnswer(
              invocation -> {
                config.removeFromSystemProperties();
                config1.removeFromSystemProperties();
                return null;
              })
          .when(mock)
          .removeFromSystemProperties();

      return mock;
    }
  }

  private static final DropwizardAppRule<Config> DW =
      new DropwizardAppRule<>(
          MorphiaTestApp.class,
          null,
          RandomPortsConfigOverride.randomPortsConfig(),
          // other
      );
```

(Extending `ConfigOverride` is not possible since it has a private constructor).

dhenneke:
I opened a PR at Dropwizard to implement the last option in a cleaner way, i.e. having a `ConfigOverride.randomPort()`.

sda-commons-client-jersey-example/src/test/java/org/sdase/commons/client/jersey/JerseyClientExampleIT.java
christopher-cudennec: 👍

christopher-cudennec:
> While we are on to move closer to Dropwizard and closer to production configuration: Is there a way to configure the random ports without a dedicated test-config.yaml and use the real defaults of the production application from config.yaml?
One reason for the rule helper was to get rid of a second config file and use defaults from the configuration where possible. It also has it's downsides because it did not consider the config.yaml but defaults in the code. It avoided confusion with two yaml files that had to be kept as similar as possible manually while introducing other confusion.

Has anyone ever tried to use the proper `config.yml` and add an `EnvironmentRule`? That would be exactly what we do in production.

JoergSiebahn:
> Has anyone ever tried to use the proper `config.yml` and add an `EnvironmentRule`? That would be exactly what we do in production.

As far as I know we do that only in a `ConfigIT` to test the validity of the production `config.yml` when it contains mandatory environment variables.

dhenneke:
That would work, but I think that our AuthRule isn't compatible with such an approach, yet.

dhenneke:
Put aside the random ports and test-config.yaml discussions, any more comments to the changes of the PR?

Fox32 (approved):
👍

PR #637, opened 2020-11-10
Add dedicated config class for Api clients

JoergSiebahn:
In many service we had to extend the HttpClientConfiguration to add a base url of the API which is required for API clients based on interfaces.
This commit provides a built in approach to simplify configuration and creation of API clients.
The downside is that there are some classes partially duplicated due to the builder pattern.
But that gives the opportunity to get rid of the deprecated timeout settings of the previous fluent configuration.

### DIFF
--- a/sda-commons-client-jersey/README.md
+++ b/sda-commons-client-jersey/README.md
@@ -63,10 +63,11 @@ Clients for the SDA Platform have some magic added that clients for an external 
   SDA Platform clients can be configured to pass through the Authorization header of an incoming request context:
   
   ```java
-  clientFactory.platformClient()
+  // using ApiHttpClientConfiguration property from service config
+  clientFactory
+        .platformClient(configuration.getOtherSdaServiceClient()) 
         .enableAuthenticationPassThrough()
-        .api(OtherSdaServiceClient.class)
-        .atTarget("http://other-sda-service.sda.net/api");
+        .api(OtherSdaServiceClient.class);
   ```
 - _Consumer-Token_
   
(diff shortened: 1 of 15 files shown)


dhenneke:
I like the idea, but I think that the changes makes the use quite complex. Now you already need to think about the type of configuration class to use even before you decided whether you want to get a Client or an API-Client. I have the impression that it is sometimes not obvious that the API-clients exist at all.

Just as an experiment, what about the following:

1. Let the `ApiHttpClientConfiguration` extend `HttpClientConfiguration` (not possible now due to overloading)
1. Deprecate the `platformClient(<configuration>)` and `externalClient(<configuration)` option.
2. Provide a configuration and a name _after_ entering the api/generic mode.
3. Use a new `genericClient()` option beside the `api(Any.class)` option.

```java
// Create a platform API Client
app.getJerseyClientBundle()
    .getClientFactory()
    .platformClient() // same for externalClient()
    .enableAuthenticationPassThrough()
    .api(MockApiClient.class)
    // a new option beside the (deprecated) atTarget(String url)
    .withConfiguration(new ApiHttpClientConfiguration())
    .build();

// Create a platform API Client with a custom name
app.getJerseyClientBundle()
    .getClientFactory()
    .platformClient()
    .enableAuthenticationPassThrough()
    // don't provide the name here...
    .api(MockApiClient.class)
    .withConfiguration(new ApiHttpClientConfiguration())
    // ... provide it here.
    .build("Custom Name");

// Create a external generic client with a name
app.getJerseyClientBundle()
    .getClientFactory()
    .externalClient()
    // don't call buildGenericClient(String name) directly but enter the "generic mode"
    .genericClient()
    // same options as before. but here withConfiguration is optional
    .withConfiguration(new HttpClientConfiguration())
    // name is required in this branch
    .build("asdf");
```

Would require some Generic-Interface-Juggling, but it could work. A downside would be, that we won't get rid of the deprecated timeout settings with this approach. Plus there would be some edge-cases since the old `externalClient(<configuration>)` would lead to the same builders and one could set a configuration twice.

JoergSiebahn:
I tried another approach in #647 with a new entrypoint to the fluent builder.

PR #638, opened 2020-11-10
fix(deps): align json-unit-core in all dependencies

dhenneke:
We get dependency conflicts when an application uses both wiremock and json-unit-assertj in the same classpath. We currently don't get a conflict in sda-dropwizard-commons since json-unit-assertj is only included as a test dependency.

I'm not sure if this is a smart decision. Feel free to comment 😄

### DIFF
--- a/sda-commons-dependencies/build.gradle
+++ b/sda-commons-dependencies/build.gradle
@@ -14,7 +14,7 @@ ext {
   swaggerCoreV3Version = '2.1.5'
   weldVersion = '3.1.5.Final'
   openTracingUtilVersion = '0.33.0'
-  jsonUnitVersion = '2.19.0'
+  jsonUnitVersion = '2.21.0'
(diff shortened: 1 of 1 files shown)


Fox32 (dismissed):
👍 I would be fine with it.

sda-commons-dependencies/build.gradle
```diff
@@ -14,6 +14,7 @@ ext {
   swaggerCoreV3Version = '2.1.5'
   weldVersion = '3.1.5.Final'
   openTracingUtilVersion = '0.33.0'
+  jsonUnitVersion = '2.19.0'
```
JoergSiebahn, line 17: Is there a reason why we don't use [2.21.0](https://search.maven.org/artifact/net.javacrumbs.json-unit/json-unit/2.21.0/bundle). If there is one, can we add a comment to avoid upgrade by mistake?

dhenneke: Correct. We would have got a seprarate version from https://github.com/SDA-SE/sda-dropwizard-commons/pull/636. I'll update it in this PR.

JoergSiebahn (approved):

PR #640, opened 2020-11-11
feat(server-starter-swagger): rename '*-starter' to '*-starter-swagger'

huebner-j:
The old 'sda-commons-server-starter' module can still be used by importing the new module 'sda-commons-server-starter-swagger'

This change prepares a new revised version of 'sda-commons-server-starter' including OpenApi 3 support.


Fox32:
This is a breaking change and should be marked in the commit message as such. Please also add the steps for the mitigation (using the new module) to that message.

huebner-j:
> This is a breaking change and should be marked in the commit message as such. 
Thank you @Fox32 , I missed that one. I added 'BREAKING CHANGE:' to the commit message.

> Please also add the steps for the mitigation (using the new module) to that message.
I've already added the steps the users of sda-commos need to take, but I've tried to point them out more clearly. So please check again, if the description is ok for you now...

Fox32:
If this creates a new major release, we have to update some documentation that mentions to 2.0.0 release. For example the readme and here https://github.com/SDA-SE/sda-dropwizard-commons/security/policy

We should mention that we only to security updates for 3.0.0

huebner-j:
> If this creates a new major release, we have to update some documentation that mentions to 2.0.0 release. For example the readme and here https://github.com/SDA-SE/sda-dropwizard-commons/security/policy

Good point. I've found two files where versions are mentioned:
* [SECURITY.md](https://github.com/SDA-SE/sda-dropwizard-commons/blob/feat/openapi3-starter-bundle/SECURITY.md)
* [README.md](https://github.com/SDA-SE/sda-dropwizard-commons/blob/feat/openapi3-starter-bundle/README.md)

I've changed these sections. Please let me know if something is still missing.

> We should mention that we only to security updates for 3.0.0

The current security policy says, that version `1.x.x` and version `2.x.x` receives critical security updates.  Wouldn't that be a break, if we only do security updates for version 3.0.0?

dhenneke:
> The current security policy says, that version `1.x.x` and version `2.x.x` receives critical security updates. Wouldn't that be a break, if we only do security updates for version 3.0.0?

It was more meant in the way "`1.x.x` and `latest release in master`".

huebner-j:
> > The current security policy says, that version `1.x.x` and version `2.x.x` receives critical security updates. Wouldn't that be a break, if we only do security updates for version 3.0.0?
> 
> It was more meant in the way "`1.x.x` and `latest release in master`".

Thank you for the clarification. Please check the documentation again :)

huebner-j:
@SDA-SE/sda-commons-maintainer have decided to keep the `sda-dropwizard-server-starter` module as is and to introduce a new starter module called `sda-dropwizard-starter`. The new module will be a revised version of `sda-dropwizard-server-starter` containing OpenApi 3 support. The advantage of this approach is that no breaking change is needed.

PR #648, opened 2020-11-13
Provide annotations to suppress exceptions when using API clients

JoergSiebahn:
We realized, that client errors are handled in only two ways very often:

- some errors are mitigated in the same way, weather it is a 404 or a 403
- all other errors are rethrown to produce a 500 for the caller

Therefore we wrote a lot of code to catch, verify and handle exceptions.
This code can be avoided with the new suppression annotations.
Consumers may still handle exceptions for technical reasons, e.g. do a retry on connection errors, but use the suppression feature for business implementation if for example 404 is an expected result.

### DIFF
--- a/sda-commons-client-jersey/README.md
+++ b/sda-commons-client-jersey/README.md
@@ -228,3 +228,47 @@ org.glassfish.jersey.client.JerseyClientBuilder
 ```
 
 This works if the library that requires the other implementation does not rely on the Java ServiceLoader.
+
+
+### Consume API clients without catching exceptions
+
+The [`Suppress*`](./src/main/java/org/sdase/commons/client/jersey/proxy/annotation) annotations can be used to suppress
+specific `ClientRequestException` thrown in any defined API method where the return type is not a `Response`.
+There are configurable suppressors for 
+[HTTP errors](./src/main/java/org/sdase/commons/client/jersey/proxy/annotation/SuppressHttpErrorsToNull.java),
+[connection timeouts](./src/main/java/org/sdase/commons/client/jersey/proxy/annotation/SuppressConnectTimeoutErrorsToNull.java),
+[read timeouts](./src/main/java/org/sdase/commons/client/jersey/proxy/annotation/SuppressReadTimeoutErrorsToNull.java)
+and all other [processing errors](./src/main/java/org/sdase/commons/client/jersey/proxy/annotation/SuppressProcessingErrorsToNull.java).
+
+Suppressed exceptions are logged at info level.
+
+The following client will convert a _404 Not Found_ error to `null` so that the consumer can handle 
+an entity that is not found by checking if the return value is `null`.
+Any other error will lead to a _500 Internal Server Error_ because the `ClientRequestException` is converted in the
+[`ClientRequestExceptionMapper`](./src/main/java/org/sdase/commons/client/jersey/error/ClientRequestExceptionMapper.java).
+
+```java
+@Path("/examples")
+public interface ExampleClient {
+
+  @GET
+  @Path("/{exampleId}")
+  @Produces(MediaType.APPLICATION_JSON)
+  @SuppressHttpErrorsToNull(404)
+  Example getExample(@PathParam("exampleId") String exampleId);
+
+}
+
+public class ExampleService {
+   
+  private ExampleClient exampleClient;
+  private DoSomethingService doSomethingService;
+
+  public void doSomethingIfExampleExists(String exampleId) {
+    Example example = exampleClient.getExample(exampleId);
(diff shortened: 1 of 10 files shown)


sda-commons-client-jersey/README.md
```diff
@@ -228,3 +228,47 @@ org.glassfish.jersey.client.JerseyClientBuilder
...
+[`ClientRequestExceptionMapper`](./src/main/java/org/sdase/commons/client/jersey/error/ClientRequestExceptionMapper.java).
+
+```java
+@Path("/examples")
+public interface ExampleClient {
+
+  @GET
+  @Path("/{exampleId}")
+  @Produces(MediaType.APPLICATION_JSON)
+  @SuppressHttpErrors(404)
```
Fox32, line 257: Do I understand it correctly that this annotation suppresses the exception and returns `null` instead?

Fox32: If yes: How about a name like `@ReturnNullForHttpErrors(404)`. On the other side I wonder if there aren't any REST Client Libraries for Java ecosystem that can already do this, instead of writing our own here.

JoergSiebahn: Yes, in this case only 404 Not Found errors. The _value_ is an array so that multiple status codes can be defined to be suppressed.

JoergSiebahn: As far as I know, JEE REST Clients are the most common clients in the Java ecosystem that use interfaces as alternative to fluent or other request builders. The clients based on the JEE standard are available in most frameworks like Dropwizard, Micronaut, Quarkus and monolithic JEE applications. One drawback we already worked around is that JEE uses the same exception types for client requests and server responses. Therefore we already added a layer around it so that developers carefully choose how to pass through client errors to their consumers. This PR just adds another layer of convenience to the common JEE implementation. It is based on the things we often repeat in our daily business. I think it can be transferred to other JEE implementations easily when we go for another framework. In my opinion this is far from „writing our own REST Client Library“.

`@ReturnNullFor*` is quite long compared to `@Suppress*`. What do the other @SDA-SE/sda-commons-maintainer  think? Is it more helpful to describe the result in the annotation? I'm fine with both variants.

christopher-cudennec: I'd vote for `@Suppress...` because I find it more common and the name contains what it does: suppressing an exception in case of http errors.

Fox32: What I miss about suppress is explicitly stating what happens instead. I had to read the code of this PR to understand it. Could also mean that another exception is thrown instead.

JoergSiebahn: There may be some inaccuracies in the wording but I think _suppress_ clearly states that no exception is thrown. What do you think about `@Suppress*ErrorsToNull`?

JoergSiebahn: I updated the PR to the pattern `@Suppress*ErrorsToNull`.

sda-commons-client-jersey/src/main/java/org/sdase/commons/client/jersey/proxy/ApiClientInvocationHandler.java
```diff
@@ -39,12 +60,79 @@ public Object invoke(Object proxy, Method method, Object[] args) throws Throwabl
...
+      suppressedStatusCodes.addAll(ALL_CLIENT_ERRORS);
+    }
+    if (suppressedHttpErrors.allServerErrors()) {
+      suppressedStatusCodes.addAll(ALL_SERVER_ERRORS);
+    }
+    return suppressedStatusCodes.contains(httStatusCode);
+  }
+
+  private boolean isProcessingErrorSuppressed(Method invokedMethod) {
+    return invokedMethod.getDeclaredAnnotation(SuppressProcessingErrorsToNull.class) != null;
```
christopher-cudennec, line 128: For readability I prefer `isAnnotationPresent`

christopher-cudennec (commented):
I am torn: On the one hand I find the solution interesting, on the other hand we should think carefully about which code we add, because we also have to maintain it.

Do we have a mechanism of flagging features as "beta" to try them out? I don't want to risk a breaking change if we come to another decision in the future regarding this feature.

sda-commons-client-jersey/src/main/java/org/sdase/commons/client/jersey/proxy/ApiClientInvocationHandler.java
JoergSiebahn: Fixed.

christopher-cudennec (approved):

PR #651, opened 2020-11-16
feat(starter): implement a new starter bundle with OpenApi 3 support

huebner-j:
This change introduces a new version of the `sda-commons-server-starter` module. The change contains the support of OpenApi 3. By integrating this module only OpenApi 3 annotations are supported in REST endpoint definitions (follow the the links for further details [1][2])

Due to the support of OpenApi 3 there are also changes to the SdaPlatformBundle. The creation of a dropwizard application differs slightly when using OpenApi 3. The following example shows how an application is created (see [3]):

```java
...

@OpenAPIDefinition(
    info =
        @Info(
            title = "SDA Platform Example Application",
            description =
                "This application is an example to show a developer how to create a SDA Dropwizard application",
            version = "1.1.1",
            contact =
                @Contact(name = "John Doe", email = "info@example.com", url = "j.doe@example.com")))
public class SdaPlatformExampleApplication extends Application<MyCustomConfiguration> {

  public static void main(String[] args) throws Exception {
    new SdaPlatformExampleApplication().run(args);
  }

  @override
  public void initialize(Bootstrap<MyCustomConfiguration> bootstrap) {
    bootstrap.addBundle(
        SdaPlatformBundle.builder()
            .usingSdaPlatformConfiguration(MyCustomConfiguration.class)
            // Additional Swagger documentation properties may be set here until
            // the packages that should be scanned for Swagger documentation annotations are defined
	    .addOpenApiResourcePackageClass(this.getClass())
            // or use an existing existing OpenAPI definition.
            // provide the path to the existing openapi file (yaml or json) in the classpath
            .withExistingOpenAPIFromClasspathResource("/custom-openapi.yaml")
            .build());
  }

  ...
```

Hint: Swagger / OpenApi 2 can be used by importing the module `sda-commons-server-starter-swagger`

[1] - https://github.com/SDA-SE/sda-dropwizard-commons/tree/master/sda-commons-server-openapi
[2] - https://github.com/swagger-api/swagger-core/wiki/Swagger-2.X---Annotations
[3] - https://github.com/SDA-SE/sda-dropwizard-commons/blob/master/sda-commons-server-starter-example/src/main/java/org/sdase/commons/server/starter/example/SdaPlatformExampleApplication.java

### DIFF
--- a/sda-commons-server-openapi/src/main/java/org/sdase/commons/server/openapi/OpenApiBundle.java
+++ b/sda-commons-server-openapi/src/main/java/org/sdase/commons/server/openapi/OpenApiBundle.java
@@ -219,13 +219,6 @@ public final class OpenApiBundle implements ConfiguredBundle<Configuration> {
   }
 
   public interface FinalBuilder extends InitialBuilder {
-    /**
-     * Disables automatic addition of the embed query parameter if embeddable resources are
-     * discovered.
-     *
-     * @return the builder.
-     */
-    FinalBuilder disableEmbedParameter();
 
     OpenApiBundle build();
   }
@@ -241,12 +234,6 @@ public final class OpenApiBundle implements ConfiguredBundle<Configuration> {
       addResourcePackageClass(HalLinkDescriptionModifier.class);
       addResourcePackageClass(EmbedParameterModifier.class);
       addResourcePackageClass(OpenAPISorter.class);
-    }
-
-    @Override
-    public FinalBuilder disableEmbedParameter() {
-      this.resourcePackages.remove(getResourcePackage(EmbedParameterModifier.class));
-      return this;
     }
 
     @Override
--- a/sda-commons-starter-example/README.md
+++ b/sda-commons-starter-example/README.md
@@ -1,0 +1,25 @@
+# SDA Commons Starter Example
+
+This example module shows an 
+[application](./src/main/java/org/sdase/commons/starter/example/SdaPlatformExampleApplication.java) that uses the 
+[`SdaPlatformBundle`](../sda-commons-starter/src/main/java/org/sdase/commons/starter/SdaPlatformBundle.java)
+to bootstrap a Dropwizard application for use in the SDA Platform.
+
+Beside the initialization of the bundle, it includes a 
+[REST endpoint](./src/main/java/org/sdase/commons/starter/example/people/rest/PersonEndpoint.java) to demonstrate
+registration of endpoints to map resources.
+
+The 
+[integration test](./src/test/java/org/sdase/commons/serr/starter/example/SdaPlatformExampleApplicationIT.java) 
+shows how the application is bootstrapped in tests. The tests show the capabilities of a standard platform application.
+
+The provided [`local-config.yaml`](./local-config.yaml) allows to start the 
+[application](./src/main/java/org/sdase/commons/starter/example/SdaPlatformExampleApplication.java) without the 
+need for authentication locally using a run configuration of the favourite IDE that defines the program arguments 
+`server sda-commons-starter-example/local-config.yaml`. Note that there will be no data available and the example
+application does not provide POST endpoints. All that is available is an empty array at `GET /people` and the simple
+Swagger documentation at `GET /swagger.json` or `GET /swagger.yaml`
+
+The [`config.yaml`](./config.yaml) is an example how the application can be started in production. Such file should be 
+copied in the Docker container so that the variables can be populated using the environment configured by the 
+orchestration tool (e.g. Kubernetes).   
--- a/sda-commons-starter/README.md
+++ b/sda-commons-starter/README.md
@@ -1,0 +1,116 @@
+# SDA Commons Starter
+
+[![javadoc](https://javadoc.io/badge2/org.sdase.commons/sda-commons-starter/javadoc.svg)](https://javadoc.io/doc/org.sdase.commons/sda-commons-starter)
+
+The module `sda-commons-starter` provides all basics required to build a service for the SDA Platform with
+Dropwizard.
+
+Apps built with the [`SdaPlatformBundle`](./src/main/java/org/sdase/commons/starter/SdaPlatformBundle.java)
+automatically contain
+
+- [Support for environment variables in configuration files and default console appender configuration](../sda-commons-server-dropwizard/README.md)
+- [Trace Token support](../sda-commons-server-trace/README.md)
+- [a tolerant `ObjectMapper`, HAL support and a field filter](../sda-commons-server-jackson/README.md)
+- [Security checks on startup](../sda-commons-server-security/README.md)
+- [Authentication support](../sda-commons-server-auth/README.md)
+- [Prometheus metrics](../sda-commons-server-prometheus/README.md)
+- [OpenApi documentation](../sda-commons-server-openapi/README.md)
+- [Open Tracing](../sda-commons-server-opentracing/README.md) and [Jaeger](../sda-commons-server-jaeger/README.md)
+
+They may be configured easily to
+
+- [require a Consumer token from clients](../sda-commons-server-consumer/README.md)
+- [allow cross origin resource sharing](../sda-commons-server-cors/README.md)
+- [use the Open Policy Agent for authorization](../sda-commons-server-auth/README.md)
+
+Using the [`SdaPlatformBundle`](./src/main/java/org/sdase/commons/starter/SdaPlatformBundle.java) is the easiest
+and fastest way to create a service for the SDA Platform.
+
+To bootstrap a Dropwizard application for the SDA Platform only the 
+[`SdaPlatformBundle`](./src/main/java/org/sdase/commons/starter/SdaPlatformBundle.java) has to be added. The 
+API should be documented with Swagger annotations: 
+
+```java
+import io.dropwizard.Application;
+import io.dropwizard.setup.Bootstrap;
+import io.dropwizard.setup.Environment;
+import io.swagger.v3.oas.annotations.OpenAPIDefinition;
+import SdaPlatformBundle;
+import SdaPlatformConfiguration;
+
+@OpenAPIDefinition(info = @Info(
+    title = "My First App",
+    description =
+        "The description of my first application",
+    version = "1.0.0",
+    contact =
+    @Contact(
+        name = "John Doe",
+        email = "info@example.com",
+        url = "j.doe@example.com"),
+    license = @License(name = "Sample License")))
+public class MyFirstApp extends Application<SdaPlatformConfiguration> {
+
+   public static void main(String[] args) throws Exception {
+      new MyFirstApp().run(args);
+   }
+
+   @Override
+   public void initialize(Bootstrap<SdaPlatformConfiguration> bootstrap) {
+      bootstrap.addBundle(SdaPlatformBundle.builder()
+            .usingSdaPlatformConfiguration()
+            .withRequiredConsumerToken()
+            // more Swagger data that may also be added with annotations
+            .addSwaggerResourcePackageClass(this.getClass())
+            // or use an existing OpenApi definition
+            .withExistingOpenAPI("openApiJsonOrYaml")
+            .build());
+   }
+
+   @Override
+   public void run(SdaPlatformConfiguration configuration, Environment environment) {
+      environment.jersey().register(MyFirstEndpoint.class);
+   }
+
+}
+```
+
+The [`SdaPlatformConfiguration`](./src/main/java/org/sdase/commons/starter/SdaPlatformConfiguration.java) may be
+extended to add application specific configuration properties.
+
+The `config.yaml` of the 
+[`SdaPlatformConfiguration`](./src/main/java/org/sdase/commons/starter/SdaPlatformConfiguration.java) supports
+configuration of [authentication](../sda-commons-server-auth/README.md) and [CORS](../sda-commons-server-cors/README.md)
+additionally to the defaults of Dropwizard's `Configuration`:
+
+```yaml
+
+# See sda-commons-server-auth
+auth:
+  disableAuth: ${DISABLE_AUTH:-false}
+  leeway: ${AUTH_LEEWAY:-0}
+  keys: ${AUTH_KEYS:-[]}
+
+# See sda-commons-server-cors
+cors:
+  # List of origins that are allowed to use the service. "*" allows all origins
+  allowedOrigins:
+    - "*"
+  # Alternative: If the origins should be restricted, you should add the pattern
+  # allowedOrigins:
+  #    - https://*.sdase.com
+  #    - https://*test.sdase.com
+  # To use configurable patterns per environment the Json in Yaml syntax may be used with an environment placeholder:
+  # allowedOrigins: ${CORS_ALLOWED_ORIGINS:-["*"]}
+```
+
+Instead of `.usingSdaPlatformConfiguration()`, the configuration may be fully customized using 
+`.usingCustomConfig(MyCustomConfiguration.class)` to support configurations that do not extend 
+[`SdaPlatformConfiguration`](./src/main/java/org/sdase/commons/starter/SdaPlatformConfiguration.java). This may 
+also be needed to disable some features of the starter module or add special features such as
+Authorization.
+
+Please note that `.withOpaAuthorization(MyConfiguration::getAuth, MyConfiguration::getOpa)`
+will configure the `AuthBundle` to use `.withExternalAuthorization()`. Please read the 
+[documentation of the Auth Bundle](../sda-commons-server-auth/README.md) carefully before
+using this option.
--- a/sda-commons-starter/README.md
+++ b/sda-commons-starter/README.md
@@ -91,6 +91,12 @@ auth:
   leeway: ${AUTH_LEEWAY:-0}
   keys: ${AUTH_KEYS:-[]}
 
+opa:
+  disableOpa: ${OPA_DISABLE:-false}
+  baseUrl: ${OPA_URL:-http://localhost:8181}
+  policyPackage: ${OPA_POLICY_PACKAGE:-<your_package>}
+  readTimeout: ${OPA_READ_TIMEOUT:-500}
+
 # See sda-commons-server-cors
 cors:
   # List of origins that are allowed to use the service. "*" allows all origins
@@ -104,7 +110,10 @@ cors:
   # allowedOrigins: ${CORS_ALLOWED_ORIGINS:-["*"]}
 ```
 
-Instead of `.usingSdaPlatformConfiguration()`, the configuration may be fully customized using 
+By using `.usingSdaPlatformConfiguration()` or `.usingSdaPlatformConfiguration(MyCustomConfiguration.class)` 
+the authorization including the open policy agent are automatically enabled as well as the cors settings. 
+
+Instead of `.usingSdaPlatformConfiguration()` and `.usingSdaPlatformConfiguration(MyCustomConfiguration.class)`, the configuration may be fully customized using 
 `.usingCustomConfig(MyCustomConfiguration.class)` to support configurations that do not extend 
 [`SdaPlatformConfiguration`](./src/main/java/org/sdase/commons/starter/SdaPlatformConfiguration.java). This may 
 also be needed to disable some features of the starter module or add special features such as
--- a/sda-commons-starter/README.md
+++ b/sda-commons-starter/README.md
@@ -2,6 +2,11 @@
 
 [![javadoc](https://javadoc.io/badge2/org.sdase.commons/sda-commons-starter/javadoc.svg)](https://javadoc.io/doc/org.sdase.commons/sda-commons-starter)
 
+> #### ⚠️ Experimental ⚠
+>
+> Please be aware that this API is in an early stage and might change in the future.
+>
+
 The module `sda-commons-starter` provides all basics required to build a service for the SDA Platform with
 Dropwizard.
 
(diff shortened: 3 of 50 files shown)


sda-commons-starter/build.gradle
```diff
@@ -0,0 +1,22 @@
...
+  compile project(':sda-commons-server-prometheus')
+  compile project(':sda-commons-server-security')
+  compile project(':sda-commons-server-openapi:')
+  compile project(':sda-commons-server-trace')
+  compile project(':sda-commons-server-healthcheck')
+  compile project(':sda-commons-server-jaeger')
+  compile project(':sda-commons-server-opentracing')
+
+  testCompile project(':sda-commons-server-testing')
+  testCompile 'org.assertj:assertj-core'
```
dhenneke, line 16: Is already imported by `sda-commons-server-testing`.

sda-commons-starter/build.gradle
```diff
@@ -0,0 +1,22 @@
+dependencies {
+  compile project(':sda-commons-server-auth')
+  compile project(':sda-commons-server-consumer')
+  compile project(':sda-commons-server-cors')
+  compile project(':sda-commons-server-dropwizard')
+  compile project(':sda-commons-server-jackson')
+  compile project(':sda-commons-server-prometheus')
+  compile project(':sda-commons-server-security')
+  compile project(':sda-commons-server-openapi:')
```
dhenneke, line 9: No `:` at the end.

sda-commons-starter/build.gradle
```diff
@@ -0,0 +1,22 @@
+dependencies {
```
dhenneke, line 1: Very minor: We could sort this.

sda-commons-starter/src/main/java/org/sdase/commons/starter/SdaPlatformBundle.java
```diff
@@ -0,0 +1,352 @@
...
+          opaBundleBuilder,
+          corsBundleBuilder,
+          consumerTokenBundleBuilder,
+          openApiBundleBuilder);
+    }
+
+    // InitialBuilder
+
+    @Override
+    public ConsumerTokenConfigBuilder<SdaPlatformConfiguration> usingSdaPlatformConfiguration() {
```
dhenneke, line 152: Open for discussion: Since most applications derive their configuration from `SdaPlatformConfiguration` it would be nice to give them a shortcut like `public <T extends SdaPlatformConfiguration> AuthConfigProviderBuilder<T> usingSdaPlatformConfiguration(Class<T> configurationClass)`. I would also like to enable OPA by default in this builder. No OPA would require to use the `usingCustomConfig()` branch.

sda-commons-starter/src/main/java/org/sdase/commons/starter/SdaPlatformBundle.java
```diff
@@ -0,0 +1,352 @@
...
+      return this;
+    }
+
+    @Override
+    public OpenApiInitialBuilder<C> withConsumerTokenConfigProvider(
+        ConsumerTokenConfigProvider<C> consumerTokenConfigProvider) {
+      this.consumerTokenBundleBuilder =
+          ConsumerTokenBundle.builder().withConfigProvider(consumerTokenConfigProvider);
+      return this;
+    }
```
dhenneke, line 233: Open for discussion: Is there ever a need to make this configurable? I would also vote to remove it from the default `SdaPlatformConfiguration` class if it is not the default.

sda-commons-starter/src/main/java/org/sdase/commons/starter/SdaPlatformBundle.java
```diff
@@ -0,0 +1,352 @@
...
+
+    @Override
+    public OpenApiInitialBuilder<C> withExcludePatternsForRequiredConsumerToken(String... regex) {
```
dhenneke, line 236: Is this used often?

sda-commons-starter/src/main/java/org/sdase/commons/starter/SdaPlatformBundle.java
```diff
@@ -0,0 +1,352 @@
...
+      return this;
+    }
+
+    @Override
+    public PlatformBundleBuilder<C> withCorsAdditionalExposedHeaders(
+        String... additionalExposedHeaders) {
+      validateConfigureCors();
+      this.corsBundleBuilder.withAdditionalExposedHeaders(additionalExposedHeaders);
+      return this;
+    }
```
dhenneke, line 274: Open for discussion: Do we prefer to do this in code or in config?

sda-commons-starter/src/main/java/org/sdase/commons/starter/SdaPlatformBundle.java
```diff
@@ -0,0 +1,352 @@
...
+
+    @Override
+    public OpenApiFinalBuilder<C> addOpenApiResourcePackageClass(Class<?> swaggerResourcePackage) {
+      this.openApiBundleBuilder =
+          getOpenApiBuilder().addResourcePackageClass(swaggerResourcePackage);
+      return this;
+    }
+
+    @Override
+    public OpenApiFinalBuilder<C> disableSwaggerEmbedParameter() {
```
dhenneke, line 334: I never seen someone using it and nobody complained about it in the last 19 months it is present, so we should remove the option.

Since the OpenAPI module is still "experimental" we might remove it from there too.

JoergSiebahn: As the embed parameter is only documented if the response type has embeddable objects, this setting could lead to wrong documentation. In my opinion, it could be removed in both modules.

Fox32:
I can't ignore the code duplications. Maybe we set it in gradle instead?

sda-commons-starter/build.gradle
huebner-j: I've removed the dependency..

sda-commons-starter/build.gradle
huebner-j: I have no clue how this came about.. Removed!

sda-commons-starter/src/main/java/org/sdase/commons/starter/SdaPlatformBundle.java
huebner-j: I will investigate..

sda-commons-starter/src/main/java/org/sdase/commons/starter/SdaPlatformBundle.java
huebner-j: If nobody is using this option, we should definitely remove it... Done :)

sda-commons-starter/src/main/java/org/sdase/commons/starter/SdaPlatformBundle.java
JoergSiebahn: I could not find the method name in the whole SDA GitHub ORG apart from the commons repo.

sda-commons-starter/src/main/java/org/sdase/commons/starter/SdaPlatformBundle.java
huebner-j: >  In my opinion, it could be removed in both modules.

@JoergSiebahn isn't it a breaking change if we remove the method from the `sda-commons-server-starter` modul, even though nobody is actually using it?

JoergSiebahn: I think, we can safely remove it `disableEmbedParameter` from `OpenApiBundle` which is experimental and the new `SdaPlatformBundle` in `sda-commons-starter`.
You are right, we should NOT remove it in `sda-commons-server-starter` or `sda-commons-server-swagger`.
Maybe I should have clarified  _„both modules“_.

sda-commons-starter/src/main/java/org/sdase/commons/starter/SdaPlatformBundle.java
Fox32: I don't think that this is part of the deployment. This is part of the implementation. There shouldn't be any need to expose additional headers if the implementation doesn't require it. If you still want to override the default behavior of the implementation, use an ingress.

JoergSiebahn: Some services in our org use it to allow _"Content-Range"_ for a list result instead of providing paging parameters in a response model. Do you know other cases? This example does not match our Rest Guide, I think.

Fox32: `Content-Range` is not designed for paging as it requests specific bytes in a resource, but I won't judge this. My point is, that if the REST Api has such a feature, the REST API implementer has to allow the access to the header. We should let the consumer figure out that it doesn't work.

We had some implementations that expose the total count for paging as a header, but even there we have another recommendation in the REST guide.

sda-commons-starter/src/main/java/org/sdase/commons/starter/SdaPlatformBundle.java
huebner-j: I've created a PR for this change. Let's discuss the suggestion there. 
[https://github.com/SDA-SE/sda-dropwizard-commons/pull/654](https://github.com/SDA-SE/sda-dropwizard-commons/pull/654)

sda-commons-starter/src/main/java/org/sdase/commons/starter/SdaPlatformBundle.java
huebner-j: I couldn't find any use of this method either. I'll remove it from the builder..

huebner-j: Team @SDA-SE/communications uses this method [here](https://github.com/SDA-SE/malware-scanner-facade/blob/e36436bd4531a2376e6ebcbce093142bb1897415/malware-scanner-service/src/main/java/com/sdase/malware/scanner/facade/MalwareScannerFacadeApp.java#L88). So the answer is, that it is not often used, it is used exactly once.

christopher-cudennec: ... at least in our organization. Can't speak for other customers.

huebner-j: That's true. But I think, the decision was made by your use of the method, that we leave the method as it is.

JoergSiebahn: Oops, I missed that match when looking into the [search results](https://github.com/search?q=org%3ASDA-SE+withExcludePatternsForRequiredConsumerToken&type=code). Thanks, that you checked it again.

sda-commons-starter/src/main/java/org/sdase/commons/starter/SdaPlatformBundle.java
dhenneke: I thought that this was also part of the configuration class but it isn't. So no need for further actions.

sda-commons-starter/src/main/java/org/sdase/commons/starter/SdaPlatformBundle.java
huebner-j: Ok perfect, I've removed it from the `OpenApiBundle` & the `sda-commons-starter` builder..

sda-commons-starter/build.gradle
huebner-j: What criteria do you like to sort the dependencies by? By name?

sda-commons-starter/src/main/java/org/sdase/commons/starter/SdaPlatformBundle.java
```diff
@@ -0,0 +1,346 @@
...
+      consumerTokenConfig = new ConsumerTokenConfig();
+      consumerTokenConfig.setOptional(false);
+      this.consumerTokenBundleBuilder =
+          ConsumerTokenBundle.builder().withConfigProvider(c -> consumerTokenConfig);
+      return this;
+    }
+
+    @Override
+    public OpenApiInitialBuilder<C> withConsumerTokenConfigProvider(
+        ConsumerTokenConfigProvider<C> consumerTokenConfigProvider) {
```
huebner-j, line 229: Open for discussion: When I first used the `ConsumerToken` configuration in the starter bundle I was wondering why there is no `consumerToken` in the configuration until i understood that this consumer token configuration is just for making consumer tokens optional or required for incoming requests. I think the naming of the configuration is quite confusing since the name is `withConsumerTokenConfigProvider(SdaPlatformConfiguration::getConsumerToken)`. When I read `getConsumerToken` then I would expect the actual consumer token and not a configuration. Furthermore if you would like to define a consumer token for outgoing calls, you have to name it like 'consumerTokenName'. I would vote for a renaming here to make clear that this configuration is for incoming requests.

sda-commons-starter/src/main/java/org/sdase/commons/starter/SdaPlatformBundle.java
huebner-j: I've searched in our organization and found some occurrences:  
[1](https://github.com/SDA-SE/edithcare-participation-service/blob/fa61f6d544d833373d895721b3b803ad60ff4245/config.yml#L21)
[2](https://github.com/SDA-SE/customer-service/blob/fea7b0fb679a51d2835b0621940fc91ea31ab0fc/src/integTest/resources/test-config.yml#L16)
[3](https://github.com/SDA-SE/billing-external/blob/15fe3b544b7606f3dcfd68903789609a6f55f893/billing-external-service/config.yaml#L10)
And it seems some others are using it as well. [search in our organization](https://github.com/search?l=YAML&q=org%3ASDA-SE+consumerToken&type=Code)

I would also vote to remove it from the `SdaPlatformConfiguration`.

sda-commons-starter/src/main/java/org/sdase/commons/starter/SdaPlatformBundle.java
christopher-cudennec: I've also heard from different people that the name is not intuitively clear.

Although the name is used according to other config properties. E.g. the mongo configuration is mostly added as "mongo".

What would be a better name? `consumerTokenSettings`? `consumerTokenServer`?

JoergSiebahn: The consumer token or consumer name seems like a workaround since I'm working for the SDA. At the moment it is only used for logging. I wonder if it is needed to set it technically as required in a service while it is a (temporary?) required pattern for the platform in general that may change when we introduce ISTIO.
It was meant to require identification of the calling service which should be part of the authorization we do in OPA. Actually it is nothing more than an indicator of who is calling as it can't be verified. I know moving that decision to OPA and not requiring it in the service may have a big impact on existing services but gives more flexibility for the future. I'd like to hear about @Fox32 opinion. I think he has quite a lot of insights about the requirements of the consumer token, the use in current services at SDA and at developing customers and about the future evolution of service authentication in context of a service mesh. Maybe @dhenneke can also point out some arguments about moving such a decision to OPA.

huebner-j:
> I can't ignore the code duplications. Maybe we set it in gradle instead?

@Fox32 all files from `sda-commons-server-starter` are excluded from copy & paste detection by this property: [https://github.com/SDA-SE/sda-dropwizard-commons/blob/feat/new-starter-module/build.gradle#L190](https://github.com/SDA-SE/sda-dropwizard-commons/blob/feat/new-starter-module/build.gradle#L190). Is it fine for you?

sda-commons-starter/src/main/java/org/sdase/commons/starter/SdaPlatformBundle.java
dhenneke: My main concern with `withConsumerTokenConfigProvider(...)` is that we have two options to configure the consumer token: You can use `withoutConsumerTokenSupport()` / `withRequiredConsumerToken()` / `withOptionalConsumerToken()` OR the `withConsumerTokenConfigProvider(...)`. My current understanding of the consumer token (and I guess the majority selection) is that the service should decide whether it want to require a consumer token during build-time. We shouldn't give the users of the `SdaPlatformBundle` the choice between deciding it in code vs. deciding it in config, regardless of the decision whether it should be optional or required. Especially since there is no guidance on what the preferred way is. (And I agree that it is very easy to confuse the server-setting and the client-setting in a configuration).

About the need, our wiki cites:
> A consumer token is used to identify the consumer of a service. This is used to aggregate metrics by consumer, to filter logs, and for billing purposes. The consumer token should be provided with every request in form of the `Consumer-Token` request header. If no consumer token is provided, the access is denied with status `401 - UNAUTHORIZED`.

It's not really an actual authentication measure but more a technical/organizational guideline. In this regard, maybe the `401` was not the best choice, but that shouldn't be discussed in this PR.

christopher-cudennec: So your conclusion / recommendation is to remove `withConsumerTokenConfigProvider`?

dhenneke: Or remove the other three? Or we decide that this needs a deeper look and decide it in the future.

Fox32: As @dhenneke already said: Consumer-Token is only used for logging and metrics. So more or less a simple version of tracing. Maybe we could even make it completely optional and not require it at all. I even wonder how this works well with customers that don't use the consumer token in the infrastructure. That is probably also the reason for making it configurable at deployment time. Otherwise I would vote for making it an implementation time decision - so remove the option. However I'm not sure if the customers will like that.

JoergSiebahn: > Maybe we could even make it completely optional and not require it at all.

That would mean, we make _withOptionalConsumerToken()_ the default and remove all other options?

I would like that approach. We could add the other options if some requests for it.

sda-commons-starter/src/main/java/org/sdase/commons/starter/SdaPlatformBundle.java
huebner-j: Please take a look at [PR-662](https://github.com/SDA-SE/sda-dropwizard-commons/pull/662)

sda-commons-starter/src/main/java/org/sdase/commons/starter/SdaPlatformBundle.java
huebner-j: Please take a look at [PR-662](https://github.com/SDA-SE/sda-dropwizard-commons/pull/662)

JoergSiebahn (dismissed):

sda-commons-starter-example/README.md
```diff
@@ -0,0 +1,25 @@
+# SDA Commons Starter Example
```
dhenneke, line 1: As a next step, we should think about aligning this example nearer to our template-project and maybe also make an effort in improving the overall documentation, if needed.

sda-commons-starter/README.md
```diff
@@ -0,0 +1,128 @@
...
+    title = "My First App",
+    description =
+        "The description of my first application",
+    version = "1.0.0",
+    contact =
+    @Contact(
+        name = "John Doe",
+        email = "info@example.com",
+        url = "j.doe@example.com"),
+    license = @License(name = "Sample License")))
```
dhenneke, line 55: We often store all rest-interfaces in a separate package that should also contain this definition. It is not always the best idea to register the application as a starting point. Sometimes the teams use generated client-code which also includes openapi-annotations. And we don't want these client interfaces be part of the OpenAPI file of the service.

But nothing we need to fix now.

sda-commons-starter/README.md
```diff
@@ -0,0 +1,128 @@
...
+auth:
+  disableAuth: ${DISABLE_AUTH:-false}
+  leeway: ${AUTH_LEEWAY:-0}
+  keys: ${AUTH_KEYS:-[]}
+
+opa:
+  disableOpa: ${OPA_DISABLE:-false}
+  baseUrl: ${OPA_URL:-http://localhost:8181}
+  policyPackage: ${OPA_POLICY_PACKAGE:-<your_package>}
+  readTimeout: ${OPA_READ_TIMEOUT:-500}
```
dhenneke, line 101: This is a deprecated field. Might also be changed in a documentation-follow-up.

sda-commons-server-openapi/src/main/java/org/sdase/commons/server/openapi/OpenApiBundle.java
```diff
@@ -219,13 +219,6 @@ default FinalBuilder withExistingOpenAPIFromClasspathResource(String path) {
   }
 
   public interface FinalBuilder extends InitialBuilder {
-    /**
-     * Disables automatic addition of the embed query parameter if embeddable resources are
-     * discovered.
-     *
-     * @return the builder.
-     */
-    FinalBuilder disableEmbedParameter();
```
dhenneke, line 228: It would be nice to have this change in a separate commit with the correct `feat(server-openapi)` scope. Or can we leave it as-is @SDA-SE/sda-commons-maintainer?

dhenneke (commented):
Thanks 🏆 

I would approve, but am happy to wait for the feedback from others.

PS: We should remember to add the experimental notice after the release has been created, similar to https://github.com/SDA-SE/sda-dropwizard-commons/releases/tag/2.3.0.

JoergSiebahn (approved):

sda-commons-server-openapi/src/main/java/org/sdase/commons/server/openapi/OpenApiBundle.java
christopher-cudennec: Would be nice to see it in an extra commit. Last request for change, promise! 😃

huebner-j: done :)

dhenneke (approved):

Fox32 (approved):
👍

PR #652, opened 2020-11-16
feat(jackson): Implement ExceptionMapper modifier for JacksonConfigurationBundle

rschwedasi:
Hey folks,
for us, using hard coded exception mappers in your Jackson configuration bundle is a bummer. Not being able to add custom modifications to the exception handling for already registered exception types, makes maintaining our services impossible (e.g. lack of logging statements in case of bad Requests provided by our own applications and so on).

This PR facilitates a way of handling default exception mapper registration for future users of your bundle. Please feel free to comment.

### DIFF
--- a/sda-commons-server-jackson/src/main/java/org/sdase/commons/server/jackson/JacksonConfigurationBundle.java
+++ b/sda-commons-server-jackson/src/main/java/org/sdase/commons/server/jackson/JacksonConfigurationBundle.java
@@ -8,7 +8,10 @@ import io.dropwizard.server.AbstractServerFactory;
 import io.dropwizard.setup.Bootstrap;
 import io.dropwizard.setup.Environment;
 import java.time.ZonedDateTime;
+import java.util.ArrayList;
+import java.util.List;
 import java.util.function.Consumer;
+import javax.ws.rs.ext.ExceptionMapper;
 import org.sda.commons.server.jackson.hal.HalLinkProvider;
 import org.sdase.commons.server.jackson.errors.ApiExceptionMapper;
 import org.sdase.commons.server.jackson.errors.EarlyEofExceptionMapper;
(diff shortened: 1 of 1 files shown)


JoergSiebahn:
We provide these exception mappers by intend. The main purpose is to avoid personal information in log files and providing information about the used libraries to attackers.

I think the brute force way to manipulate an internal list by the consumer is very flexible but can lead to unintended mistakes that we are trying to mitigate for the user of sda commons by default.

I would prefer a more targeted approach that more precisely addresses the problem you have. Could you provide an example of the changes you would do to the exception mappers list with the introduced API in the bundle?

rschwedasi:
Thank you for your feedback. Your main purpose of not logging personal information is a blocker for us. We need that actual information to be able to trace unsuccessful contract requests coming in from sales or consumers via our online contract services. The most common use case in our application is a bad request due to failing validations on older or unsupported clients which often lead to incidents we have to retrace and report to our customer service.

Therefore we desperately need logging / tracing capabilities of some kind in our exception handlers despite of the amount of personal information logged, it might even be only technical data. So if overriding the registered mappers is not an option, do you see some other solutions? I don't know, lifecycle-handlers in the exception mappers itself e.g.?

christopher-cudennec:
> The most common use case in our application is a bad request due to failing validations on older or unsupported clients 
In other words: This is due to missing backwards compatibility?

What about additional logging for 422 responses, e.g. the fields with validation errors? That is something that I've missed a couple of times for other services in the past. Would that help in your case?

JoergSiebahn:
We discussed a bit internally and we are not that happy with the approach to modify the whole list and need to collect some more ideas.

To fix your issue in a timely manner you may try the following workaround:

- Create a custom bundle that registers your own ExceptionMappers in Jersey
- Add this Bundle before the JacksonConfigurationBundle in the `bootstrap` method.

The earlier registered ExceptionMappers will be used by Jersey instead of the later registered ones.

Note: Registering in `run` of the application will not work, as it is executed after `run` of the bundles.

rschwedasi:
@JoergSiebahn Ok, we will stick to our own bundle then. Thank you.
@christopher-cudennec Yes, something like Dropwizards LoggingExceptionMapper with customizable log-levels would be a great start for us.

PR #653, opened 2020-11-17
feat(server-jackson): Log fields with validation errors before responding with status 422

christopher-cudennec:
As discussed in #652


sda-commons-server-jackson/src/main/java/org/sdase/commons/server/jackson/errors/JerseyValidationExceptionMapper.java
```diff
@@ -43,6 +48,11 @@ public Response toResponse(JerseyViolationException e) {
                           CaseFormat.UPPER_UNDERSCORE,
                           annotation.substring(annotation.lastIndexOf('.') + 1)));
               invalidParameters.add(invalidParameter);
+
+              LOGGER.debug(
```
JoergSiebahn, line 52: I think if it should help someone, we need info level.

If there are many constraint failures, you get a lot of log messages for a single failed request.
Usually one would introduce toString in the ApiError to log the result properly. But it should not contain all fields.

christopher-cudennec: > I think if it should help someone, we need info level.

I changed it to INFO.

christopher-cudennec: > If there are many constraint failures, you get a lot of log messages for a single failed request.

Changed it to only one statement.

JoergSiebahn (dismissed):

JoergSiebahn (approved):

PR #659, opened 2020-11-19
fix: deprecate accidentally introduced package `org.sda.`

JoergSiebahn:
All classes have been moved to `org.sdase.`

The change should be backward compatible therefore the tests still point to the deprecated classes.

Once the deprecated code is removed, the new exception must be thrown and the test must use the types from the same package.


jeskosda (approved):
Thanks for being investigative 👍

christopher-cudennec (approved):

PR #665, opened 2020-11-23
chore: add catalog-info.yaml

Fox32:
This is required to make the library show up in Backstage.

**Note:** after merging it, we have to register it.

### DIFF
--- a/catalog-info.yaml
+++ b/catalog-info.yaml
@@ -1,0 +1,15 @@
+apiVersion: backstage.io/v1alpha1
+kind: Component
+metadata:
+  name: sda-dropwizard-commons
+  annotations:
+    github.com/project-slug: 'sda-se/sda-dropwizard-commons'
+    fossa.io/project-name: sda-dropwizard-commons
+    sonarqube.org/project-key: SDA-SE_sda-dropwizard-commons
+  description: |
+    SDA Dropwizard Commons is a set of libraries to bootstrap services easily that follow
+    the patterns and specifications promoted by the SDA SE.
+spec:
+  type: library
+  lifecycle: production
+  owner: plattform-team


catalog-info.yaml
```diff
@@ -0,0 +1,15 @@
...
+kind: Component
+metadata:
+  name: sda-dropwizard-commons
+  annotations:
+    github.com/project-slug: 'sda-se/sda-dropwizard-commons'
+    fossa.io/project-name: sda-dropwizard-commons
+    sonarqube.org/project-key: SDA-SE_sda-dropwizard-commons
+  description: |
+    SDA Dropwizard Commons is a set of libraries to bootstrap services easily that follow
+    the patterns and specifications promoted by the SDA SE.
```
JoergSiebahn, line 11: I would prefer to clarify the backend aspect and to name the technology. Is Markdown allowed?
Then something like this would be nice:
```suggestion
    SDA Dropwizard Commons is a set of libraries to bootstrap Java backend services easily.
    It configures [Dropwizard](https://www.dropwizard.io/) applications to follow the patterns 
    and specifications promoted by the SDA SE.
```

catalog-info.yaml
```diff
@@ -0,0 +1,15 @@
...
+spec:
+  type: library
+  lifecycle: production
+  owner: plattform-team
```
JoergSiebahn, line 15: _pla_**t**_form_(-team) like the GitHub group?

Fox32: No, that is wrong. The name is plattform-team

catalog-info.yaml
Fox32: That field doesn't support markdown and should be a short description. Maybe we should make it even shorter.

JoergSiebahn (dismissed):
We should align GitHub groups and promoted team names.
See https://github.com/orgs/SDA-SE/teams/platform and https://sda-se.atlassian.net/wiki/spaces/platform/pages/982909012/Platform+2020 and https://sda-se.atlassian.net/wiki/spaces/platform/overview?homepageId=881160 and https://sda-se.atlassian.net/wiki/spaces/platform/pages/701562881/Platform+Team+2019 and https://sda-se.atlassian.net/secure/RapidBoard.jspa?rapidView=105&projectKey=PLP&view=detail

Where is a team named Plattform-Team?

dhenneke (approved):

PR #666, opened 2020-11-24
document how to keep an updated copies of the api documentation in the service repository

dhenneke:
We often use the pattern that the services want to have the latest `openapi.yaml` or `asyncapi.yaml` under source control. This PR adds helper methods and a custom assertion to minimize the code that is needed in the service repositories.

Docs and helpers are added for: OpenAPI, Swagger, AsyncAPI.

Example for an `openapi.yaml`:

```java
public class OpenApiDocumentationTest {

  @Test
  public void shouldHaveSameOpenApiInRepository() throws IOException {
    // receive the openapi.yaml from your service
    String expected = ...; // TODO: get your openapi.yaml from the server

    // specify where you want your file to be stored
    Path filePath = Paths.get("openapi.yaml");

    // check and update the file
    GoldenFileAssertions.assertThat(filePath)
        .hasContentAndUpdateGolden(normalizeOpenApiYaml(expected));
  }
}
```

Example for an `asyncapi.yaml`:
```java
public class AsyncApiDocumentationTest {
  @Test
  public void generateAndVerifySpec() throws IOException {
    String expected = AsyncApiGenerator
            .builder()
            .withAsyncApiBase(BaseEvent.class.getResource("/asyncapi.yaml"))
            .withSchema("./schema.json", BaseEvent.class)
            .generateYaml();

    // specify where you want your file to be stored
    Path filePath = Paths.get("asyncapi.yaml");

    // check and update the file
    GoldenFileAssertions.assertThat(filePath).hasContentAndUpdateGolden(expected);
  }
}
```

Plus: I found a test that I missed in my refactorings in #628 and that keeps failing in the builds.


sda-commons-server-testing/src/main/java/org/sdase/commons/server/testing/GoldenFileAssertions.java
```diff
@@ -0,0 +1,87 @@
...
+          .as(
+              "The current %s file is not up-to-date. If this happens locally, "
+                  + "just run the test again. The %s file is updated automatically after running "
+                  + "this test. If this happens in the CI, make sure that you have committed the "
+                  + "latest %s file!",
+              fileName, fileName, fileName)
+          .hasContent(expected);
+    } finally {
+      // always update the file content
+      Files.write(actual, expected.getBytes());
```
JoergSiebahn, line 82: As we promote this assertion for API specs, should we provide a specific Yaml-Assertion and write the file only if it differs semantically (or does not exist for initial setup)?
We had many cases in the past where no semantic changes occur due to technical reasons in the past. That caused useless changes that appeared in reviews.

Fox32: We had several PRs in the past that fixed the non-semantic differences. At least from our experience they don't exist anymore.
In the base for this code we used Yaml based recursive assertions, but the output was really hard to understand. The `hasContent` assertion has a nice output that humans can understand.

sda-commons-server-testing/README.md
```diff
@@ -21,14 +21,42 @@ For testing some frameworks are included:
 | `org.mockito`    | `mockito-core`       | 2.23.0  |
 | `org.assertj`    | `assertj-core`       | 3.11.1  |
```
Fox32, line 22: Not related to this PR, but we should remove the table above 😉

Fox32 (approved):
👍

sda-commons-server-testing/src/main/java/org/sdase/commons/server/testing/GoldenFileAssertions.java
JoergSiebahn: Well, I see such changes regularly in PRs.

sda-commons-server-testing/README.md
Fox32: See #668

sda-commons-server-testing/src/main/java/org/sdase/commons/server/testing/GoldenFileAssertions.java
Fox32: Then we should fix the root cause instead of patching it somewhere else

christopher-cudennec (approved):

sda-commons-server-testing/src/main/java/org/sdase/commons/server/testing/GoldenFileAssertions.java
dhenneke: I added an `assertThatYaml` variant that ignores key-order. But I would also prefer to fix the root-cause since I also seen a large number of PRs that update the file even-though there were no updates. And that makes the review harder.

sda-commons-server-testing/src/main/java/org/sdase/commons/server/testing/GoldenFileAssertions.java
```diff
@@ -0,0 +1,120 @@
...
+    try {
+      // assert if exists
+      Assertions.assertThat(actual).as(ASSERTION_TEXT, fileName, fileName, fileName).exists();
+
+      if (asYaml) {
+        // assert YAML / JSON
+        ObjectMapper objectMapper = Jackson.newObjectMapper(new YAMLFactory());
+        Assertions.assertThat(objectMapper.readTree(actual.toFile()))
+            .as(ASSERTION_TEXT, fileName, fileName, fileName)
+            .isEqualTo(objectMapper.readTree(expected));
```
dhenneke, line 105: The `ObjectNode#equals()` method is smart enough to do what we want. It also gives a more-or-less useful output.

sda-commons-server-testing/src/main/java/org/sdase/commons/server/testing/GoldenFileAssertions.java
```diff
@@ -0,0 +1,120 @@
...
+      } else {
+        // assert normal text
+        Assertions.assertThat(actual)
+            .as(ASSERTION_TEXT, fileName, fileName, fileName)
+            .hasContent(expected);
+      }
+
+    } finally {
+      // always update the file content
+      Files.write(actual, expected.getBytes());
```
christopher-cudennec, line 115: Small observation: `getBytes` should always be used with correct encoding, i.e. usually UTF-8.

dhenneke: fixed it!

JoergSiebahn (approved):

sda-commons-server-testing/src/main/java/org/sdase/commons/server/testing/GoldenFileAssertions.java
christopher-cudennec: 🦄

PR #670, opened 2020-11-25
fix(server-testing): refactor the golden-file-assertions api to match the assertThat() pattern

dhenneke:
Take two on #666.

This refactors
```java
GoldenFileAssertions.assertThat(filePath).hasContentAndUpdateGolden(expected);
GoldenFileAssertions.assertThatYaml(filePath).hasContentAndUpdateGolden(expected);
```

to
```java
GoldenFileAssertions.assertThat(filePath).hasContentAndUpdateGolden(expected);
GoldenFileAssertions.assertThat(filePath).hasYamlContentAndUpdateGolden(expected);
```

to better match the AssertJ style.

This also adds a clarification that one should prefer the first option since it has better outputs. The AsyncAPI and OpenAPI generations are automatically sorted and stable. The swagger export is problematic and not easy to fix.

PS: I cancelled the old release to not need a breaking change.

### DIFF
--- a/sda-commons-server-swagger/README.md
+++ b/sda-commons-server-swagger/README.md
@@ -151,14 +151,16 @@ public class SwaggerDocumentationTest {
 
     // check and update the file
     GoldenFileAssertions.assertThat(filePath)
-        .hasContentAndUpdateGolden(normalizeSwaggerYaml(expected));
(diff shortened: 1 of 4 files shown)


Fox32 (approved):
👍

PR #679, opened 2020-11-27
test: use the randomPorts config override instead of test-config.yaml files

dhenneke:
This reverts some changes of #628 where we needed to add a number of `test-config.yaml` files to configure random ports. Dropwizard 2.0.16 includes a new config override `ConfigOverride.randomPorts()` that can be used instead. I didn't touched files that contained additional settings, yet.

I also made sure that the `OpaConfig` always has a default `OpaClientConfiguration` set. This is necessary to allow to use the `DropwizardAppRule` without a config file (i.e. `new DropwizardAppRule<>(Application.class, null, randomPorts())`. This change will not affect services since there already was a null check in the code before: https://github.com/SDA-SE/sda-dropwizard-commons/blob/a08c00b6b0fabbe619b11e8573bc88aa9dbabd4a/sda-commons-server-auth/src/main/java/org/sdase/commons/server/opa/OpaBundle.java#L171-L172


Fox32 (approved):
☄️

PR #698, opened 2020-12-11
feat: include metrics for health checks at /metrics/prometheus

Fox32:
Previously these metrics were only available at the additional /healthcheck/prometheus endpoint that is now deprecated.


christopher-cudennec (approved):
Noice! 👍

PR #699, opened 2020-12-11
Upgrade db rider

JoergSiebahn:
Although it doesn't hurt us, https://github.com/database-rider/database-rider/issues/104 blocked us for a long time for upgrading, but is fixed now.

### DIFF
--- a/sda-commons-dependencies/build.gradle
+++ b/sda-commons-dependencies/build.gradle
@@ -67,12 +67,7 @@ dependencies {
     api "org.flywaydb:flyway-core:6.5.5"
 
     // sda-commons-server-hibernate-testing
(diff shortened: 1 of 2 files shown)


build.gradle
```diff
@@ -21,6 +21,7 @@ allprojects {
 // we need a repository for the combination of jacoco and sonarcloud
 repositories {
   mavenCentral()
+  maven { url "https://oss.sonatype.org/content/repositories/snapshots" }
```
dhenneke, line 24: Do we still need this?

JoergSiebahn: No. I already removed it with the last force push but maybe missed something. Should be fixed now.

Fox32 (approved):
☄️

Fox32 (approved):
🥁

PR #536, opened 2020-09-21
Add an experimental OpenAPI 3.0 module

dhenneke:
This PR adds an `OpenApiBundle` module that can be used as an alternative to the `SwaggerBundle` and provides OpenAPI 3.0 specifications at the `/openapi.{yaml|json}` resource.

Some features:
* Generate OpenAPI 3.0 documentation with the new swagger-v3 annotations
* Use an existing OpenAPI 3.0 file as basis (for API-first approaches)
* HAL documentation support

> This bundle is not yet part of the `SdaPlatformBundle` and should be considered as an `alpha` module! We will decide in the future if and how we want to transition from the SwaggerBundle to the OpenApiBundle.

Some problems/changes:
* In contrast to the `SwaggerBundle`, I decided that it should not be possible to configure things like title or version in the `builder()` anymore. Instead the services should configure them in a `@OpenApiDefinition(info = @Info(…))` annotation.
* It is not possible to have `$ref` properties alongside `description`. This will lead to broken documentations for things like `HALLink` or other classes that are reused. In fact, it will use the description of the first property as description for the model and will discard all following. This will change in [OpenAPI 3.1.0](https://github.com/OAI/OpenAPI-Specification/releases/tag/3.1.0-rc0) , which isn't released yet and is also lacking a `swagger-core` implementation (though there is already a [3.1](https://github.com/swagger-api/swagger-core/tree/3.1) branch).

PR #729, opened 2021-01-04
fix(server-openapi): upgrade swagger-hal to 2.0.0 and remove the custom HAL-logic for openapi

dhenneke:
This PR includes two changes:
1. Update swagger-core (ref #686) and swagger-parser (ref #688) as they should be bumped together.
2. Update to the latest release of the `swagger-hal` library (ref #722).

### Swagger HAL for OpenAPI

#536 added a custom extension to render HAL features in OpenAPI since the formerly used `swagger-hal` library didn't support the v3 version of the swagger library. Now, the library got updated and is fully compatible with swagger-v3. Unfortunately, this was not done in a new library (as in swagger-v3) but in a `2.0.0` major release, which adds complexity since `sda-commons-server-swagger` still depends on version `1.0.4`.

I did the following steps:
1. Move the dependency `swagger-hal:1.0.4` from `sda-commons-dependencies` to the `sda-commons-server-swagger` module.
2. Add the dependency `swagger-hal:2.0.0` to the `sda-commons-server-openapi` module.
3. Modify the `sda-commons-dependency-check` to ignore the `swagger-hal` dependency from the swagger module.
4. Remove our custom model resolver and all tests that were copied from `swagger-hal`.

This introduces some additional constraints:
* The `sda-commons-server-swagger` module can't be used in the same project as the `sda-common-server-openapi` module. This leads to (additional) incompatibilities between the `sda-commons-server-starter` and the `sda-commons-starter` module.
* We have different versions of `swagger-hal` in sda-dropwizard-commons but only check for dependency conflicts with version `2.0.0` and ignore version `1.0.4`.


JoergSiebahn (approved):
Do we have new Sonar rules and quality gates?

A single added deprecation is a good pattern in library development when using SemVer (especially when added to the release notes 👍 ) in my opinion.

PR #730, opened 2021-01-04
feat: add AbstractOpa

ericrath:


sda-commons-server-auth-testing/src/main/java/org/sdase/commons/server/opa/testing/AbstractOpa.java
```diff
@@ -0,0 +1,345 @@
...
+    }
+
+    MappingBuilder matchInput(String httpMethod, String[] paths) {
+      return matchAnyPostUrl()
+          .withRequestBody(matchingJsonPath("$.input.httpMethod", equalTo(httpMethod)))
+          .withRequestBody(matchingJsonPath("$.input.path", equalToJson(getJson(paths))));
+    }
+
+    static String[] splitPath(String path) {
+      if (path.startsWith("/")) {
```
RappC, line 339: is this also working with Windows as it uses backslashes to separate paths? 
I know that this is just copy&paste from the previous code, but I'm still curious.

RappC (approved):

sda-commons-server-auth-testing/src/main/java/org/sdase/commons/server/opa/testing/AbstractOpa.java
ericrath: as discussed- it's just the http path

PR #740, opened 2021-01-11
feat(server-testing): add jupiter support for environmentRule 

ericrath:


sda-commons-server-testing/src/main/java/org/sdase/commons/server/testing/DropwizardRuleHelper.java
```diff
@@ -90,4 +91,9 @@ private DropwizardRuleHelper(Class<A> appClass) {
   public DropwizardAppRule<C> build() {
     return new DropwizardAppRule<>(appClass, configurationBuilder.build());
   }
+
+  @Override
+  public DropwizardAppExtension<C> buildExtension() {
+    return new DropwizardAppExtension<>(appClass, configurationBuilder.build());
+  }
```
Fox32, line 98: I don't think that we should add new features to this deprecated class.

sda-commons-server-testing/src/main/java/org/sdase/commons/server/testing/junit5/EnvironmentExtension.java
```diff
@@ -0,0 +1,75 @@
...
+ * <pre>
+ *   class MyTest {
+ *
+ *     &#64;RegisterExtension
+ *     public final env = new EnvironmentExtension().setEnv("DISABLE_JWT", "true");
+ *
+ *   }
+ * </pre>
+ */
+public class EnvironmentExtension implements BeforeAllCallback, AfterAllCallback {
```
Fox32, line 26: Did you checked out if it's an option to use [system-rules](https://github.com/stefanbirkner/system-rules/issues/55), [system-lambda](https://github.com/stefanbirkner/system-lambda) or some kind of fork of it as @JoergSiebahn suggested?

sda-commons-server-testing/src/main/java/org/sdase/commons/server/testing/junit5/DropwizardAppExtension.java
```diff
@@ -0,0 +1,127 @@
...
+import javax.annotation.Nullable;
+import org.junit.jupiter.api.extension.AfterAllCallback;
+import org.junit.jupiter.api.extension.BeforeAllCallback;
+import org.junit.jupiter.api.extension.ExtensionContext;
+
+/**
+ * This class only adds {@link BeforeAllCallback} and {@link AfterAllCallback} to the original
+ * {@link io.dropwizard.testing.junit5.DropwizardAppExtension}.
+ */
+public class DropwizardAppExtension<C extends Configuration>
```
Fox32, line 20: Can we add a comment linking to https://github.com/dropwizard/dropwizard/pull/3649 here and rename this to avoid the name clash with Dropwizard? That way to can run into trouble if you are using the wrong import?

Fox32: Sonar has the same opinion here 😉

sda-commons-server-testing/src/main/java/org/sdase/commons/server/testing/junit5/EnvironmentExtension.java
christopher-cudennec: There is one project that we would like to use for the replacement of the RetryRule anyway: https://junit-pioneer.org/docs/
Unfortunately it does not offer the feature to set the value using  a supplier function (what we need to lazily set the values).

There is not a framework that I know of that offers exactly what we want here.

sda-commons-server-testing/src/main/java/org/sdase/commons/server/testing/DropwizardRuleHelper.java
ericrath: Since we can't resolve the deprecation of the DwRuleHelper, because we want to keep the JUnit4 support, would it make sense to declare a new class over modifying this existing but deprecated?
If we have a separate class for Jupiter, how can the decision be determined?

Fox32: Nobody should use the DropwizardRuleHelper anymore, it's deprecated. We shouldn't extend a deprecated feature with new stuff?

ericrath: we need to shift this change to another pr, since we dont require it for the environment rule right now

sda-commons-shared-error/build.gradle
```diff
@@ -4,4 +4,5 @@ dependencies {
 
   testCompile 'junit:junit'
   testCompile 'org.assertj:assertj-core'
+  testRuntimeOnly 'org.junit.vintage:junit-vintage-engine'
```
Fox32, line 7: Why does shared-error need this but no other module?

sda-commons-server-testing/build.gradle
```diff
@@ -11,7 +11,13 @@ dependencies {
   compile 'jakarta.servlet:jakarta.servlet-api'
 
   compile 'junit:junit'
+  compile 'org.junit.jupiter:junit-jupiter'
+  compile 'org.junit.jupiter:junit-jupiter-api'
+  // compile 'org.junit.jupiter:junit-platform-commons'
```
Fox32, line 16: Can we remove this comment? Seems to be unused code

Fox32 (commented):
Not sure about the dependency part, but otherwise it already looks promising

sda-commons-server-testing/build.gradle
```diff
@@ -11,7 +11,12 @@ dependencies {
   compile 'jakarta.servlet:jakarta.servlet-api'
 
   compile 'junit:junit'
+  compile 'org.junit.jupiter:junit-jupiter'
+  compile 'org.junit.jupiter:junit-jupiter-api'
```
christopher-cudennec, line 15: Seems to be enough to only add "junit-jupiter-api". Can you remove the other dependency?

sda-commons-server-testing/build.gradle
```diff
@@ -11,7 +11,11 @@ dependencies {
   compile 'jakarta.servlet:jakarta.servlet-api'
 
   compile 'junit:junit'
+  compile 'org.junit.jupiter:junit-jupiter-api'
   compile 'org.hamcrest:hamcrest'
   compile 'org.mockito:mockito-core'
   compile 'org.assertj:assertj-core'
+
+  runtimeOnly 'org.junit.jupiter:junit-jupiter-engine'
+  runtimeOnly 'org.junit.vintage:junit-vintage-engine'
```
JoergSiebahn, line 20: Do we need vintage in runtime scope or should we leave it to the consumers if they want to support vintage in their tests? I guess, finally projects should decide if they use Junit4 or Junit5 but should not mix. Or is it better to provide everything that may be needed?

ericrath: we need it here. without vintage, the junit4 tests are not executed.
vice versa: without jupiter junit5 tests are not executed

sda-commons-server-testing/src/main/java/org/sdase/commons/server/testing/junit5/EnvironmentExtension.java
JoergSiebahn: I've [seen use of EnvironmentRule](https://github.com/search?p=1&q=org%3ASDA-SE+EnvironmentRule&type=Code) that needs lazy values in sda commons and some archived services like [otp-service|https://github.com/SDA-SE/otp-service/search?q=TWILIO_URL] and some services where it can easily replaced with Dropwizard config overrides. All other occurrences in [SDA SE|https://github.com/SDA-SE] should be replaceable with the [annotations](https://junit-pioneer.org/docs/environment-variables/) of the mentioned lib as they use static values.

sda-commons-server-testing/build.gradle
JoergSiebahn: So for vintage `testCompile` should be fine, but we have to decide if we wan't to provide it for consumers.
@Fox32, @christopher-cudennec what do you think?

christopher-cudennec: I think it's common practice to not include the engine itself as compile scope dependency but to define it as `runtimeOnly`. All the projects that want to use need to add the engine themselves AFAIK.

JoergSiebahn: Ah, Ok. runtime scope is quite tricky.
`runtimeOnly` or the Maven result `runtime` [defines dependencies for execution](http://andresalmiray.com/maven-scopes-vs-gradle-configurations/#:~:text=Maven%20defines%206%20scopes%3A%20compile,system%2C%20test%2C%20and%20import.&text=runtime%20This%20scope%20indicates%20that,but%20not%20the%20compile%20classpath.). That lead to my idea that it affects the consumers. But it also means that [the dependency is not inherited](https://stackoverflow.com/a/12273576).

Should we add a general part in the README about what is needed to use Junit5?

christopher-cudennec: We've just double checked one consumer: It's not necessary to add the engine dependency because you already get it from "sda-commons-server-testing". You only need to activate Junit platform for the test plugin in the build.gradle.
We'll add some documentation to the README 👍

sda-commons-server-testing/src/main/java/org/sdase/commons/server/testing/junit5/EnvironmentExtension.java
christopher-cudennec: otp-service is archived but I think we need the lazy feature anyways.
Would be an option to contribute that behaviour to "junit-pioneer" for example.
But that's nothing I would do rightaway in our current work to add JUnit 5 functionality.

sda-commons-server-testing/build.gradle
christopher-cudennec: > So for vintage testCompile should be fine, but we have to decide if we wan't to provide it for consumers.
@Fox32, @christopher-cudennec what do you think?

I think the current setup is fine. Although the engines will be passed down to consumers with "test runtime" scope, tests will still be executed the known way with JUnit 4. JUnit 5 execution will only be activated by explicitly configuring the "test" plugin of your project.

sda-commons-server-testing/src/main/java/org/sdase/commons/server/testing/junit5/EnvironmentExtension.java
JoergSiebahn: If we really need the lazy feature, the extension is fine for me.

According the system lambdas: As there is no explicit Junit5 extension available yet, we need to implement our own wrapper anyways. I do not see much benefit in wrapping another dependency over reusing (a copy of) our existing code.

christopher-cudennec:
Long story short:
After discussion with @JoergSiebahn and the team we decided that we do not want to provide a JUnit 5 extension as a replacement for the `EnvironmentRule` because all the basic functionality can be achieved using other frameworks like junit-pioneer. The only additional functionality (setting the value with a supplier function) is not actively used by any of our repos.

Therefore we introduce JUnit 5 with another PR: #741

PR #742, opened 2021-01-11
feat(server-auth): reloading key-cache periodically to avoid security-leaks

Uwe-Fuchs:


dhenneke:
Please use the scope `server-auth` for the commit message: `feat(server-auth): ...` and briefly describe what has changed and why in the message.

sda-commons-server-auth/src/test/java/org/sdase/commons/server/auth/key/RsaKeyLoaderSchedulerTest.java
```diff
@@ -0,0 +1,70 @@
...
+    keyLoaderScheduler.start(0, period, timeUnit);
+    assertThat(keyLoader.getKey(kidForValidKey)).isSameAs(validKey);
+    assertThat(keyLoader.getKey(kidForKeyToBeRevoked)).isSameAs(keyToBeRevoked);
+
+    // when
+    // simulate revoking of key:
+    includingKeyToBeRevoked = false;
+
+    // wait for scheduler-execution:
+    timeUnit.sleep(period);
```
Fox32, line 58: Please use the awaitability package to use wait for the condition below to succeed. This should avoid flaky tests. Don't configure timeouts when using it, the defaults are fine. Like [here ](https://github.com/SDA-SE/sda-dropwizard-commons/blob/1b60c8ba20f00601e35ca91be2725b48d93ed135/sda-commons-server-opentracing/src/test/java/org/sdase/commons/server/opentracing/OpenTracingBundleTest.java#L65-L70).

It doesn't feel that it's guaranteed that the key load is completed after waiting for two seconds as the scheduler is also waiting two seconds, right?

Uwe-Fuchs: That's a cool tool. I didn't know it yet, thanks for the tip! Solves a lot of problems around testing asynchronous code.

Yes, two seconds maybe too short. I think the `await().untilAsserted` - statement is much more elegant for solving this.

christopher-cudennec:
Nice feature. Have you already tested the SNAPSHOT in one of your projects?

christopher-cudennec:
Can you please squash your commits and rephrase the message to match semantic versioning?
More information: https://sda-se.atlassian.net/wiki/spaces/DG/pages/863010951/Semantic+Commits

The message should start with: `feat(server-auth): ...`

sda-commons-server-auth/README.md
```diff
@@ -172,6 +172,18 @@ In addition, the client consumes the standard [proxy system properties](https://
 Please note that a specific proxy configuration in the `HttpClientConfiguration` disables the proxy system properties for the client using that configuration.
 This can be helpful when all clients in an Application should use the same proxy configuration (this includes all clients that are created by the [`sda-commons-client-jersey` bundle](../sda-commons-client-jersey).
 
+
+## Periodical reloading of public keys provided via JWKS
+
+Public keys provided via a [`JSON Web Key Set`](https://tools.ietf.org/html/draft-ietf-jose-json-web-key-41#section-5) (JWKS) are stored in a local key-cache.
+This is a potential security leak, because keys can be revoked by their issuer. The following scenario is possible:
```
dhenneke, line 179: Can we rephrase this text? A bit shorter and more on point. Focus on the behavior, not on the potential Problem. It could also tell (again) on how new keys are fetched. Something like:
* the first sentence with the key-cache
* for unknown keys, the cache is refreshed to search for the kid
* the key cache is refreshed periodically (each 5 min) to remove potentially revoked (I.e. removed from the key set) keys.
* If the key source is unavailable, the old keys stay valid (correct?) and no new kid's are loaded

dhenneke (commented):
I'm fine with activating this for everyone, but please check that it doesn't disturb availability on edge-case scenarios before.

sda-commons-server-auth/README.md
JoergSiebahn: Actually I like to read about the problem that is solved to gain awareness about this security issue. But @dhenneke is right. The first part could be a bit shorter.

sda-commons-server-auth/src/main/java/org/sdase/commons/server/auth/AuthBundle.java
```diff
@@ -130,6 +139,10 @@ private KeySource createKeySources(KeyLocation keyLocation, Client client) {
     }
   }
 
+  private void setupManagedThreadManager(Environment environment) {
+    environment.lifecycle().manage(onShutdown(() -> keyLoaderScheduler.shutDown()));
```
JoergSiebahn, line 143: As you get the executor service from the lifecycle, it is already managed and will be shut down by the framework.
Although it doesn't harm the shutdown, this call of `shutDown()` is not necessary and should be removed.

Uwe-Fuchs: Done! My look on the `ManagedShutdownListener` definitely was too superficial. Now I think I got how it works (and removed it from the AuthBundle).

sda-commons-server-auth/src/test/java/org/sdase/commons/server/auth/key/RsaKeyLoaderSchedulerTest.java
```diff
@@ -0,0 +1,72 @@
...
+    KeySource keySource =
+        new KeySource() {
+          @Override
+          public List<LoadedPublicKey> loadKeysFromSource() throws KeyLoadFailedException {
+            if (includingKeyToBeRevoked) {
+              return Arrays.asList(
+                  new LoadedPublicKey(kidForValidKey, validKey, this),
+                  new LoadedPublicKey(kidForKeyToBeRevoked, keyToBeRevoked, this));
+            } else {
+              return singletonList(new LoadedPublicKey("kidForValidKey", validKey, this));
```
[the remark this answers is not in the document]
Uwe-Fuchs: Done! I think it's definitely a possible scenario, so it should be tested as well.

sda-commons-server-auth/README.md
```diff
@@ -172,6 +172,18 @@ In addition, the client consumes the standard [proxy system properties](https://
...
+
+## Periodical reloading of public keys provided via JWKS
+
+Public keys provided via a [`JSON Web Key Set`](https://tools.ietf.org/html/draft-ietf-jose-json-web-key-41#section-5) (JWKS) are stored in a local key-cache.
+This is a potential security leak, because keys can be revoked by their issuer. The following scenario is possible: 
+We have loaded and cached the public keys. One of these keys is revoked by the issuer in the meantime. Now a client
+gives us a JWT whose ID references exactly the formerly revoked key. Because we didn't notice the revocation, we finally accept the request.
+
+To minimize the risk of getting into this scenario, the key-cache is reloaded periodically. The period is set to a fix value (currently 5 min).
+Maybe the period will be made configurable in a future-release, e.g. via the [Dropwizard configuration](https://www.dropwizard.io/en/latest/manual/configuration.html).
```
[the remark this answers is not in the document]
Uwe-Fuchs: "maybe"-part removed!

sda-commons-server-auth/README.md
Uwe-Fuchs: I've shortened the description and put it right after the `Configuration`-part. Hopefully it's more concise now. Otherwise let me know!

sda-commons-server-auth/src/main/java/org/sdase/commons/server/auth/key/RsaKeyLoaderScheduler.java
```diff
@@ -0,0 +1,51 @@
...
+  }
+
+  public static RsaKeyLoaderScheduler create(
+      RsaPublicKeyLoader keyLoader, ScheduledExecutorService reloadKeysExecutorService) {
+    Validate.notNull(keyLoader, "keyLoader should not be null");
+    Validate.notNull(reloadKeysExecutorService, "executorService should not be null");
+    return new RsaKeyLoaderScheduler(keyLoader, reloadKeysExecutorService);
+  }
+
+  public RsaKeyLoaderScheduler start(int initialDelay, int period, TimeUnit timeUnit) {
```
JoergSiebahn, line 31: Must this method check for something like isStarted and silently ignore it or stop and start again?
Or simpler: Can we make it package private and document „only call once“ so that nobody from outside can use it wrong?

sda-commons-server-auth/src/test/java/org/sdase/commons/server/auth/key/RsaKeyLoaderSchedulerTest.java
```diff
@@ -0,0 +1,103 @@
...
+        .untilAsserted(
+            () -> {
+              assertThat(keyLoader.getKey(kidForValidKey)).isSameAs(validKey);
+              // revoked key not found anymore
+              assertThat(keyLoader.getKey(kidForKeyToBeRevoked)).isNull();
+            });
+  }
+
+  @Test
+  public void shouldContinueJobInCaseOfError() {
```
JoergSiebahn, line 71: Is the name correct? I see an assertion for a successful run followed by the assertion to still have valid data after errors.
The other way around would be helpful as well to verify that the scheduler continues after error. Usually the executor stops it's schedule after an exception is thrown in execution.

sda-commons-server-auth/src/test/java/org/sdase/commons/server/auth/key/RsaKeyLoaderSchedulerTest.java
```diff
@@ -0,0 +1,103 @@
...
+
+    keyLoader.addKeySource(keySource);
+    keyLoaderScheduler.start(0, 2, TimeUnit.SECONDS);
+    await().untilAsserted(() -> assertThat(keyLoader.getKey(kidForValidKey)).isSameAs(validKey));
+
+    // when -> simulate error-occurrence:
+    elseCondition = true;
+
+    // then -> key is found anyway:
+    await().untilAsserted(() -> assertThat(keyLoader.getKey(kidForValidKey)).isSameAs(validKey));
```
dhenneke, line 96: Hmm. This is a tricky one. I am not sure if this really tests what it intends to do. We can't know for sure that the keysource was actually reloaded (it probably didn't).

I had a deeper look into the code (didn't do that on the phone in the morning) and figured that the `reloadKeys()` functionality is tested here: https://github.com/SDA-SE/sda-dropwizard-commons/blob/55762502ee95073157656a191b261ed69659d5f7/sda-commons-server-auth/src/test/java/org/sdase/commons/server/auth/key/RsaPublicKeyLoaderTest.java#L53-L83

Maybe this test case isn't as important as I thought. So you could go with one of the following options:
1. Make sure the test works as expected.
2. Remove the test (sorry for the inconvenience...).

At least a test to see that an exception did not break anything would be good.

sda-commons-server-auth/src/test/java/org/sdase/commons/server/auth/key/RsaKeyLoaderSchedulerTest.java
```diff
@@ -0,0 +1,103 @@
...
+            }
+          }
+        };
+
+    keyLoader.addKeySource(keySource);
+    keyLoaderScheduler.start(0, 2, TimeUnit.SECONDS);
+    await().untilAsserted(() -> assertThat(keyLoader.getKey(kidForValidKey)).isSameAs(validKey));
+
+    // when -> simulate error-occurrence:
+    elseCondition = true;
```
JoergSiebahn, line 93: I think, we have to verify that key loading is scheduled a second time after changing the behavior in this case.

sda-commons-server-auth/src/test/java/org/sdase/commons/server/auth/key/RsaKeyLoaderSchedulerTest.java
JoergSiebahn: As @dhenneke reported below, we may only need the other way round and assert that the scheduler continues after error.

sda-commons-server-auth/src/main/java/org/sdase/commons/server/auth/key/RsaKeyLoaderScheduler.java
Uwe-Fuchs: Done! I thought about setting it package-private as well!

sda-commons-server-auth/README.md
```diff
@@ -147,6 +147,14 @@ In this case, the `AUTH_KEYS` variable should contain a JSON array of
...
 ```
 
+### Periodical reloading of public keys
+
+The public keys are cached locally. For keys with an unknown key-id (kid), 
+the cache is refreshed, and the key is searched again afterwards.<br/>
+In addition, the key-cache is refreshed periodically (currently each 5 min) to remove keys
+which are revoked by their issuers. <br/>
+If the refresh fails (e.g. key-source unavailable, network-problems etc.), the old keys in the cache stay valid.
+
```
JoergSiebahn, line 157: NIT: You don't need `<br/>` in Markdown, two spaces at the end of the line are enough to produce a line break. But in this case, a single paragraph without line breaks should be fine as well.

```suggestion
The public keys are cached locally. For keys with an unknown key-id (kid), 
the cache is refreshed, and the key is searched again afterwards.
In addition, the key-cache is refreshed periodically (currently each 5 min) to remove keys
which are revoked by their issuers.
If the refresh fails (e.g. key-source unavailable, network-problems etc.), the old keys in the cache stay valid.

```

sda-commons-server-auth/src/test/java/org/sdase/commons/server/auth/key/RsaKeyLoaderSchedulerTest.java
```diff
@@ -0,0 +1,110 @@
...
+
+    // then:
+    await()
+        // wait 1 second longer then reload-period to make sure the reload-job is executed before
+        // checking:
+        .pollDelay(RELOAD_PERIOD + 1, TimeUnit.SECONDS)
```
JoergSiebahn, line 101: I think it is better to count the invocations and wait until another invocation happened rather than adding another second to the test runtime.

sda-commons-server-auth/README.md
Uwe-Fuchs: Done, thanks for the hint!

sda-commons-server-auth/src/test/java/org/sdase/commons/server/auth/key/RsaKeyLoaderSchedulerTest.java
Uwe-Fuchs: I gave it a new try - hopefully it's more concise now.

sda-commons-server-auth/src/test/java/org/sdase/commons/server/auth/key/RsaKeyLoaderSchedulerTest.java
```diff
@@ -0,0 +1,107 @@
...
+    keyLoader.addKeySource(keySource);
+    keyLoaderScheduler.internalStart(0, RELOAD_PERIOD, TimeUnit.SECONDS);
+
+    await().untilAsserted(() -> assertThat(numberOfCalls.get()).isEqualTo(2));
+    assertThat(keyLoader.getKey(kidForValidKey)).isSameAs(validKey);
+
+    // simulate keyLoad-failure:
+    keyLoadShouldFail.set(true);
+    await().untilAsserted(() -> assertThat(numberOfCalls.get()).isEqualTo(3));
+    assertThat(anErrorHasOccurred.get()).isEqualTo(true);
```
JoergSiebahn, line 98: Well named condition 👍

sda-commons-server-auth/src/test/java/org/sdase/commons/server/auth/key/RsaKeyLoaderSchedulerTest.java
```diff
@@ -0,0 +1,128 @@
...
+    waitForNextReload();
+  }
+
+  @After
+  public void shutdownExecutorService() {
+    executorService.shutdown();
+  }
+
+  @Test
+  public void shouldNotBeAuthenticatedWithRevokedKey() {
```
dhenneke, line 58: Since we don't authenticate anything but just check if the keys are there, maybe:

```suggestion
  public void shouldRemoveRevokedKey() {
```

sda-commons-server-auth/src/test/java/org/sdase/commons/server/auth/key/RsaKeyLoaderSchedulerTest.java
```diff
@@ -0,0 +1,128 @@
...
+        }
+        return actualKeys;
+      };
+
+  @Before
+  public void startKeyLoader() {
+    keyLoader.addKeySource(keySource);
+    keyLoaderScheduler.internalStart(0, RELOAD_PERIOD, TimeUnit.SECONDS);
+    // wait for initial load
+    waitForNextReload();
```
dhenneke, line 49: The `private final …` variables are reused for the tests so the `keySource` is added multiple times. `actualInvocationFails` also doesn't get reset which might break the tests if JUnit decides to execute the tests in a different order.

JoergSiebahn: Because they are not static, a new instance is created for each test case. They are final for each instance and JUnit creates a new instance for each test case. There should be no conflict regarding execution order.

dhenneke: ok. nevermind 👍

dhenneke (approved):
Thanks!

PR #762, opened 2021-01-20
Feat/junit5 hibernate

christopher-cudennec:


sda-commons-server-testing/README.md
```diff
@@ -16,6 +16,11 @@ In case you want to use JUnit 5 you also have to activate it in your build.gradl
   test {
     useJUnitPlatform()
   }
+
+  // in case you use the integrationTest plugin:
+  integrationTest {
+    useJUnitPlatform()
```
ericrath, line 22: we might want to use inheritance here
let me check the syntax

ericrath (changes requested):

sda-commons-server-testing/README.md
ericrath: configurations {
    integrationTest.extendsFrom test   
}

ericrath (changes requested):

sda-commons-server-testing/README.md
RappC: @ericrath what is the advantage of your proposal over the suggestion made by @christopher-cudennec

ericrath: it makes sure, that our integration tests dont differ from the junit base.
Especially if we add functionality to test{}

christopher-cudennec: I think I would go with your suggestion if there was more configuration to the test plugin. 
For now it's the only thing that needs to be configured, so I would stick with the current version.

ericrath (approved):

JoergSiebahn (approved):

PR #792, opened 2021-02-01
fix: provide fixed path and name to mongod executable

eikenilsknopp:
- extract mongodb executable to fixed path and filename to avoid recurring firewall requests on mac
- move some tests to separate class to avoid clashes with conflicting mongodb executables

Since its not a change to the behaviour of the code, only runs on MacOS and can not be really covered by tests, can we accept the low code coverage here?

### DIFF
--- a/sda-commons-server-mongo-testing/src/test/java/org/sdase/commons/server/mongo/testing/MongoDbVersionTest.java
+++ b/sda-commons-server-mongo-testing/src/test/java/org/sdase/commons/server/mongo/testing/MongoDbVersionTest.java
@@ -1,0 +1,63 @@
+package org.sdase.commons.server.mongo.testing;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+import de.flapdoodle.embed.mongo.distribution.IFeatureAwareVersion;
+import de.flapdoodle.embed.mongo.distribution.Version;
+import java.util.concurrent.atomic.AtomicReference;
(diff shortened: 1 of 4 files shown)


sda-commons-server-mongo-testing/src/test/java/org/sdase/commons/server/mongo/testing/MongoDbVersionTest.java
```diff
@@ -0,0 +1,63 @@
...
+
+import de.flapdoodle.embed.mongo.distribution.IFeatureAwareVersion;
+import de.flapdoodle.embed.mongo.distribution.Version;
+import java.util.concurrent.atomic.AtomicReference;
+import org.apache.commons.lang3.SystemUtils;
+import org.junit.Test;
+import org.junit.runner.Description;
+import org.junit.runners.model.Statement;
+
+public class MongoDbVersionTest {
```
christopher-cudennec, line 13: Do we want new tests to be JUnit 5?

eikenilsknopp: They are not really "new" I just moved them to a new class. I would keep them in Junit4

christopher-cudennec:
Can you please squash your two commits into one? 🙂

JoergSiebahn (approved):

PR #796, opened 2021-02-02
docs(server-auth): OPA called even if request does not contain jwt

RappC:


sda-commons-server-auth/README.md
```diff
@@ -189,6 +189,13 @@ This can be helpful when all clients in an Application should use the same proxy
 Details about the authorization with Open Policy Agent are documented within the authorization concept (see Confluence). 
 In short, Open Policy Agent acts as policy decision point and is started as sidecar to the actual service.
 
+The OPA Bundle acts as a client to the Open Policy Agent and is hooked in as request filter handling requests after they have 
+been validated by the `JwtAuthenticator` and before they reach the service implementation.
+
+The OPA Bundle requires the AUT Bundle in place so that the JWT can be verified against public keys before it is handed over to OPA. 
+In case the request does not contain a JWT token at all the JWT verification in the Auth Bundle will be skipped without error and further 
+checks should be done using OPA.
+
```
dhenneke, line 198:
```suggestion
The OPA Bundle requires the [`AuthBundle`](./src/main/java/org/sdase/commons/server/auth/AuthBundle.java) in place so the JWT can be verified against public keys before it is handed over to OPA. 
In case the request does not contain a JWT token at all, the JWT verification in the `AuthBundle` will be skipped without error and further 
checks should be part of the OPA policy.
```

sda-commons-server-auth/README.md
```diff
@@ -189,6 +189,12 @@ This can be helpful when all clients in an Application should use the same proxy
 Details about the authorization with Open Policy Agent are documented within the authorization concept (see Confluence). 
 In short, Open Policy Agent acts as policy decision point and is started as sidecar to the actual service.
 
+The OPA Bundle acts as a client to the Open Policy Agent and is hooked in as request filter handling requests after they have 
+been validated by the `JwtAuthenticator` and before they reach the service implementation.
+
+The OPA Bundle requires the [`AuthBundle`](./src/main/java/org/sdase/commons/server/auth/AuthBundle.java) in place so the JWT can be verified against public keys before it is handed over to OPA. 
+In case the request does not contain a JWT token at all, the JWT verification in the `AuthBundle` will be skipped without error and further 
+checks should be part of the OPA policy.
```
Fox32, line 197: Should we mention that this behavior depends on the selected configuration? Because it is possible to disable this behavior.

RappC: If that is possible we should describe it here and also tell the audience how disabling it works. What needs to be set?

RappC: @Fox32 I have added what you have requested. Can you re-check the PR and possibly approve.

dhenneke (approved):
Sorry that this took so long. I'm happy with the changes 👍

PR #829, opened 2021-02-26
fix(server-openapi): Prevent NullPointerException when no schemas exist

christopher-cudennec:


RappC (approved):

Fox32:
I guess if you remove the unrelated changes in the test, coverage should be fine. I think there is a rule that coverage is ignored if less than three lines are changed.

JoergSiebahn (approved):

PR #559, opened 2020-10-05
fix(deps): update kafka_2.12 and kafka-clients from 2.2.1 to 2.6.0

dhenneke:

PR #847, opened 2021-03-09
feat(server-kafka): Update to Kafka 2.6.1

christopher-cudennec:


christopher-cudennec:
Similiar PR: #559

sda-commons-dependencies/build.gradle
```diff
@@ -15,8 +15,8 @@ ext {
   weldVersion = '3.1.5.Final'
   openTracingUtilVersion = '0.33.0'
   jsonUnitVersion = '2.24.0'
-  scalaVersion = '2.12.8' // Always prefer the version from kafka_2.12. Make sure that the major version is always aligned with kafka, s3mock, and mbknor-jackson-jsonschema.
-  kafkaVersion = '2.2.1' // Fix version as decided in https://sda-se.atlassian.net/l/c/ce8EFM2e. Be careful when upgrading this dependency.
+  scalaVersion = '2.12.11' // Always prefer the version from kafka_2.12. Make sure that the major version is always aligned with kafka, s3mock, and mbknor-jackson-jsonschema.
+  kafkaVersion = '2.6.1'
```
dhenneke, line 19: Why not use `2.7.0`? https://search.maven.org/artifact/org.apache.kafka/kafka_2.12/2.7.0/jar

christopher-cudennec: > https://github.com/salesforce/kafka-junit/releases/tag/v3.2.2

Our testing library says: "Validate against Kafka versions 2.5.x and 2.6.x"

christopher-cudennec:
Asking for feature "disable Zookeeper admin server" at Salesforce: https://github.com/salesforce/kafka-junit/issues/49

sda-commons-server-kafka/src/test/java/org/sdase/commons/server/kafka/KafkaTopicIT.java
```diff
@@ -117,28 +118,31 @@ public void checkTopicSuccessfulComplex() {
...
+                          .withValueDeserializer(new StringDeserializer())
+                          .withListenerStrategy(
+                              new AutocommitMLS<String, String>(
+                                  record -> results.add(record.value()),
+                                  new IgnoreAndProceedErrorHandler<>()))
+                          .build());
+
+              assertThat(stringStringMessageListener).isNotNull();
+            })
+        .isInstanceOf(MismatchedTopicConfigException.class);
```
Fox32, line 141: Is the assert statement in line 139 really executed? You are expecting the previous code to throw, so I'm a bit confused.

JoergSiebahn: I agree and expect that line 139 will not be executed. But actually line 137 in the test before did not either. I disagree with Sonar although it catches the same point here. If you follow Sonars advice completely you need to pick the exact method that should throw the exception but in my opinion we should ensure that building a message listener with this configuration throws an exception in total and not a single specific method. It is an integration test. That check was added in May last year and since then it's annoying me from time to time because tests often get less readable when you follow that rule.

christopher-cudennec: I'll remove line 139 as it is not needed. Didn't see that when converting the test from JUnit 4 to 5.

Fox32 (approved):
👍

sda-commons-server-kafka-example/src/test/java/org/sdase/commons/server/kafka/KafkaExampleConsumerIT.java
```diff
@@ -30,24 +31,20 @@
           // fresh kafka instance
           .withBrokerProperty("group.initial.rebalance.delay.ms", "0");
 
-  private static final KafkaBrokerEnvironmentRule KAFKA_BROKER_ENVIRONMENT_RULE =
```
dhenneke, line 33: 👍 I thought that I removed every use of this Rule 🤔

PR #858, opened 2021-03-16
feat(client-jersey): add oidc client

eikenilsknopp:


sda-commons-client-oidc/src/main/java/org/sdase/commons/client/oidc/OidcClient.java
```diff
@@ -0,0 +1,103 @@
...
+import java.nio.charset.StandardCharsets;
+import java.util.Base64;
+
+import static org.sdase.commons.client.jersey.proxy.ApiClientInvocationHandler.createProxy;
+
+/**
+ * Service that can retrieve a new access token. Inspired by <a
+ * href="https://github.com/SDA-SE/bike-claim-report-ui/blob/develop/interaction-service/src/main/java/com/sdase/claim/bike/interaction/oidc/OidcAuthProvider.java">Bike
+ * Claim Report UI</a>
+ */
```
dhenneke, line 24: No need to link to the source

christopher-cudennec:
Some general comments:
* can you link the repo(s) where you've tested these changes?
* your commit message should also contain the context/module, e.g. `feat(server-kafka): This and that`

sda-commons-client-oidc/README.md
```diff
@@ -0,0 +1,95 @@
...
+
+  private OidcClient oidcClient;
+
+  public OidcRequestFilter(ClientFactory clientFactory, OidcConfiguration oidc) {
+      this.oidcClient = new OidcClient(clientFactory, oidc);
+  }
+
+  @Override
+  public void filter(ClientRequestContext requestContext) {
+    requestContext.getHeaders().add(AUTHORIZATION, "Bearer " + oidcResult.getAccessToken());
```
christopher-cudennec, line 37: Don't we have a constant for that "Bearer" string?

sda-commons-client-oidc/README.md
```diff
@@ -0,0 +1,95 @@
...
+  issuerUrl: ${OIDC_ISSUER_URL}
+```
+
+- _OIDC_DISABLED_
+    * Disable retrieving a new token completely. Not meant for production! 
+    * Example: `false`
+
+- _OIDC_GRANT_TYPE_
+    * Sets the OIDC grant type.
+    * Example: `client_credentials`
```
christopher-cudennec, line 75: We could list all supported values here.

sda-commons-client-oidc/README.md
```diff
@@ -0,0 +1,95 @@
...
+    * A username. Only used for grant type 'password'.
+    * Example: `john`
+
+- _OIDC_PASSWORD_
+    * A password. Only used for grant type 'password'.
+    * Example: `pa$$word`
+
+- _OIDC_ISSUER_URL_
+    * Contains the URL to the OpenID provider configuration document.
+    * Example: `https://<keycloak.domain>/auth/realms/<realm>`
```
christopher-cudennec, line 94: What's the right wording here? Isn't that the OIDC discovery URL?

sda-commons-client-oidc/src/main/java/org/sdase/commons/client/oidc/OidcClient.java
```diff
@@ -0,0 +1,132 @@
+package org.sdase.commons.client.oidc;
+
+import static org.sdase.commons.client.jersey.proxy.ApiClientInvocationHandler.createProxy;
+
+import com.github.benmanes.caffeine.cache.Cache;
```
christopher-cudennec, line 5: I think we should add Caffeine as compile dependency explicitly.

sda-commons-client-oidc/src/main/java/org/sdase/commons/client/oidc/OidcClient.java
```diff
@@ -0,0 +1,132 @@
...
+   */
+  public OidcResult createAccessToken() {
+    if (config.isDisabled()) {
+      LOGGER.debug("OIDC client is disabled");
+      return new OidcResult().setState(OidcState.SKIPPED);
+    }
+
+    TokenResource tokenResource;
+
+    if (cache == null) {
```
christopher-cudennec, line 68: I think the creation of the cache must be synchronized.

sda-commons-client-oidc/src/main/java/org/sdase/commons/client/oidc/OidcClient.java
```diff
@@ -0,0 +1,132 @@
...
+      tokenResource = getTokenResource();
+      if (tokenResource != null) {
+        cache =
+            Caffeine.newBuilder()
+                .expireAfterWrite(tokenResource.getAccessTokenExpiresInSeconds(), TimeUnit.SECONDS)
```
christopher-cudennec, line 73: Can we be sure that all tokens use the same expiration time?
Can we tell the cache to expire certain values instead of a global configuration?

sda-commons-client-oidc/src/main/java/org/sdase/commons/client/oidc/OidcClient.java
```diff
@@ -0,0 +1,132 @@
...
+    }
+    if (tokenResource != null) {
+      return new OidcResult().setState(OidcState.OK).setAccessToken(tokenResource.getAccessToken());
+    } else {
+
+      return new OidcResult().setState(OidcState.ERROR);
+    }
+  }
+
+  private TokenResource getTokenResource() {
```
christopher-cudennec, line 89: It would be even nicer to separate the actual business logic, i.e. token retrieval and the cache functionality.

sda-commons-client-oidc/src/main/java/org/sdase/commons/client/oidc/rest/model/TokenResource.java
```diff
@@ -0,0 +1,45 @@
...
+ * href="https://tools.ietf.org/html/rfc6749#section-5.1">RFC 6749 section 5.1</a>
+ */
+public class TokenResource {
+  @JsonProperty("token_type")
+  private String tokenType;
+
+  @JsonProperty("access_token")
+  private String accessToken;
+
+  @JsonProperty("expires_in")
```
christopher-cudennec, line 16: I thought this value is "exp"? Can that be changed? 
https://tools.ietf.org/html/rfc7519#section-4.1.4

sda-commons-client-oidc/src/test/resources/fixtures/tokenResponse.json
```diff
@@ -0,0 +1,5 @@
+{
+  "access_token": "eyJhbGciOiJSUzI1NiIsInR5cCIgOiAiSldUIiwia2lkIiA6ICI1VFVDOVlZOXdXRk0xNzZqT2RvcWRYUWtUMmxNbklXYTh6MU5qVlVrNm5ZIn0.eyJleHAiOjE2MTI0MjkwMzksImlhdCI6MTYxMjQyODk3OSwianRpIjoiNWNkZmQ4YTAtYWUyYS00OGQxLWE3YmItZDgwY2I5NjRhODAxIiwiaXNzIjoiaHR0cHM6Ly9rZXljbG9hay5zZGFkZXYuc2RhLXNlLmlvL2F1dGgvcmVhbG1zL21hbHdhcmUtc2Nhbm5lciIsImF1ZCI6Im1hbHdhcmUtc2Nhbm5lciIsInN1YiI6IjJiZjY2MzQzLWUyMDQtNDc1NS1iZTFkLWU3OWVjZTAyZTQ4NCIsInR5cCI6IkJlYXJlciIsImF6cCI6Im1hbHdhcmUtc2Nhbm5lciIsInNlc3Npb25fc3RhdGUiOiJkMGZjYjgxNi02MjU3LTQ4NzEtYTg1MC1hYjI2MjFhNjg1ZWYiLCJhY3IiOiIxIiwiYWxsb3dlZC1vcmlnaW5zIjpbIioiXSwic2NvcGUiOiJlbWFpbCBwcm9maWxlIiwiZW1haWxfdmVyaWZpZWQiOnRydWUsInByZWZlcnJlZF91c2VybmFtZSI6Im1hbHdhcmUtc2Nhbm5lci1lbmQyZW5kIn0.yMVQCgZkDA7-glUaJVNrAt65eei3r7_OOVX2DCXcA33dVw7XXy5a3WkY-IsbJiaH1euFU8-Vw_a5JCP2QB94AJxWpa5fSMKytspXQKeKf6FtiOHRtvRfpNMvwv2zuEzITJg-IqFg3yo4Bpfhmepv1lvxdPu6ip1u_OznRX_-uR829mfq-J7SW81RSTaxDmDVweoIDSi24tdDkJqjOET3PziVWhhY8AqJUmUFf41mHhM2P5w-52KZldSbvuPN32Z5pb-MJv_GonNei2B-8IbWdHml_v7HVlB4JYCOCyABWIYrorKYnxDeFc8j4NungzMQJ9_uwRcN1jatVqKBjt0l1w",
+  "expires_in": 60,
```
christopher-cudennec, line 3: Ah, I see. It's from the OAuth spec:
https://tools.ietf.org/html/rfc6749#page-73

sda-commons-client-oidc/README.md
```diff
@@ -0,0 +1,95 @@
...
+- _OIDC_CLIENT_SECRET_
+    * The client secret.
+    * Example: `s3cr3t`
+     
+- _OIDC_USERNAME_
+    * A username. Only used for grant type 'password'.
+    * Example: `john`
+
+- _OIDC_PASSWORD_
+    * A password. Only used for grant type 'password'.
```
Fox32, line 89: I wonder what @wurstbrot thinks about support the `password` grant type

sda-commons-client-oidc/src/main/java/org/sdase/commons/client/oidc/OidcClient.java
JoergSiebahn: I suggested yesterday via Slack to use a [custom eviction policy](https://www.baeldung.com/java-caching-caffeine#2-time-based-eviction) to evict each token after the actual expiration time (maybe a few seconds earlier). If thats implemented, the cache could be initialized in advance.

sda-commons-client-oidc/src/main/java/org/sdase/commons/client/oidc/OidcClient.java
JoergSiebahn: See [my comment](https://github.com/SDA-SE/sda-dropwizard-commons/pull/858#discussion_r597469291) above.

sda-commons-client-oidc/src/main/java/org/sdase/commons/client/oidc/OidcClient.java
JoergSiebahn: But according the synchronization: Fetching new access tokens could be synchronized as well. As the client is always getting a token for the same user, a second invocation can wait until a concurrent invocation is complete and then use the cache.

README.md
```diff
@@ -259,6 +259,11 @@ sda-commons library versions.
 The module [`sda-commons-client-jersey-example`](./sda-commons-client-jersey-example/README.md)
 presents an example application that shows how to invoke services.
 
+#### OIDC
+
+The module [`sda-commons-client-oidc`](./sda-commons-client-oidc/README.md) provides support for requesting OIDC access tokens
+from respective provides (e.g. Keycloak) for subsequent requests to other services.
+
```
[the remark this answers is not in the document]
JoergSiebahn: As the dependencies seem to be pretty similar it could be part of the client-jersey module in my opinion. That gives an opportunity for easier integration later. For example one could create an OIDC Client directly from the bundle or we can add the filter that is mentioned in the readme within the builder of the regular clients.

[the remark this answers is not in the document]
dhenneke: This was the plan in my original PR. But back then it wasn’t clear whether the service account/technical user approach was the universal Solution and should be used everywhere.

sda-commons-client-oidc/src/main/java/org/sdase/commons/client/oidc/rest/model/TokenResource.java
dhenneke: This is not a token but the response that _delivers_ the token. It’s more convenient to use this expiration value in seconds than to use with the Unix-timestamp in the token where you need to make sure you don’t confuse the different time zones.

christopher-cudennec: I see - thanks for the explanation!

sda-commons-client-oidc/README.md
wurstbrot: Password Grant Type is not allowed in OAuth2/2.1.
I recommend to add a note, for example:

> deprecated: A password. Only used for grant type 'password' (must not be used in OAuth2.1).

sda-commons-client-oidc/README.md
eikenilsknopp: As mentioned in the documentation, issuerUrl should be fine. https://openid.net/specs/openid-connect-discovery-1_0.html#ProviderConfig

sda-commons-client-oidc/README.md
eikenilsknopp: I marked it as deprecated in the documentation, I think we have some use cases where customers still rely on versions older than 2/2.1

sda-commons-client-jersey/README.md
```diff
@@ -213,6 +213,104 @@ In addition, each client can consume the standard [proxy system properties](http
...
+OidcResult oidcResult = oidcClient.createAccessToken();
+String accessToken = oidcResult.getAccessToken();
+```
+
+An easy way to integrate the OidcClient into existing PlatformClients is by creating a ClientRequestFilter
+and passing it to the PlatformClientBuilder:
+
+
+```java
+public class OidcRequestFilter implements ClientRequestFilter {
```
christopher-cudennec, line 239: Shouldn't we provide this class instead of just given the example code in the documentation?

sda-commons-client-jersey/README.md
```diff
@@ -213,6 +213,104 @@ In addition, each client can consume the standard [proxy system properties](http
...
+- _OIDC_DISABLED_
+    * Disable retrieving a new token completely. Not meant for production! 
+    * Example: `false`
+
+- _OIDC_GRANT_TYPE_
+    * Sets the OIDC grant type.
+    * Example: `client_credentials`
+    * Supported values:
+        * `client_credentials`
+        * *Deprecated*: `password`
```
christopher-cudennec, line 290: You could also describe the default value here. Maybe just use "Default:" instead of "Example:".

sda-commons-client-jersey/src/main/java/org/sdase/commons/client/jersey/oidc/OidcClient.java
```diff
@@ -0,0 +1,159 @@
...
+    this.cache =
+        Caffeine.newBuilder()
+            .expireAfter(
+                new Expiry<String, TokenResource>() {
+
+                  @Override
+                  public long expireAfterCreate(
+                      @NonNull String key, @NonNull TokenResource value, long currentTime) {
+                    long expirationTime =
+                        value.getAccessTokenExpiresInSeconds() * NANOSECOND_CONVERSION_FACTOR;
```
christopher-cudennec, line 72: Can't you use `Duration#toNanos` for this conversion?

sda-commons-client-jersey/src/main/java/org/sdase/commons/client/jersey/oidc/OidcClient.java
```diff
@@ -0,0 +1,159 @@
...
+                    long buffer = 5 * NANOSECOND_CONVERSION_FACTOR;
+                    return (expirationTime - buffer) > 0 ? (expirationTime - buffer) : 0;
```
christopher-cudennec, line 74: You could also use `Math.max(0, expirationTime - buffer)` for this expression.

sda-commons-client-jersey/src/main/java/org/sdase/commons/client/jersey/oidc/OidcClient.java
```diff
@@ -0,0 +1,159 @@
...
+import org.sdase.commons.client.jersey.oidc.rest.model.OpenIdDiscoveryResource;
+import org.sdase.commons.client.jersey.oidc.rest.model.TokenResource;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/** Service that can retrieve a new access token. */
+public class OidcClient {
+
+  private static final String CACHE_KEY = "oidcResultKey";
+  private static final long NANOSECOND_CONVERSION_FACTOR = 1000000000;
```
JoergSiebahn, line 32: NIT: A little bit more readability:
```suggestion
  private static final long NANOSECOND_CONVERSION_FACTOR = 1_000_000_000;
```

sda-commons-client-jersey/src/main/java/org/sdase/commons/client/jersey/oidc/OidcClient.java
```diff
@@ -0,0 +1,159 @@
...
+                    return currentDuration;
+                  }
+                })
+            .build();
+  }
+
+  /**
+   * Retrieves a new access token from the token endpoint using the given {@link OidcConfiguration}.
+   */
+  public synchronized OidcResult createAccessToken() {
```
christopher-cudennec, line 101: I think it would be nicer for transparency / good design to move the code for the actual retrieval to a separate class.

sda-commons-client-jersey/README.md
eikenilsknopp: Im not sure, I'd rather just give an example and leave it up to the developer how he wants to implement it into the application.

sda-commons-client-jersey/README.md
```diff
@@ -213,6 +213,110 @@ In addition, each client can consume the standard [proxy system properties](http
...
+- _OIDC_CLIENT_ID_
+    * The client id.
+     * Example: `client_id`
+- _OIDC_CLIENT_SECRET_
+    * The client secret.
+    * Example: `s3cr3t`
+     
+- _OIDC_USERNAME_
+    * *Deprecated*: A username. Only used for grant type 'password'.
+        * Example: `john`
```
christopher-cudennec, line 303: Indentation looks strange.

sda-commons-client-jersey/README.md
JoergSiebahn: Maybe it is something for a follow up PR when it is needed/used in more than one service?

sda-commons-client-jersey/README.md
```diff
@@ -213,6 +213,110 @@ In addition, each client can consume the standard [proxy system properties](http
...
+- _OIDC_GRANT_TYPE_
+  * Sets the OIDC grant type.
+  * Default: `client_credentials`
+  * Supported values:
+    * `client_credentials`
+    * *Deprecated*: `password`
+
+- _OIDC_CLIENT_ID_
+  * The client id.
+    * Example: `client_id`
```
christopher-cudennec, line 296: Sorry to be so picky: The bullet points of the env variables are still not aligned completely. Sometimes "Example:" is indented, sometimes it's not.

sda-commons-client-jersey/src/main/java/org/sdase/commons/client/jersey/oidc/OidcConfiguration.java
```diff
@@ -0,0 +1,166 @@
...
+    if (!"client_credentials".equals(grantType) && !"password".equals(grantType)) {
+      return false;
+    }
+
+    if ("password".equals(grantType) && (username == null || password == null)) { // NOSONAR
+      return false;
+    }
+
+    return true;
+  }
```
JoergSiebahn, line 165: I would like to improve the validation to get more useful error messages when something goes wrong. As my idea requires changes at the beginning and the end, my suggestion comes for the whole file:

```java
package org.sdase.commons.client.jersey.oidc;

import javax.validation.Valid;
import javax.validation.constraints.AssertTrue;
import org.apache.commons.lang3.StringUtils;
import org.sdase.commons.client.jersey.HttpClientConfiguration;
import org.sdase.commons.client.jersey.oidc.cache.CacheConfiguration;

/**
 * Supports two grant types:
 *
 * <h2>client_credentials</h2>
 *
 * Can be used for technical users. You only need to provide {@link #clientId} and {@link
 * #clientSecret}.
 *
 * <h2>password</h2>
 *
 * In addition to your client credentials you can specify {@link #username} and {@link #password}
 */
public class OidcConfiguration {

  private static final String GRANT_TYPE_CLIENT_CREDENTIALS = "client_credentials";
  private static final String GRANT_TYPE_PASSWORD = "password";

  private boolean disabled;

  /**
   * Supported grant types:
   *
   * <ul>
   *   <li>client_credentials
   *   <li>password
   * </ul>
   */
  private String grantType = GRANT_TYPE_CLIENT_CREDENTIALS;

  private String clientId;

  private String clientSecret;

  /** Store 'client_id' + 'client_secret' in the authorization header using Basic auth. */
  private boolean useAuthHeader;

  private String username;

  private String password;

  private CacheConfiguration cache = new CacheConfiguration();

  /**
   * Contains the URL to the OpenID provider configuration document, e.g. {@code
   * https://<keycloak.domain>/auth/realms/<realm>} for Keycloak
   */
  private String issuerUrl;

  @Valid private HttpClientConfiguration httpClient = new HttpClientConfiguration();

  public boolean isDisabled() {
    return disabled;
  }

  public OidcConfiguration setDisabled(boolean disabled) {
    this.disabled = disabled;
    return this;
  }

  public String getGrantType() {
    return grantType;
  }

  public OidcConfiguration setGrantType(String grantType) {
    this.grantType = grantType;
    return this;
  }

  public String getClientId() {
    return clientId;
  }

  public OidcConfiguration setClientId(String clientId) {
    this.clientId = clientId;
    return this;
  }

  public String getClientSecret() {
    return clientSecret;
  }

  public OidcConfiguration setClientSecret(String clientSecret) {
    this.clientSecret = clientSecret;
    return this;
  }

  public boolean isUseAuthHeader() {
    return useAuthHeader;
  }

  public OidcConfiguration setUseAuthHeader(boolean useAuthHeader) {
    this.useAuthHeader = useAuthHeader;
    return this;
  }

  public String getUsername() {
    return username;
  }

  public OidcConfiguration setUsername(String username) {
    this.username = username;
    return this;
  }

  public String getPassword() {
    return password;
  }

  public OidcConfiguration setPassword(String password) {
    this.password = password;
    return this;
  }

  public String getIssuerUrl() {
    return issuerUrl;
  }

  public OidcConfiguration setIssuerUrl(String issuerUrl) {
    this.issuerUrl = issuerUrl;
    return this;
  }

  public HttpClientConfiguration getHttpClient() {
    return httpClient;
  }

  public OidcConfiguration setHttpClient(HttpClientConfiguration httpClient) {
    this.httpClient = httpClient;
    return this;
  }

  public CacheConfiguration getCache() {
    return cache;
  }

  public OidcConfiguration setCache(CacheConfiguration cache) {
    this.cache = cache;
    return this;
  }

  @AssertTrue(message = "httpClient is required")
  public boolean isHttpClientValid() {
    if (disabled) {
      return true;
    }

    return httpClient != null;
  }

  @AssertTrue(message = "grantType must be one of client_credentials or password")
  public boolean isGrantTypeValid() {
    if (disabled) {
      return true;
    }

    return GRANT_TYPE_CLIENT_CREDENTIALS.equals(grantType) || GRANT_TYPE_PASSWORD.equals(grantType);
  }

  @AssertTrue(message = "clientId is required")
  public boolean isClientIdValid() {
    if (disabled) {
      return true;
    }

    return StringUtils.isNotBlank(clientId);
  }

  @AssertTrue(message = "clientSecret is required")
  public boolean isClientSecretValid() {
    if (disabled) {
      return true;
    }

    return StringUtils.isNotBlank(clientSecret);
  }

  @AssertTrue(message = "username is required for grant type password")
  public boolean isUsernameValid() {
    if (disabled) {
      return true;
    }
    if (!GRANT_TYPE_PASSWORD.equals(grantType)) {
      return true;
    }

    return StringUtils.isNotBlank(username);
  }

  @AssertTrue(message = "password is required for grant type password")
  public boolean isPasswordValid() {
    if (disabled) {
      return true;
    }
    if (!GRANT_TYPE_PASSWORD.equals(grantType)) {
      return true;
    }

    return StringUtils.isNotBlank(password);
  }
}
```

eikenilsknopp: Looks good, I added this as well as the [actual validation](https://github.com/SDA-SE/sda-dropwizard-commons/pull/858/files#diff-54d86ea4920af6f948dcbcdcc3b849a5ca9e248f894d644e1ab2b3bad309788fR28) so it produces an exception if there is any misconfiguration.

JoergSiebahn (approved):
From my point of view this PR looks fine now. Thanks for the contribution.

@christopher-cudennec do you like to take a final look?

sda-commons-client-jersey/src/main/java/org/sdase/commons/client/jersey/oidc/rest/IssuerClient.java
```diff
@@ -0,0 +1,96 @@
...
+  /**
+   * Retrieves a TokenResource from the configured OIDC provider.
+   *
+   * @return A TokenResource containing the accessToken, tokenType and its expirationTime.
+   */
+  public TokenResource getTokenResource() {
+    OpenIdDiscoveryResource discoveryResource = discoveryApi.getConfiguration();
+
+    if (discoveryResource == null || discoveryResource.getTokenEndpoint() == null) {
+      throw new IllegalStateException("Error loading discovery configuration");
```
christopher-cudennec, line 50: Have you tested the behaviour? I'd thought we'd just return `null` because that return type is already handled as ERROR state. What happens if you throw the exception?

sda-commons-client-jersey/src/main/java/org/sdase/commons/client/jersey/oidc/OidcClient.java
```diff
@@ -0,0 +1,84 @@
...
+    this.config = config;
+
+    try (ValidatorFactory validatorFactory = Validation.buildDefaultValidatorFactory()) {
+      Validator validator = validatorFactory.getValidator();
+      Set<ConstraintViolation<OidcConfiguration>> violations = validator.validate(config);
+
+      if (!violations.isEmpty()) {
+        throw new ConstraintViolationException(violations);
+      }
+    }
```
christopher-cudennec, line 35: Is this really necessary? I think Dropwizard will already do that for us. If the config is invalid the app will not start up. Of course, users of the OIDC client need to add `@Valid` to the connfiguration in their configuration class.

eikenilsknopp: Right, with `@Valid ` in the configuration we don't need this manual validation, I'll remove it.

sda-commons-client-jersey/src/main/java/org/sdase/commons/client/jersey/oidc/rest/IssuerClient.java
eikenilsknopp: Right, this way its somewhat inconsistent as sometimes we are throwing exceptions and sometimes setting the `ERROR` state. I'll simply return null here to keep it consistent.

christopher-cudennec (approved):
Hooray! 🚀

PR #881, opened 2021-03-29
Update LICENSE

wurstbrot:


christopher-cudennec:
Thanks for pointing out. I'll double-check with the legal team.

christopher-cudennec:
This is the answer from our legal team:

> Background: 
>
> The copyright notice origins in the USA and there (until 1983) it was necessary to put a copyright notice on a creation. Otherwise one would not have the right to claim copyright to it!   Since 1983 there is no need for it, but as the copyright is allowed to be sold, the copyright notice is used to show who is the current holder.
>
> How is it in Germany?
>
> In Germany there is no need to use a copyright notice at all. Why? because the “Urheberschaft” is something which cannot be sold or lost. It is “glued” to the creator forever. However the copyright notice in Germany still makes sense as it is understood as an indicator that the user of the copyright mark is the “Urheber”. Therefore it makes sense to use the copyright mark.
> 
> With year or without the year? 
>
> The right way would be to use the year of the GoLive. Example: If a website went live in 2012 it should say in 2021 “Copyright 2012”. BUT: In he US or other Non-EU jurisdictions it might be understood as the copyright was only in effect in 2012. This is why creations or products aimed for an international audience should bear a copyright like this: Copyright (GoLive date) - (today’s date) , e.g. “Copyright 2012 - 2021” or “Copyright SDA SE 2012 - 2021” (examples) or at least the current year “2021”.

christopher-cudennec:
So two things about your suggested change:
* could you change the commit message to use the sematic prefix "chore:"?
* could you change the year according to the suggestion of our legal team to `2019-2021`?

Thanks a lot!

christopher-cudennec (approved):

PR #911, opened 2021-04-20
fix(server-auth-testing): Increase default OPA client timeout for test stability

christopher-cudennec:


JoergSiebahn (approved):
Trying the snapshot in one of our services that fails quite often with OPA errors looks very promising. Let's give it a try and adjust when errors occur again.

PR #922, opened 2021-04-28
feat(auth): add ability to read static public keys

JoergSiebahn:
In addition to read the public keys from static certificates, we are now able to read public keys as well.

- [x] Add tests


sda-commons-server-auth/src/test/java/org/sdase/commons/server/auth/key/PemKeySourceTest.java
```diff
@@ -38,7 +38,7 @@ public void shouldLoadPemKeyFromHttp() {
   }
 
   @Test
-  public void shouldLoadPemKeyFromFile() {
+  public void shouldLoadPemKeyFromCertifcateFile() {
```
dhenneke, line 41:
```suggestion
  public void shouldLoadPemKeyFromCertificateFile() {
```

sda-commons-server-auth/src/main/java/org/sdase/commons/server/auth/key/PemKeySource.java
```diff
@@ -91,4 +104,46 @@ private RSAPublicKey extractRsaPublicKey(X509Certificate certificate)
...
+
+  private RSAPublicKey loadPublicKey(URI pemKeyLocation) {
+    LOG.info("Loading public key for token signature verification from PEM {}", pemKeyLocation);
+    try (InputStream is = pemKeyLocation.toURL().openStream()) {
+      String pemPublicKeyContent = readContent(is);
+      String publicKeyPem =
+          pemPublicKeyContent
+              .replace("-----BEGIN PUBLIC KEY-----", "")
+              .replaceAll("(\\r\\n|\\r|\\n)", "")
+              .replace("-----END PUBLIC KEY-----", "");
```
dhenneke, line 126: Sad that there seems to be no build-in function that supports this format. We probably don't want to add a library just for this use case so I guess it's fine.

sda-commons-server-auth/src/test/java/org/sdase/commons/server/auth/key/PemKeySourceTest.java
JoergSiebahn: [Second try](https://github.com/SDA-SE/sda-dropwizard-commons/compare/adc53e5ae7f9a988b3b3a2ab7ef7df9d9fe374bb..438dd0cfa63434ef9b07d670c37017209d0d8e40).

sda-commons-server-auth/src/main/java/org/sdase/commons/server/auth/key/PemKeySource.java
JoergSiebahn: I'm not aware of any lib we already have as a dependency in this module. My best guess would be to use Bouncy Castle, but that brings in a lot of stuff.

sda-commons-server-auth/src/main/java/org/sdase/commons/server/auth/key/PemKeySource.java
```diff
@@ -91,4 +104,46 @@ private RSAPublicKey extractRsaPublicKey(X509Certificate certificate)
...
+      return (RSAPublicKey) keyFactory.generatePublic(keySpec);
+    } catch (ClassCastException
+        | IOException
+        | NoSuchAlgorithmException
+        | InvalidKeySpecException e) {
+      throw new KeyLoadFailedException("Failed to load public key at " + pemKeyLocation, e);
+    }
+  }
+
+  private String readContent(InputStream content) throws IOException {
```
christopher-cudennec, line 140: What's our policy of just using `IOUtils#toString` for these basic operations?

JoergSiebahn: Apache Commons IO is not in the classpath of the Auth module and I'm a bit wary with Guava as they introduce breaking changes quite often. This pattern is quite common unless we fully upgrade to Java 11.

christopher-cudennec: I also wouldn't go with Guava. Ok, then I'll approve it :)

christopher-cudennec (approved):

PR #932, opened 2021-05-04
test: added test to validate jetty cors filter pattern matching for subdomain and domain name wildcard usage.

nolem:
…ubdomains and hostname with wildcards.


Fox32 (approved):
Looks good, thanks

PR #935, opened 2021-05-05
fix(server-jackson): fix link creation of HalLinkProvider

JoergSiebahn:
HalLinkProvider can create URIs for methods with params that are not annotated as `@QueryParam` or `@PathParam` now. These method parameters are ignored when building the URI.

### DIFF
--- a/sda-commons-server-jackson/src/main/java/org/sdase/commons/server/jackson/hal/HalLinkProvider.java
+++ b/sda-commons-server-jackson/src/main/java/org/sdase/commons/server/jackson/hal/HalLinkProvider.java
@@ -95,7 +95,7 @@ public class HalLinkProvider implements Feature {
     try {
       if (invocation != null || !methodInvocationState.isProcessed()) {
         // Deprecated type needed for backward compatibility until deprecated package is removed
-        // afterwards the the class from this package can be used
+        // afterwards the class from this package can be used
         throw new org.sda.commons.server.jackson.hal.HalLinkMethodInvocationException(
             "No proxied method invocation processed.");
       }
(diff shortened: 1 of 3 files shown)


sda-commons-server-jackson/src/main/java/org/sdase/commons/server/jackson/hal/HalLinkProvider.java
```diff
@@ -92,23 +92,28 @@ public static LinkResult linkTo(Object invocation) {
   private LinkResult linkToInvocation(Object invocation) {
     final HalLinkInvocationStateUtility.MethodInvocationState methodInvocationState =
         HalLinkInvocationStateUtility.loadMethodInvocationState();
-    if (invocation != null || !methodInvocationState.isProcessed()) {
-      // Deprecated type needed for backward compatibility until deprecated package is removed
-      // afterwards the the class from this package can be used
+    try {
+      if (invocation != null || !methodInvocationState.isProcessed()) {
+        // Deprecated type needed for backward compatibility until deprecated package is removed
+        // afterwards the the class from this package can be used
```
christopher-cudennec, line 98: "the the"

christopher-cudennec (commented):
Looks good to me. Can you remember the reason why other parameters were not supported initially?

sda-commons-server-jackson/src/main/java/org/sdase/commons/server/jackson/hal/HalLinkProvider.java
JoergSiebahn: fixed

JoergSiebahn:
> Looks good to me. Can you remember the reason why other parameters were not supported initially?

I guess to tackle mistakes in the method signature/annotation combination. But the URI builder is doing the validation quite well. Maybe there was also an expectation that GET endpoints have nothing else than query params and params.

christopher-cudennec (approved):

PR #998, opened 2021-06-10
feat/fix-platform-key-annotation

ahoeing:


sda-commons-server-key-mgmt/src/main/java/org/sdase/commons/keymgmt/validator/PlatformKey.java
```diff
@@ -22,7 +23,7 @@
  *
  * <p>The validator accepts <code>null</code> values.
  */
-@Target({FIELD, PARAMETER})
+@Target({ METHOD, FIELD, ANNOTATION_TYPE, CONSTRUCTOR, PARAMETER, TYPE_USE })
```
JoergSiebahn, line 26: Is there anything that can be validated at the constructor or are the Hibernate docs missing something?

```suggestion
@Target({ FIELD, METHOD, PARAMETER, ANNOTATION_TYPE, TYPE_USE })
```
Taken from here: https://docs.jboss.org/hibernate/stable/validator/reference/en-US/html_single/?v=7.0#validator-customconstraints-constraintannotation

Note: Spotless strikes back again.

ahoeing: I took the list from org.hibernate.validator.internal.constraintvalidators.bv.number.sign.NegativeOrZeroValidatorForBigInteger but I am fine with the suggestion

christopher-cudennec (commented):
Do you want to add some tests to make sure all targets are supported and work as expected?

ahoeing:
I added the tests for the options used by dropwizard. I think it is sufficient.

christopher-cudennec (approved):

PR #1030, opened 2021-07-01
feat: added implementation and configuration for issuer verification of jwt token. Added new attribute requiredIssuer to auth keys configuration.

nolem:


Fox32:
Is this something one could validate in an OPA policy?

sda-commons-server-auth/src/main/java/org/sdase/commons/server/auth/config/KeyLocation.java
```diff
@@ -38,6 +38,12 @@
    */
   private String pemKeyId;
 
+  /**
+   * Optional <a href="https://openid.net/specs/openid-connect-core-1_0.html#IDToken">iss</a> that
+   * specifies the required issuer of the jwt token, if not empty.
+   */
+  private String requiredIssuer;
```
frederic-huebner, line 45: Maybe you could add an example to the README.md.

nolem:
> Is this something one could validate in an OPA policy?

If we only need a single issuer I would solve this by opa. But we will use a list of discovery/jwks endpoints. 
I would assume, this is far more complex with pure opa and I got at this point not a real idea how to realize it. We need to validate the the Key, that has been successfully validated the signature of the token and for this key we need to validate the correlated issuer. Finally we need a list of x keys and any of this has a different issuer / key. Jörg and me was discussing this for some time and this Jörg mentioned that was a missing security feature to ensure the validity of the issuer. And opa is in this case a optional feature for the commons. So this feature should make sense. But if we find a more lightweight solution it would be fine.

sda-commons-server-auth/README.md
```diff
@@ -89,6 +89,9 @@ Multiple sources for public keys for verification of the signature can be config
 The authentication can be disabled for use in test and development environments. **Be careful to NEVER disable
 authentication in production.**
 
+For the authentication provider root URL and a jwks source a required issuer can be configured by the attribute `requiredIssuer`.
+If the attribute `requiredIssuer` is set, the issuer of the token must match to the provided required issuer, and
+the host name of the source url mast be equal to the host name of the token issuer.
```
christopher-cudennec, line 94: typo: "must"

sda-commons-server-auth/src/main/java/org/sdase/commons/server/auth/key/RsaPublicKeyLoader.java
```diff
@@ -1,6 +1,6 @@
 package org.sdase.commons.server.auth.key;
 
-import java.security.interfaces.RSAPublicKey;
+import com.google.common.collect.Lists;
```
christopher-cudennec, line 3: I'd prefer a simple `new ArrayList(...)` instead of the dependency to the google lib.

sda-commons-server-auth/src/main/java/org/sdase/commons/server/auth/service/AuthRSA256Service.java
```diff
@@ -68,6 +80,82 @@ public AuthRSA256Service(RsaPublicKeyLoader rsaPublicKeyLoader, long leeway) {
...
+   * @return <code>true</code> if the jwt issuer is the same as the configured requiredIssuer and
+   *     the baseUrl is the same as the {@link KeySource} uri or the configuration attribute
+   *     requiredIssuer is not set. Otherwise <code>false</code> is returned.
+   */
+  private boolean validateIssuer(DecodedJWT decodedJWT, LoadedPublicKey loadedPublicKey) {
+
+    if (StringUtils.isNotBlank(loadedPublicKey.getRequiredIssuer())
+        && (!StringUtils.equals(decodedJWT.getIssuer(), loadedPublicKey.getRequiredIssuer())
+            || !validateIssuerAgainstKeySourceUrl(
+                decodedJWT.getIssuer(), loadedPublicKey.getKeySource()))) {
```
christopher-cudennec, line 99: * I'd prefer an early exit in case the required issuer is blank to simplify this statement.
* What's the benefit of `StringUtils#equals` here? Can't we just use `loadedPublicKey.getRequiredIssuer().equals(...)`?

sda-commons-server-auth/src/main/java/org/sdase/commons/server/auth/service/AuthRSA256Service.java
```diff
@@ -68,6 +80,82 @@ public AuthRSA256Service(RsaPublicKeyLoader rsaPublicKeyLoader, long leeway) {
...
+      LOG.error(
```
christopher-cudennec, line 100: I don't think this should be logged as error because you're handling that case properly. I'd rather use WARN or INFO. Compare this article: https://www.section.io/engineering-education/how-to-choose-levels-of-logging/

sda-commons-server-auth/src/main/java/org/sdase/commons/server/auth/service/AuthRSA256Service.java
```diff
@@ -68,6 +80,82 @@ public AuthRSA256Service(RsaPublicKeyLoader rsaPublicKeyLoader, long leeway) {
...
+   * @param lhsUrl optional nullable {@link String}, that contains a url.
+   * @param rhsUrl optional nullable {@link String}, that contains a url.
+   * @return <code>true</code>, if both arguments are blank or do contain the same domain. Otherwise
+   *     <code>false</code> is returned.
+   */
+  private boolean bothBlankOrHaveSameDomain(String lhsUrl, String rhsUrl) {
+    if (!StringUtils.isAllBlank(lhsUrl, rhsUrl)) {
+      if (StringUtils.isBlank(lhsUrl) ^ StringUtils.isBlank(rhsUrl)) {
+        return false;
+      }
```
christopher-cudennec, line 147: Would be nice to see the same code like in the documentation:
```
if (StringUtils.isAllBlank(lhsUrl, rhsUrl)) {
  return true;
}
...
```

christopher-cudennec (commented):
General comment:
Can you please squash your commits and use the semantic commit scheme: `feat(server-auth): ...`

sda-commons-server-auth/src/main/java/org/sdase/commons/server/auth/service/AuthRSA256Service.java
nolem: In this case I do not agree. The problem of this line is, when you do receive this log message, this is a ultra critical attack on your environment. It means someone tried to use a valid private key to sign the token, that is the same key as your idp is using. As follow ups task you are enforced to change the keyset for your IDP. This is the reason for choosing the the error level. I'm not sure how we want to set the level for this behavior. I'm fine with warn too. But this is not related to the application state, this log level is just security related. How do we log issues like this one in the commons?

sda-commons-server-auth/src/main/java/org/sdase/commons/server/auth/service/AuthRSA256Service.java
nolem: I'm not sure about you intention. Is the early exit what you ment?

sda-commons-server-auth/src/main/java/org/sdase/commons/server/auth/service/AuthRSA256Service.java
christopher-cudennec: @JoergSiebahn : Do you have an opinion on it? With this explanation I would also be fine with ERROR.

sda-commons-server-auth/src/main/java/org/sdase/commons/server/auth/service/AuthRSA256Service.java
christopher-cudennec: Yes. I like early exits, because it makes your code more readable IMHO. It also helps with the "keep left" rule.

JoergSiebahn:
> Is this something one could validate in an OPA policy?

In my opinion, that is not possible in OPA if you grant access to tokens from multiple IDPs. Actually this is a preparation so that you can rely in OPA policies on the issuer value and distinguish different IDPs so that you can provide different constraints for the principals authenticated by the differnt IDPs. If you put additional requirements on the IDP, you could just configure required issuers and derive the public keys by using the `iss` value as base URL of the realm. That would work for example with Keycloak. But the `iss` is defined as `StringOrURI`. So you can't be sure that the value is an URL you can use for discovery.

sda-commons-server-auth/src/main/java/org/sdase/commons/server/auth/service/AuthRSA256Service.java
JoergSiebahn: I see this as a warning because we mitigated the treat correctly. Actually, if someone was able to steal the private key of the IDP, he will most likely set the issuer value correct as well.
In my opinion two cases could be more common here:

- the IDP admin changed the issuer value and did not inform the consumers
- an admin of another configured IDP changed the issuer value with intention or by accident to the same value as the IDP with the mentioned key
 
But [as this log is an error](https://github.com/SDA-SE/sda-dropwizard-commons/pull/1030/files#diff-d7e2444a83694da1cbd5193621217ca2afee1b9091ec0b1e621a1cf35457eb4bR68) too, we could keep error here as well.

sda-commons-server-auth/src/main/java/org/sdase/commons/server/auth/service/AuthRSA256Service.java
```diff
@@ -68,6 +80,84 @@ public AuthRSA256Service(RsaPublicKeyLoader rsaPublicKeyLoader, long leeway) {
...
+   * Validate, that the provided issuer has the same base url as the keySource uri if set. This is
+   * only for the implementations {@link JwksKeySource} and {@link OpenIdProviderDiscoveryKeySource}
+   *
+   * @param issuer The issuer url.
+   * @param keySource tThe {@link KeySource} that is used to verify it's key uri.
+   * @return <code>true</code> if the {@link KeySource} implemenation key uri is null or the base of
+   *     the {@link KeySource} key uri is the same as the issuer. Otherwise <code>false</code> is
+   *     returned.
+   */
+  private boolean validateIssuerAgainstKeySourceUrl(String issuer, KeySource keySource) {
```
JoergSiebahn, line 122: As the issuer is not defined as an URI but as a [String or URI](https://datatracker.ietf.org/doc/html/rfc7519#section-4.1.1), I think this validation is not suitable. Why do we need more validation than an exact match to the configured `requiredIssuer`?

sda-commons-server-auth/src/main/java/org/sdase/commons/server/auth/key/OpenIdProviderDiscoveryKeySource.java
```diff
@@ -88,6 +92,10 @@ public String toString() {
     return "OpenIdProviderDiscoveryKeySource{" + "issuerUrl='" + issuerUrl + '\'' + '}';
   }
 
+  public String getIssuerUrl() {
+    return issuerUrl;
+  }
```
JoergSiebahn, line 97: As said below, I think it is not necessary to validate domains of the issuer in the token compared to the issuer of the configuration because we compare the exact required issuer. So it is not needed to expose the value here.

Maybe logging a warning if requiredIssuer and issuerUrl do not match the same domain on startup could be an alternative.

sda-commons-server-auth/src/main/java/org/sdase/commons/server/auth/key/JwksKeySource.java
```diff
@@ -89,6 +93,10 @@ public String toString() {
     return "JwksKeySource{" + "jwksUri='" + jwksUri + '\'' + '}';
   }
 
+  public String getJwksUri() {
+    return jwksUri;
+  }
+
```
JoergSiebahn, line 99: Probably the same as in [OpenIdProviderDiscoveryKeySource](https://github.com/SDA-SE/sda-dropwizard-commons/pull/1030/files#r667886860):

As said below, I think it is not necessary to validate domains of the issuer in the token compared to the issuer of the configuration because we compare the exact required issuer. So it is not needed to expose the value here.

Maybe logging a warning if requiredIssuer and issuerUrl do not match the same domain on startup could be an alternative.

sda-commons-server-auth/src/main/java/org/sdase/commons/server/auth/service/AuthRSA256Service.java
nolem: finally I removed this line because it was covered by the business logic before.

sda-commons-server-auth/src/main/java/org/sdase/commons/server/auth/service/AuthRSA256Service.java
nolem: Correct. It must be verified, that the string contains a ':' to be a URI by specification. 
The reason for this kind of validation we already discussed by slack. It is, that we may ensure, that the original host of the public keys for the token validation is the same as the issuer. Of course we may remove this check as we might only verify the issuer entry.

sda-commons-server-auth/src/main/java/org/sdase/commons/server/auth/key/OpenIdProviderDiscoveryKeySource.java
nolem: This way sounds better and I would prefer this way to go.

sda-commons-server-auth/src/main/java/org/sdase/commons/server/auth/service/AuthRSA256Service.java
JoergSiebahn: I think that is not necessary in the flexible configuration we provide with this PR. With the current configuration we also support handmade IDPs (although we do not like them) that use phrases like `INSURANCE_X_IDP` as `iss`. In my opinion a warning about a possible misconfiguration on initialization is enough to cover this case and avoid failures in deployment. As the requiredIssuer is final, the exact match against the issuer in the token will fulfill the same restriction, reduces complexity in the token validation code and has a (minimal) positive effect on performance when validating.

Not related to this PR, another type of configuration could cover the case we talked about. But that would cause more refactorings and restricts that type to IDPs like Keycloak that use the base URL of the realm as `iss`. In that case you are able to configure only the required issuer(s) and automatically configure the `OpenIdProviderDiscoveryKeySource`s with required issuer.

sda-commons-server-auth/src/main/java/org/sdase/commons/server/auth/service/AuthRSA256Service.java
```diff
@@ -38,28 +47,31 @@ public AuthRSA256Service(RsaPublicKeyLoader rsaPublicKeyLoader, long leeway) {
...
-          log.warn("Verifying token without kid trying {} public keys", keysWithoutId.size());
+          LOG.warn("Verifying token without kid trying {} public keys", keysWithoutId.size());
         }
         Collections.reverse(keysWithoutId);
         return keysWithoutId.stream()
-            .map(k -> verifyJwtSignature(authorizationToken, k))
-            .filter(Optional::isPresent)
+            .map(k -> Pair.of(verifyJwtSignature(authorizationToken, k.getPublicKey()), k))
+            .filter(p -> p.getKey().isPresent())
+            .filter(p -> validateIssuer(p.getKey().get(), p.getValue()))
```
JoergSiebahn, line 58: You could skip this manual validation (also in the `else` branch). [`JWT.require(…)`](https://github.com/SDA-SE/sda-dropwizard-commons/pull/1030/files#diff-d7e2444a83694da1cbd5193621217ca2afee1b9091ec0b1e621a1cf35457eb4bR165) that is used in `verifyJwtSignature(…)` has a method `withIssuer(…)` that is doing exactly this validation.

sda-commons-server-auth/README.md
```diff
@@ -89,6 +89,9 @@ Multiple sources for public keys for verification of the signature can be config
 The authentication can be disabled for use in test and development environments. **Be careful to NEVER disable
 authentication in production.**
 
+For the authentication provider root URL and a jwks source a required issuer can be configured by the attribute `requiredIssuer`.
+If the attribute `requiredIssuer` is set, the issuer of the token must match to the provided required issuer, and
+the host name of the source url must be equal to the host name of the token issuer.
```
JoergSiebahn, line 94:
```suggestion
If the attribute `requiredIssuer` is set, the issuer of the token must match to the provided required issuer.
A warning will be logged if the host name of the source url does not match the host name of the required issuer.
```

sda-commons-server-auth/src/main/java/org/sdase/commons/server/auth/AuthBundle.java
```diff
@@ -124,18 +127,48 @@ private Client createKeyLoaderClient(Environment environment, AuthConfig config,
...
+          LOG.warn(
+              "The required issuer host name <{}> for the key <{}> does not match to the key"
+                  + " source uri host name <{}>.",
+              issuerUri.getHost(),
+              location,
+              location.getHost());
+        }
+      } catch (URISyntaxException e) {
+        throw new IllegalArgumentException(
+            "The requiredIssuer <" + requiredIssuer + "> is no valid stringOrURI", e);
```
JoergSiebahn, line 167: I would consider this case the same as not matching and log a warning. It may be a valid configuration because not every String with a colon is an URI.

nolem: By specification it it :)

JoergSiebahn: 👍  I'm fine with that, thanks for quoting the full spec.

JoergSiebahn (approved):

PR #1065, opened 2021-07-20
docs(shared-certificates): add README documentation

AnkitaSrivastavaSDA:


sda-commons-server-morphia/README.md
```diff
@@ -97,10 +97,11 @@ mongo:
   username: "${MONGODB_USERNAME:-}"
   password: "${MONGODB_PASSWORD:-}"
   useSsl: ${MONGODB_USE_SSL:-true}
-  caCertificate: "${MONGODB_CA_CERTIFICATE:-}"
```
jeskosda, line 100: As far as I remember, the option to pass the certificate via env string should still be possible, right?

AnkitaSrivastavaSDA: We deprecated it in the code ([MongoConfiguration](https://github.com/SDA-SE/sda-dropwizard-commons/commit/67d1c87b7a1c6425bbe0af42f6972351b0d2f923#diff-b8c0229f8e7925f373c12ffc9bd2322b763ddd8bbc0a1db9457a69dd28d52e90R63)), that's why thought to remove from the example config. Not sure if we should delete it or add the note for deprecation in the README?

JoergSiebahn: Yes, we should still support and mention the existing approach.

jeskosda: Do we really require a deprecation here? I mean for simple deployments it should be still a possibility to pass the certificate via string.

sda-commons-server-morphia/README.md
```diff
@@ -97,10 +97,11 @@ mongo:
...
+  
+caCertificate:
+  customCaCertificateDir: "${CA_CERTIFICATE_DIR:-}"
```
JoergSiebahn, line 102: By default we do not have this configuration for MongoDB. But we use the default configuration of the CaCertificateBundle. So we should document the default directory where certificates can be mounted and that this approach is preferred over setting the caCertificate in the environment.

sda-commons-shared-certificates/README.md
```diff
@@ -0,0 +1,88 @@
...
+```yaml
+caCertificate:
+  customCaCertificateDir: "${CA_CERTIFICATE_CUSTOM_DIR:-}"
+```
+
+Example config for **developer** machines using [local-infra](https://github.com/SDA-SE/local-infra):
+```yaml
+caCertificate:
+  customCaCertificateDir: /var/certs/
+```
```
JoergSiebahn, line 88: I think, we do not need a specific example for local infra. I see no differences or specials with local-infra. Are there any?

sda-commons-server-morphia/README.md
saadmar: then do you think that we should give this in a second exemple and just propose it as an alternative ?

sda-commons-shared-certificates/README.md
```diff
@@ -0,0 +1,88 @@
...
+
+The directory that contains CA certificates in PEM format is configured in the `config.yaml` 
+of the final application config. If no custom configuration is provided, then the default directory 
+`DEFAULT_TRUSTED_CERTIFICATES_DIR = "/var/trust/certificates"` is used.
+
+Example config for **production** to be used with environment variables of the cluster configuration:
+
+```yaml
+caCertificate:
+  customCaCertificateDir: "${CA_CERTIFICATE_CUSTOM_DIR:-}"
```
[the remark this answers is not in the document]
AnkitaSrivastavaSDA: Then should I just leave it out now itself?

sda-commons-server-morphia/README.md
JoergSiebahn: As the environment variable / config.yaml option it is still available, we should keep the documentation of it.
I think we should deprecate it, so that it can be removed with the next major release of sda-dropwizard-commons.
If nothing special (namely providing a configuration option) is done by the implementor of a service, the MorphiaBundle will catch up certificates from the default directory of the CaCertificatesBundle. That is a different approach than the configuration. That's because I would mention it in a separate sentence (independent, of the config.yaml example) and document the default directory as mentioned in [the other comment](https://github.com/SDA-SE/sda-dropwizard-commons/pull/1065#discussion_r673091124).

sda-commons-shared-certificates/README.md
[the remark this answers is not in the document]
JoergSiebahn: Well, in my opinion, I would leave out certificate directory configuration in the readme of the MorphiaBundle.
Here, I would show the option to make it configurable (including repeating the default) and in an additional sentence like „If no configuration is used in a service, the default path `/var/trust/certificates` is checked for PEM certificates. This is the preferred approach to keep all services using this library consistent. But configuration may be needed if unit tests must cover the use of a custom SSLContext.“

sda-commons-server-morphia/README.md
jeskosda: Passing the certificate is more error-prone but giving the user of that bundle a less complex possiblity to choose how the certificate is passed. Are there any reasons why we should omit the option? At the end we'll force our customers to change something in there deployment, even if they are happy with that option right now.

sda-commons-shared-certificates/README.md
AnkitaSrivastavaSDA: Done!

sda-commons-server-morphia/README.md
JoergSiebahn: I'll ask the other maintainers for their preference regarding deprecation.

sda-commons-server-morphia/README.md
```diff
@@ -103,6 +103,10 @@ _Please note the double quotes around the values.
  Without them for `caCertificate`, Dropwizard will not be able to load the certificate correctly because the `MONGODB_CA_CERTIFICATE` variable contains line breaks.  
  Omitting the double quotes for other values like `password` can cause trouble in case of certain special characters (e.g. colon)_
 
+ _Instead of setting `caCertificate` in the environment, we should use the default configuration of the
+ [`CaCertificatesBundle`](../sda-commons-shared-certificates/src/main/java/org/sdase/commons/shared/certificates/ca/CaCertificatesBundle.java),
+ and mount CA Certificates to the default directory `DEFAULT_TRUSTED_CERTIFICATES_DIR = "/var/trust/certificates"`_
+
```
saadmar, line 109:
```suggestion
#### CA Certificates support

Instead of providing `caCertificate` as an environment variable, we can add Ca certificates PEM files in the directory `/var/trust/certificates`. Certificates available in sub-directories will also be loaded.

Note that this directory is also configurable through the Dropwizard config class. The config class should then provide a 
[`CaCertificateConfiguration`](../sda-commons-shared-certificates/src/main/java/org/sdase/commons/shared/certificates/ca/CaCertificateConfiguration.java) to the bundle builder.
See [`sda-commons-shared-certificates`](../sda-commons-shared-certificates/README.md) for details.

```
I think, if we are still preferring  the old way other the new one (if the are both used) then we should just present it as an option in morphia bundle documentation, and if we want to deprecate it in an other release (change the order of preference), at least then we would be sure that it is stable after at least testing that in one service.

JoergSiebahn: Deprecation is not the same as preference. The deprecation is only an announcement that an API will be removed later. Until removal (or at least until the next major release) everything should behave the same as before so that users are not forced to change anything. So preferring the old way over the new way complies with the deprecation in my opinion. Even when we are suggesting to change configuration to use the new way.

saadmar: I get your point now

JoergSiebahn: We should also remember to announce the deprecation in the release notes after merge by editing them in the GitHub UI. Unfortunately, semantic release is not able to handle deprecations from commit messages. But I also prefer that we keep one commit that includes information about the deprecation in the final PR to master. But we will update the commit history in that PR anyways.

AnkitaSrivastavaSDA: Sure, will keep in mind!

JoergSiebahn (approved):

PR #1068, opened 2021-07-21
feat: add prometheus metric healthcheck_status_overall

thrstnh:
This pull request adds another prometheus metric to get the overall healthstatus of the dropwizard service, thus aggregating the separate healthchecks into one overall value.

# Background
We are trying to measure service availability based on dropwizard healthchecks via prometheus/grafana. 
Currently each singular healthcheck component is exposed as a separate metric, however, any failed healthcheck component results in the service being not available (kubernetes readiness probe).
So we're interested in the minimum value across all healthcheck components. This could be calculated in our prometheus query, but we would like to avoid this calculation in prometheus as it maybe run fairly frequently (alerts) or over long time frames. From our point of view this aggregated view is much more useful for monitoring than each singular healthcheck component.
Therefore we propose to expose this aggregated metric in the first place.


christopher-cudennec:
Hi @thrstnh ! Thanks for your PR!
Some small comments:
* some documentation in the README would be nice
* could you rephrase your commit to "feat(server-prometheus): ..." to comply to our semantic commit messages?
* sooner or later you need to rebase your PR

sda-commons-server-prometheus/src/test/java/org/sdase/commons/server/prometheus/health/HealthCheckMetricsCollectorTest.java
```diff
@@ -11,23 +11,34 @@
...
+    collector = new HealthCheckMetricsCollector(registry);
+  }
 
+  private void initHealthcheckRegistryMock(int numHealthy, int numUnhealthy) {
     SortedMap<String, Result> results = new TreeMap<>();
-    results.put("healthy_check", Result.healthy());
-    results.put("unhealthy_check", Result.unhealthy("Something went wrong"));
+    for (int i = 0; i < numHealthy; i++) results.put("healthy_check_" + i, Result.healthy());
+    for (int i = 0; i < numUnhealthy; i++)
+      results.put("unhealthy_check_" + i, Result.unhealthy("Something went wrong"));
```
dhenneke, line 32: Please always use braces (see https://google.github.io/styleguide/javaguide.html#s4.1.1-braces-always-used). You might want to run `./gradlew spotlessApply` which should add it automatically.

thrstnh: Thank you!
I updated the pull request, used the braces, adjusted the README, renamed the commit to comply to the semantic commit messages and rebased on current master branch.
A final `./gradlew spotlessApply`ended with _build successful_.

Fox32:
> avoid this calculation in prometheus as it maybe run fairly frequently

If you are worrying about this, have a look at [recoding rules](https://prometheus.io/docs/prometheus/latest/configuration/recording_rules/). I see this as very customer specific and is probably something that is easier to solve at an infrastructure level (prometheus) than in a Java service library.

thrstnh:
> > avoid this calculation in prometheus as it maybe run fairly frequently
> 
> If you are worrying about this, have a look at [recoding rules](https://prometheus.io/docs/prometheus/latest/configuration/recording_rules/). I see this as very customer specific and is probably something that is easier to solve at an infrastructure level (prometheus) than in a Java service library.

Hello Fox32,

thanks for the advice.
We tested the metric as a recording rule. The recording rule would be rolled out at the infrastructure level, so that the customers have no access or even knowledge about there being a recording rule, because it is "hidden" in a "$servicegroup-infra" namespace with a pre-configured monitoring instance to which the teams have no access.
So the recording rule is available in the cluster in each monitoring instance for each $servicegroup, which are currently 122 monitoring instances. The recording rule would have to be implemented also for local testing. Here we would see the benefit of receiving the metric directly from the pod, then it would be available out-of-the-box everywhere, where the POD is used.
On the other hand the healthcheck_status_overall would be the minimum availability of the current POD over all specified labels. We think it would be a good addition to the metrics for all customers to make a healthcheck simple and fast.
What do you think about adding the metric to the base service?

PR #1069, opened 2021-07-21
Feat/certificate reader

saadmar:


JoergSiebahn:
fyi: I turned this PR into draft until we tested it in a service in a cluster that requires certificates for the MongoDB connection.

JoergSiebahn (approved):

PR #1165, opened 2021-09-23
feat(client-jersey): add oidc request filter

eikenilsknopp:


eikenilsknopp:
Should the oidc package also provide a classRule/extension for testing so the services do not have to stub the `.well-known` and `token` endpoints manually? Or do we say it has been tested enough in sda-commons already, it should be simply disabled in the services integration tests?

JoergSiebahn (approved):
LGTM

PR #1172, opened 2021-09-28
test(server-kafka): No need to access field modifiers

christopher-cudennec:
Makes it compatible with Java 17

To be honest: I'm a little surprised that it does not seem to be necessary to remove the final modifier. But as long as the tests pass we're good, right? 🤣


JoergSiebahn (approved):
Interesting that removing the modifier was not needed for 8 and 11 as well.

dhenneke:
> Interesting that removing the modifier was not needed for 8 and 11 as well.

Yes. I thought this was needed when I wrote the code. But it is only used internally, so it should be ok as long as the tests still work.

PR #1186, opened 2021-10-04
chore: disable test result publishing for dependabot PRs

JoergSiebahn:
Failing tests will still fail the build, but there is no easy to read information in the PR.

### DIFF
--- a/.github/workflows/java-ci.yml
+++ b/.github/workflows/java-ci.yml
@@ -65,8 +65,8 @@ jobs:
     timeout-minutes: 60
     needs: build
     runs-on: ubuntu-latest
-    # Don't run for forks because of missing secrets
-    if: (success() || failure()) && !(github.event.pull_request && github.event.pull_request.head.repo.fork)
+    # Don't run for forks and dependabot because of missing secrets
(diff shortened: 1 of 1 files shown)


.github/workflows/java-ci.yml
```diff
@@ -65,8 +65,8 @@ jobs:
     timeout-minutes: 60
     needs: build
     runs-on: ubuntu-latest
-    # Don't run for forks because of missing secrets
-    if: (success() || failure()) && !(github.event.pull_request && github.event.pull_request.head.repo.fork)
+    # Don't run for forks and dependabot because of missing secrets
+    if: (success() || failure()) && !(github.event.pull_request && github.event.pull_request.head.repo.fork) && github.actor != 'dependabot[bot]'
```
christopher-cudennec, line 69: Do you know why we need the first condition `success() || failure()`? Sounds like it can be removed completely or do we have a third option?

JoergSiebahn: No, it was there from the beginning. The [action itself](https://github.com/EnricoMi/publish-unit-test-result-action) has `always()` in the example but without additional conditions. I don't know how GH actions evaluate the condition when there is no reference to the previous state in the if clause.

christopher-cudennec (approved):

PR #1245, opened 2021-10-25
Simplify CSP and add sandbox

wurstbrot:
Hardening is important. This PRs improves the header security for an api endpoint:
- simplify CSP
- adding sandbox
according to "API Security in Action" page 58.

Complications with frontend applications like angular are not expected.
In case HTML applications are delivered to the client directly, it will infect for example javascript execution.

### DIFF
--- a/sda-commons-server-security/src/main/java/org/sdase/commons/server/security/filter/WebSecurityHeaderFilter.java
+++ b/sda-commons-server-security/src/main/java/org/sdase/commons/server/security/filter/WebSecurityHeaderFilter.java
@@ -35,16 +35,7 @@ public class WebSecurityHeaderFilter implements ContainerResponseFilter {
     webSecurityHeaders.put("X-Permitted-Cross-Domain-Policies", "none");
     webSecurityHeaders.put(
         "Content-Security-Policy",
-        String.join(
-            "; ",
-            asList(
-                "default-src 'self'",
-                "script-src 'self'",
-                "img-src 'self'",
-                "style-src 'self'",
-                "font-src 'self'",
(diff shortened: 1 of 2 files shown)


christopher-cudennec:
ping @wurstbrot

wurstbrot:
I am not having the dropwizard insights to implement the advanced requirements.

christopher-cudennec:
@JoergSiebahn / @wurstbrot : Are you aware of any application that server static files / use dropwizard-views?

christopher-cudennec:
@kairabiensda : Could you also have a look at the CSP change and the potential effects?

christopher-cudennec:
@wurstbrot : I squashed your commits and used spotless to apply our code formatting rules.

JoergSiebahn:
Services that serve simple server side rendered UIs (where an additional app is way too much) will not work any more. To make it backward compatible this change should be a configurable option. It may be suitable to use this change as a new default but have a config option to restore the old behavior and add an information to the release notes how to activate it.

Another question:
Will this only affect resources served by the service or also involved apps on the same domain? We have cases where we expose single page apps from nginx based images and APIs from services on the same domain by using the same Ingress proxy. Will such scenarios be affected?

JoergSiebahn:
We also found [a discussion on Stackoverflow](https://stackoverflow.com/q/45630376), where the use of CSP for APIs are questioned:

> It's mainly to appease infosec.

> Practically, the advantages may be slim to none.

JoergSiebahn:
@wurstbrot @christopher-cudennec please see #1641 for configuration option. It is built on top of this PR.

PR #1287, opened 2021-11-16
feat: Variable substitution now also supports system properties

christopher-cudennec:
Additionally, this helps moving on to Java 17 because changing environment variables is not really intended

### DIFF
--- a/sda-commons-server-auth-testing/src/test/java/org/sdase/commons/server/auth/testing/KeyLoaderProxyIT.java
+++ b/sda-commons-server-auth-testing/src/test/java/org/sdase/commons/server/auth/testing/KeyLoaderProxyIT.java
@@ -32,11 +32,9 @@ public class KeyLoaderProxyIT {
   private static final WireMockClassRule PROXY_WIRE =
       new WireMockClassRule(wireMockConfig().dynamicPort());
 
-  private static final SystemPropertyRule SYSTEM_PROPERTY_RULE =
-      new SystemPropertyRule().setProperty("AUTH_RULE", "{\"keys\": [{}]}");
-
   private static final SystemPropertyRule SYSTEM_PROPERTY =
       new SystemPropertyRule()
+          .setProperty("AUTH_RULE", "{\"keys\": [{}]}")
           .setProperty("http.proxyHost", "localhost")
           .setProperty("http.proxyPort", () -> "" + PROXY_WIRE.port())
           .setProperty("http.nonProxyHosts", "localhost");
@@ -50,10 +48,7 @@ public class KeyLoaderProxyIT {
 
   @ClassRule
   public static final RuleChain rule =
-      RuleChain.outerRule(PROXY_WIRE)
-          .around(SYSTEM_PROPERTY_RULE)
-          .around(SYSTEM_PROPERTY)
-          .around(DW);
+      RuleChain.outerRule(PROXY_WIRE).around(SYSTEM_PROPERTY).around(DW);
 
   @BeforeClass
   public static void setupWiremock() {
--- a/sda-commons-server-auth-testing/src/test/java/org/sdase/commons/server/auth/testing/KeyLoaderProxyIT.java
+++ b/sda-commons-server-auth-testing/src/test/java/org/sdase/commons/server/auth/testing/KeyLoaderProxyIT.java
@@ -25,7 +25,6 @@ import org.junit.Test;
 import org.junit.rules.RuleChain;
 import org.sdase.commons.server.auth.testing.test.AuthTestApp;
 import org.sdase.commons.server.auth.testing.test.AuthTestConfig;
-import org.sdase.commons.server.testing.EnvironmentRule;
 import org.sdase.commons.server.testing.SystemPropertyRule;
 
 /** A test that checks if the jersey client that is used to load keys can use a proxy server */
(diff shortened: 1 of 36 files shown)


sda-commons-server-auth-testing/src/test/java/org/sdase/commons/server/auth/testing/KeyLoaderProxyIT.java
```diff
@@ -25,16 +25,15 @@
...
 import org.sdase.commons.server.testing.SystemPropertyRule;
 
 /** A test that checks if the jersey client that is used to load keys can use a proxy server */
 public class KeyLoaderProxyIT {
   private static final WireMockClassRule PROXY_WIRE =
       new WireMockClassRule(wireMockConfig().dynamicPort());
 
-  private static final EnvironmentRule ENVIRONMENT_RULE =
-      new EnvironmentRule().setEnv("AUTH_RULE", "{\"keys\": [{}]}");
+  private static final SystemPropertyRule SYSTEM_PROPERTY_RULE =
```
JoergSiebahn, line 35: We have two system property rules now. Can we join them into one?

sda-commons-server-dropwizard/src/main/java/org/sdase/commons/server/dropwizard/bundles/SystemPropertyAndEnvironmentLookup.java
```diff
@@ -0,0 +1,15 @@
...
+
+public class SystemPropertyAndEnvironmentLookup extends EnvironmentVariableLookup {
+
+  @Override
+  public String lookup(String key) {
+    String result = System.getProperty(key);
+    if (result != null) {
+      return result;
+    }
+    return super.lookup(key);
```
JoergSiebahn, line 13: I see this line covered. But do we test if environment variables are the fallback for system properties? And do we test that system properties have higher precedence than environment variables? Should they have higher precedence?
Regarding to the upcoming Java 17: Do we even have a chance to test this behavior?

christopher-cudennec: > But do we test if environment variables are the fallback for system properties?

Not yet. This PR was just a POC to see if we could use System properties all over. I can add it if we agree that we want that change in general.

> Should they have higher precedence?

I used Spring Boots order as orientation. For backwards compatibility it might be an argument to prefer env variables.

>  Regarding to the upcoming Java 17: Do we even have a chance to test this behavior?

It's possible to add some flags to the Java process. That's not my favourite approach.

JoergSiebahn: > I can add it if we agree that we want that change in general.
To be prepared for Java 17, I would agree.

> I used Spring Boots order as orientation.
Sounds like a good argument. I'm not expecting any conflicts in the real world.

> It's possible to add some flags to the Java process.
In my opinion we should do it in sda-dropwizard-commons in the GH Actions and may conditionally ignore the specific test cases with advice for the settings locally if there is some manual setup needed (e.g. in the environment or the IDE).
Would be bad if services can't be configured with env's any more after a few more changes.

sda-commons-server-mongo-testing/src/test/java/org/sdase/commons/server/mongo/testing/UseExistingMongoDbClassExtensionTest.java
```diff
@@ -41,8 +41,10 @@ public void initUseExistingMongoDbExtension() {
             + "?"
             + EXTERNAL_DB.getOptions();
     originalMongoDbConnectionString =
-        System.getenv().get(OVERRIDE_MONGODB_CONNECTION_STRING_ENV_NAME);
-    Environment.setEnv(OVERRIDE_MONGODB_CONNECTION_STRING_ENV_NAME, MONGODB_CONNECTION_STRING);
+        new SystemPropertyAndEnvironmentLookup()
+            .lookup(OVERRIDE_MONGODB_CONNECTION_STRING_SYSTEM_PROPERTY_NAME);
+    System.setProperty(
+        OVERRIDE_MONGODB_CONNECTION_STRING_SYSTEM_PROPERTY_NAME, MONGODB_CONNECTION_STRING);
```
JoergSiebahn, line 47: Do we need to save and reset?

sda-commons-server-auth-testing/src/test/java/org/sdase/commons/server/auth/testing/KeyLoaderProxyIT.java
christopher-cudennec: Yes, done.

sda-commons-server-dropwizard/src/main/java/org/sdase/commons/server/dropwizard/bundles/SystemPropertyAndEnvironmentLookup.java
christopher-cudennec: I added a test for SystemPropertyAndEnvironmentLookup that ensures that system properties have higher priority.

sda-commons-server-mongo-testing/src/test/java/org/sdase/commons/server/mongo/testing/UseExistingMongoDbClassExtensionTest.java
christopher-cudennec: Test test setup is a little strange but the value is reset manually in `AfterEach`. I would rather not touch the code in this PR.

JoergSiebahn: Ups, sorry for the noise. I did not see that the old value is restored.

sda-commons-server-testing/README.md
```diff
@@ -60,8 +60,22 @@ content should always be reproducible. Note that the [AsyncAPI](../sda-commons-s
...
+**WARNING**
+
+Java considers environment variables to be immutable, so this extension uses reflection to change 
+them. This requires that the SecurityManager allows modifications and can potentially break on 
+different operating systems and Java versions. Be aware that this is a fragile solution and consider
+finding a better one for your specific situation.
+
+Most of the time using the `SystemPropertyRule` should also work for you since our Dropwizard setup
+supports looking up environment variables AND system properties when replacing variables in your
+configuration file.
```
JoergSiebahn, line 76: 👍

JoergSiebahn (approved):

PR #1293, opened 2021-11-18
chore: Use Gradle 7.3

christopher-cudennec:
... brings official support for Java 17 🥳


gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,5 +1,5 @@
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-7.2-bin.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-7.3-bin.zip
```
JoergSiebahn, line 3: If you upgrade with `./gradlew wrapper --gradle-version 7.3`, the `gradle-wrapper.jar`, `gradlew` and `gradlew.bat` are modified as well. I'm not an expert, but it seems to me that just changing the version number is not the best solution.

christopher-cudennec: Oh, true. Wasn't really aware of that. Will use the CLI from now on. 😃

JoergSiebahn (approved):

PR #1302, opened 2021-11-25
fix(testing): AuthRule, AuthClassExtension + KafkaBrokerEnvironmentRu…

christopher-cudennec:
…le set environment variables for Java version < 16

This restores the old behavior before PR #1287

### DIFF
--- a/sda-commons-server-auth-testing/src/main/java/org/sdase/commons/server/auth/testing/AuthClassExtension.java
+++ b/sda-commons-server-auth-testing/src/main/java/org/sdase/commons/server/auth/testing/AuthClassExtension.java
@@ -5,16 +5,19 @@ import static org.junit.Assert.fail;
 
 import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.databind.ObjectMapper;
+import java.util.Objects;
+import org.apache.commons.lang3.JavaVersion;
+import org.apache.commons.lang3.SystemUtils;
 import org.junit.jupiter.api.extension.AfterAllCallback;
 import org.junit.jupiter.api.extension.BeforeAllCallback;
 import org.junit.jupiter.api.extension.ExtensionContext;
 import org.sdase.commons.server.auth.config.AuthConfig;
+import org.sdase.commons.server.testing.Environment;
 
 public class AuthClassExtension extends AbstractAuth
     implements BeforeAllCallback, AfterAllCallback {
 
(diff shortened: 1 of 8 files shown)


sda-commons-server-auth-testing/src/main/java/org/sdase/commons/server/auth/testing/AuthClassExtension.java
```diff
@@ -5,16 +5,19 @@
...
 import org.junit.jupiter.api.extension.ExtensionContext;
 import org.sdase.commons.server.auth.config.AuthConfig;
+import org.sdase.commons.server.testing.Environment;
 
 public class AuthClassExtension extends AbstractAuth
     implements BeforeAllCallback, AfterAllCallback {
 
   @SuppressWarnings("WeakerAccess")
-  public static final String AUTH_RULE_ENV_KEY = "AUTH_RULE";
+  private static final String AUTH_RULE_ENV_KEY = "AUTH_RULE";
```
JoergSiebahn, line 20: NIT can we remove the warning above?

sda-commons-server-auth-testing/src/main/java/org/sdase/commons/server/auth/testing/AuthRule.java
```diff
@@ -104,30 +107,38 @@ private void init() {
...
       fail("Failed to create the config keys: " + e.getMessage());
     }
   }
 
+  private void initTestAuth(String authConfig) {
+    final TestRule testRule = createTestRule(authConfig);
+    delegate = RuleChain.outerRule(testRule);
+  }
+
+  public static TestRule createTestRule(String authConfig) {
```
JoergSiebahn, line 130: Could we use less visibility to make it internal or is anybody required to use this method apart from the internal tests?

christopher-cudennec: Package private should also work. Thanks for pointing it out! 👍

sda-commons-server-auth-testing/src/main/java/org/sdase/commons/server/auth/testing/AuthClassExtension.java
christopher-cudennec: Yeah. Just a leftover. Will remove it.

JoergSiebahn (approved):

PR #1308, opened 2021-11-29
fix(server-kafka): Commit next offset to process on close

christopher-cudennec:
Instead of committing the offset of the last poll


sda-commons-server-kafka/src/main/java/org/sdase/commons/server/kafka/consumer/strategies/MessageListenerStrategy.java
```diff
@@ -31,11 +42,14 @@ public void init(ConsumerTopicMessageHistogram consumerTopicMessageHistogram) {
    * Implementation of processing and commit logic during poll loop of {@link MessageListener}.
    *
    * <p>The strategy should collect the processing duration metric for each entry of records within
-   * the consumerProcessedMsgHistogram.
+   * the consumerProcessedMsgHistogram. Furthermore, each record that was process successfully
```
eikenilsknopp, line 45: Typo: processed

sda-commons-server-kafka/src/test/java/org/sdase/commons/server/kafka/consumer/strategies/MessageListenerStrategyOnCloseTest.java
```diff
@@ -0,0 +1,144 @@
...
+    recordsByPartition.put(
+        new TopicPartition(TOPIC_NAME, 0),
+        Arrays.asList(
+            new ConsumerRecord<>(TOPIC_NAME, 0, 0, "key0", "value0"),
+            new ConsumerRecord<>(TOPIC_NAME, 0, 1, "key1", "value1"),
+            new ConsumerRecord<>(TOPIC_NAME, 0, 2, "key2", "value2")));
+
+    LOGGER.info("Starting thread for strategy");
+    final KafkaConsumer<String, String> consumer = mock(KafkaConsumer.class);
+    new Thread(
```
eikenilsknopp, line 83: Threads and timings always feel like they could introduce flakiness, but I assume this polling each 100ms and sleeping for 1s works out fine, right?

christopher-cudennec: Yeah, I also don't really like using "sleep" very much. But somehow I wanted to write a test that reproduces an application that does some longer processing and is shutdown in the middle of the process without being able to finish the other records from the last poll.

sda-commons-server-kafka/src/main/java/org/sdase/commons/server/kafka/consumer/strategies/MessageListenerStrategy.java
christopher-cudennec: Fixed.

eikenilsknopp (approved):

PR #1325, opened 2021-12-06
fix: exclude Jackson Afterburner module in the latest JDKs

JoergSiebahn:
Afterburner is not compatible with Java 17.
A warning was logged since Java 9.

There is intentionally no (conditional) change in the [`omShouldHaveSameModulesAsDefaultFromDropwizardButNoFuzzyEnumModule `](https://github.com/SDA-SE/sda-dropwizard-commons/blob/master/sda-commons-server-jackson/src/test/java/org/sdase/commons/server/jackson/TolerantObjectMapperTest.java#L33) so that we get notified when we introduce a Java 17 build and can decide what to do later.


christopher-cudennec:
We could put it on the list for the next major release of sda-dropwizard-commons: https://sda-se.atlassian.net/wiki/spaces/platform/pages/3070394432/sda-dropwizard-commons+3

christopher-cudennec (approved):

PR #1401, opened 2022-01-18
Verify that UTF-8 characters are returned for openapi.yaml

JoergSiebahn:
There seems to be an issue with OpenAPI-Tests when run on windows regarding UTF-8 characters.

- This PR reproduces UTF-8 issues [in the Windows Build](https://github.com/SDA-SE/sda-dropwizard-commons/runs/4864579329?check_suite_focus=true#step:4:9500) 
- and introduces a fix by [defining the source encoding](https://github.com/SDA-SE/sda-dropwizard-commons/pull/1401/commits/34fd56c295ecec88e65390ecafc1ce8146319950).

To solve the issue, the encoding must be set in the `build.gradle` of each affected service. I think there is no way to provide the encoding setting for source code with sda-commons.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -175,6 +175,10 @@ configure(subprojects.findAll { !javaPlatformModules.contains(it.name) }) {
 
   sourceCompatibility = JavaVersion.VERSION_1_8
 
+  tasks.withType(JavaCompile) {
+    options.encoding = 'UTF-8'
+  }
+
   // Configure JaCoCo to export the reports in XML format
   jacocoTestReport {
     reports {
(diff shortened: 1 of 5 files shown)


christopher-cudennec (commented):
Another option would be that developers need to set environment variable `GRADLE_OPTS`.
https://stackoverflow.com/questions/21267234/show-utf-8-text-properly-in-gradle

Since the issue only occurs on Windows machines, I'm not a big fan of touching all `build.gradle` files in all repos.

What do you think about the alternative?

JoergSiebahn:
I would say it is OS and project specific. Of course no one with a Windows computer should touch any other project but SDA projects. But in the end, you will have various projects on your computer, especially the externals working on windows. And `GRADLE_OPTS` or `gradle.settings` are global options.

christopher-cudennec (approved):

PR #1403, opened 2022-01-19
chore: split tests, Javadoc and Spotless

JoergSiebahn:


.github/workflows/javadoc.yml
```diff
@@ -0,0 +1,28 @@
+# This workflow builds the Javadoc
+name: Javadoc
```
christopher-cudennec, line 2: Don't we need to upload the generated artifacts? 
I guess they should also be published on Maven Central.

JoergSiebahn: We store these artifacts in Java CI:

```
            build/reports/
            */build/test-results/**/*.xml
```

None of these contain the generated Javadoc.
Is there actually anything that we use from the build for publication?

JoergSiebahn: At least the snapshot repo has Javadoc jars with content.

image

christopher-cudennec: You're right. We seem to build all artifacts for publication in that specific workflow.

christopher-cudennec (approved):

PR #1421, opened 2022-02-02
feat(server-hibernate-testing): add JUnit 5 support for DAOTestRule

ericrath:


sda-commons-server-hibernate-example/src/main/java/org/sdase/commons/server/hibernate/example/DAOClassExtension.java
```diff
@@ -0,0 +1,61 @@
...
+import org.hibernate.SessionFactory;
+import org.junit.jupiter.api.extension.AfterAllCallback;
+import org.junit.jupiter.api.extension.BeforeAllCallback;
+import org.junit.jupiter.api.extension.ExtensionContext;
+
+/*
+ * This class is very similar to io.dropwizard.testing.junit5.DAOTestExtension,
+ * but is based on org.junit.jupiter.api.extension.Extension which allows using @RegisterExtension
+ */
+public class DAOClassExtension implements BeforeAllCallback, AfterAllCallback {
```
JoergSiebahn, line 19: Should this be part of `hibernate-testing` so that consumers of sda-commons can use it in their service or is it intended to be an example here in the codebase only?

PR #1430, opened 2022-02-07
fix(server-openapi): Fix NPE in the log for lists of String

christopher-cudennec:
The error occurred under certain conditions when using the `Schema` annotation for a list of `String`. E.g.

```
  @Schema(description = "The list of all partners based on the query request.", required = true)
  private List<String> partners;
```


jeskosda (approved):
LGTM 👍🏻

PR #1452, opened 2022-02-18
feat: add health check for S3 connection

frederic-huebner:


sda-commons-server-s3/src/main/java/org/sdase/commons/server/s3/S3Bundle.java
```diff
@@ -113,6 +139,7 @@ public static InitialBuilder builder() {
   public static class Builder<T extends Configuration> implements InitialBuilder, FinalBuilder<T> {
 
     private final S3ConfigurationProvider<T> configProvider;
+    private boolean healthCheckDisabled = false;
```
JoergSiebahn, line 142: The commit message (and therefore the announced feature) mentions that an external health check is added.
For me it looks like a regular internal health check is introduced that can be turned into an external health check.
What is the impact on existing services when a new health check is introduced by default?

Is this a breaking change, @SDA-SE/sda-commons-maintainer ?
If not, I would say the change is good as is, but the commit message should be changed.
Otherwise, I would change the default to external.

sda-commons-server-s3/src/main/java/org/sdase/commons/server/s3/health/S3HealthCheck.java
```diff
@@ -0,0 +1,30 @@
...
+  private final AmazonS3 s3Client;
+
+  public S3HealthCheck(AmazonS3 s3Client) {
+    this.s3Client = s3Client;
+  }
+
+  @Override
+  protected Result check() {
+    try {
+      s3Client.listBuckets();
```
christopher-cudennec, line 23: Not sure if this can be a problem with regards to privileges. See: https://aws.amazon.com/de/premiumsupport/knowledge-center/s3-access-denied-listobjects-sync/

Is there any easy alternative that only access one single bucket? Do we need a new config option "bucket for healthcheck" in that case?

frederic-huebner: I actually wanted to avoid specifying a bucket with this, but if that's the case, I guess I can't avoid it.

christopher-cudennec: I'm not really sure if it's really a security issue. I'm just cautious because of the latest issues we had with Kafka ACLs.
What about making "healthCheckBucket" an option? We could either use `listBuckets` or make a query to `healthCheckBucket` if someone configured it.

sda-commons-server-s3/src/main/java/org/sdase/commons/server/s3/health/S3HealthCheck.java
```diff
@@ -0,0 +1,32 @@
...
+
+  public S3HealthCheck(AmazonS3 s3Client, String bucketName) {
+    this.s3Client = s3Client;
+    this.bucketName = bucketName;
+  }
+
+  @Override
+  protected Result check() {
+    try {
+      s3Client.doesBucketExistV2(bucketName);
```
christopher-cudennec, line 25: Should we evaluate the return value? I.e. return `unhealthy` if the bucket does not exist?

christopher-cudennec (commented):
General comment: We should add some documentation to the modules README.md.

sda-commons-server-s3/src/main/java/org/sdase/commons/server/s3/health/S3HealthCheck.java
```diff
@@ -0,0 +1,35 @@
...
+    this.s3Client = s3Client;
+    this.bucketName = bucketName;
+  }
+
+  @Override
+  protected Result check() {
+    try {
+      return s3Client.doesBucketExistV2(bucketName)
+          ? Result.healthy()
+          : Result.unhealthy(S3_CONNECTION_FAILURE_MESSAGE);
```
christopher-cudennec, line 29: Different message in this case?

sda-commons-server-s3/src/main/java/org/sdase/commons/server/s3/health/S3HealthCheck.java
```diff
@@ -0,0 +1,35 @@
...
+
+@Async(period = 30, scheduleType = Async.ScheduleType.FIXED_DELAY)
+public class S3HealthCheck extends HealthCheck {
+
+  private static final Logger LOG = LoggerFactory.getLogger(S3HealthCheck.class);
+  public static final String S3_CONNECTION_FAILURE_MESSAGE =
+      "Could not establish an S3 connection.";
+
+  private final AmazonS3 s3Client;
+  private final String bucketName;
```
christopher-cudennec, line 17: Do we want to support multiple buckets?

christopher-cudennec (commented):
Last comment regarding the commit message:
Could you add the module to the message, i.e. `feat(server-s3): ...`
Thanks! Cool feature! 😃

christopher-cudennec (approved):
🚂

PR #1465, opened 2022-02-25
Adapt interfacing Jaeger documentation

midu-git:


christopher-cudennec (approved):
After discussing with the other sda-commons maintainers, I removed the reference to JAEGER_ENDPOINT. Should be enough to take a look at the official Jaeger documentation that is linked above. Thanks for the change! 👍

PR #1469, opened 2022-02-25
V3 / Next major release

christopher-cudennec:

PR #1470, opened 2022-02-25
Morphia 2 WIP

christopher-cudennec:
Current issues:
* validation module only available in 2.3.0-SNAPSHOT, see https://github.com/MorphiaOrg/morphia/issues/1584
* can't add custom codecs (formerly known as converters), see https://github.com/MorphiaOrg/morphia/issues/1727


JoergSiebahn:
We will move from Morphia 1 to Spring Data Mongo instead of Morphia 2 in v3 (see #1469) - both alternatives would be a breaking change.

PR #1474, opened 2022-02-28
feat(server-cloudevents): Provide new bundle to handle CloudEvents

christopher-cudennec:


christopher-cudennec:
replaced by #1507

PR #1476, opened 2022-03-01
feat: defined cloudevents traits for asyncapi

vlad-geiger-msg-systems:


vlad-geiger-msg-systems:
After SDA internal platform alignment, the decision has been taken to use CloudEvents with structured mode(https://github.com/cloudevents/spec/blob/v1.0.2/cloudevents/bindings/kafka-protocol-binding.md#33-structured-content-mode). 

Comparison of the two modes can be found here: https://sda-se.atlassian.net/wiki/spaces/platform/pages/3197894710/CloudEvents+Content+Modes.

For this reason this feature is no longer required and I close this PR.

PR #1483, opened 2022-03-02
fix: fixed a bug where external urls could not be defined as refs within the async api

vlad-geiger-msg-systems:


sda-commons-shared-asyncapi/src/main/java/org/sdase/commons/shared/asyncapi/internal/JsonSchemaEmbedder.java
```diff
@@ -135,4 +138,8 @@ private JsonNode ensureParentNode(JsonNode target, JsonPointer pointer) {
   private String replaceSpecialCharacters(String value) {
     return value.replaceAll("[^a-zA-Z0-9]", "");
   }
+
+  private boolean shouldEmbed(String value) {
+    return !value.startsWith("https://");
```
christopher-cudennec, line 143: Is this the only case? Do we already know what kind of URLs we should ignore?

christopher-cudennec: There doesn't seem to be a test for the "https" case, right? Could you please add one?

vlad-geiger-msg-systems: there is one, please check https://github.com/SDA-SE/sda-dropwizard-commons/pull/1483/files#diff-a6cc53dae6615711add1a52b0ea763497936180c5e6538c440444c518b20af78R57 and https://github.com/SDA-SE/sda-dropwizard-commons/pull/1483/files#diff-8a06b4e84467bf2f3b7fe793b3eb83226f42663b016787c691a2f30be6f53e00R17

christopher-cudennec (commented):
My last comment before approving:
Could you squash your commits and use prefix `fix(shared-asyncapi): ...` please?

christopher-cudennec (approved):

PR #1485, opened 2022-03-03
feat(keymgmt): add new annotation to validate a value with a list of …

steinerda:
…platform keys

### DIFF
--- a/sda-commons-server-key-mgmt/README.md
+++ b/sda-commons-server-key-mgmt/README.md
@@ -60,6 +60,13 @@ The annotation `@PlatformKey` will trigger a validator to verify if the received
 private String genderKey
 ```
 
+### @PlatformKeys
+The annotation `@PlatformKeys` will trigger a validator to verify if the received value is a valid key within a list of referenced platform keys.
+```java
+@PlatformKeys(values = {"GENDER, SALUTATION"})
+private String genderOrSalutationKey
+```
+
 ### Key yaml file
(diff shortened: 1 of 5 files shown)


sda-commons-server-key-mgmt/src/main/java/org/sdase/commons/keymgmt/validator/PlatformKey.java
```diff
@@ -26,6 +26,8 @@
 @Retention(RUNTIME)
 @Constraint(validatedBy = KeyValidator.class)
 @Documented
+@Deprecated()
```
christopher-cudennec, line 29: Do we really want to deprecate the single key? I can imagine that you want to use exactly one key in many cases. Is it much work maintaining both annotations?

steinerda: no, it is ok for me. it's not extra work

christopher-cudennec:
With regards to the code duplication issue:
What about extending `KeysValidator` by `KeyValidator`?
If `keyNames` was accessible, `KeyValidator` could set `super.keyNames = new String[] { keyName }`. It would not need to implement `isValid` again.

christopher-cudennec (commented):
Looks good to me! ⭐ 
Anyone else wants to review? @SDA-SE/sda-commons-maintainer

christopher-cudennec (approved):

PR #1507, opened 2022-03-18
feat(server-cloudevents): Provide basic support for CloudEvents

christopher-cudennec:


sda-commons-server-cloudevents/src/main/java/org/sdase/commons/server/cloudevents/CloudEventV1.java
```diff
@@ -0,0 +1,181 @@
...
+   *     of Attributes in the Primer</a>
+   */
+  @JsonPropertyDescription(
+      "This attribute contains a value describing the type of event related "
+          + "to the originating occurrence. Often this attribute is used for routing, observability, "
+          + "policy enforcement, etc. The format of this is producer defined and might include "
+          + "information such as the version of the type - see Versioning of Attributes in the Primer "
+          + "for more information.")
+  @NotEmpty
+  private String type;
```
vlad-geiger-msg-systems, line 55: the "type" is for a specific event always the same. How can we provide documentation "hooks" in order to reflect this?

sda-commons-server-cloudevents/src/main/java/org/sdase/commons/server/cloudevents/CloudEventV1.java
```diff
@@ -0,0 +1,181 @@
...
+          + "In some event formats the `datacontenttype` attribute MAY be omitted. For example, if a "
+          + "JSON format event has no `datacontenttype` attribute, then it is implied that the data is "
+          + "a JSON value conforming to the \"application/json\" media type. In other words: a "
+          + "JSON-format event with no `datacontenttype` is exactly equivalent to one with "
+          + "`datacontenttype=\"application/json\"`.\n"
+          + "\n"
+          + "When translating an event message with no `datacontenttype` attribute to a different format "
+          + "or protocol binding, the target `datacontenttype` SHOULD be set explicitly to the implied "
+          + "`datacontenttype` of the source.")
+  private String datacontenttype;
```
vlad-geiger-msg-systems, line 94: At SDA we always use application/json. Maybe a good idea would be to enforce this by removing the "set" method and hard-code the value here (bind it with the setData method in case that the value should be null).

sda-commons-server-cloudevents/src/main/java/org/sdase/commons/server/cloudevents/CloudEventV1.java
```diff
@@ -0,0 +1,181 @@
...
+          + "\n"
+          + "An application MAY assign a unique `source` to each distinct producer, which makes it "
+          + "easy to produce unique IDs since no other producer will have the same source. The "
+          + "application MAY use UUIDs, URNs, DNS authorities or an application-specific scheme to "
+          + "create unique source identifiers.\n"
+          + "\n"
+          + "A source MAY include more than one producer. In that case the producers MUST collaborate "
+          + "to ensure that `source` + `id` is unique for each distinct event.")
+  @NotNull
+  private URI source;
```
vlad-geiger-msg-systems, line 41: the "source" is for a specific service always the same. How can we provide documentation "hooks" in order to reflect this?

sda-commons-server-cloudevents/src/main/java/org/sdase/commons/server/cloudevents/CloudEventV1.java
```diff
@@ -0,0 +1,181 @@
...
+import java.util.UUID;
+import javax.validation.constraints.NotEmpty;
+import javax.validation.constraints.NotNull;
+
+/**
+ * @param <T> the type of the data element of this CloudEvent
+ * @see <a href="https://github.com/cloudevents/spec/blob/v1.0.2/cloudevents/spec.md">1.0.2
+ *     Specification</a>
+ */
+public class CloudEventV1<T> extends BaseCloudEvent {
```
vlad-geiger-msg-systems, line 15: Maybe we can provide some @JsonSchemaExamples with @JsonSchemaFormat and mark the required attributes also within the documentation: @JsonProperty(required = true)

christopher-cudennec (commented):
@vlad-geiger-msg-systems @JoergSiebahn 

At the moment all CloudEvent properties also have a copy of the official documentation. Do you like that approach? Or do we only want to refer to the official spec and add the SDA specific documentation instead?

What would you prefer?

sda-commons-server-cloudevents/src/main/java/org/sdase/commons/server/cloudevents/CloudEventV1.java
christopher-cudennec: The required attributes are derived from the validation annotations `NotNull` / `NotEmpty`. Have a look at the generated example documentation that I just added to the PR.

sda-commons-server-cloudevents/src/main/java/org/sdase/commons/server/cloudevents/CloudEventV1.java
christopher-cudennec: I tried out overriding `getSource` and `getType` in the inheriting classes. Jackson will prefer the `JsonPropertyDescription` annotation of overridden methods. Have a look at the generated `asyncapi.yml` example in this PR.

sda-commons-server-cloudevents/src/main/java/org/sdase/commons/server/cloudevents/CloudEventV1.java
christopher-cudennec: I set the default value to `application/json` but kept the setters. If really someone wants a different value, it's still possible.

christopher-cudennec:
We decided to add SDA specific requirements / recommendations to the documentation. I also added some examples. You can take a look at the generated `asyncapi.yaml` and open it in [Async API Stdio](https://studio.asyncapi.com/).

christopher-cudennec:
Next (final?) open question: What do we choose as module name? Should it be a "server" or "shared" module like "sda-commons-shared-asyncapi"?

sda-commons-server-cloudevents/src/main/java/org/sdase/commons/server/cloudevents/CloudEventV1.java
JoergSiebahn: The only thing, I could imagine is something like `application/my.custom.spec+json`. I appreciate that this is still possible 👍

sda-commons-server-cloudevents/src/test/java/org/sdase/commons/server/cloudevents/app/consume/ContractCreatedEvent.java
```diff
@@ -0,0 +1,46 @@
+package org.sdase.commons.server.cloudevents.app.consume;
+
+import com.fasterxml.jackson.annotation.JsonPropertyDescription;
+import java.net.URI;
+import org.sdase.commons.server.cloudevents.CloudEventV1;
+
+public class ContractCreatedEvent extends CloudEventV1<ContractCreatedEvent.ContractCreated> {
+
+  @JsonPropertyDescription(
```
JoergSiebahn, line 9: You may use `@JsonSchemaDefault` here to keep the original description.

There is also `@JsonSchemaInject` which could be used to define a constant value:

```
@JsonSchemaInject(json = "{ \"const\": \"/SDA-SE/insurance-contract/insurance-contract-stack/insurance-contract-service\"}")
or 
@JsonSchemaInject(strings = @JsonSchemaString(path = "const", value = "/SDA-SE/insurance-contract/insurance-contract-stack/insurance-contract-service"))
```

By documentation, `@JsonSchemaInject` can also be used at type level (to avoid overriding), but that seems to be buggy or I don't know how to use it.

```
This at type level:
@JsonSchemaInject(strings = @JsonSchemaString(path = "/properties/source/const", value = "/SDA-SE/insurance-contract/insurance-contract-stack/insurance-contract-service"))

leads to this in the Api-Yaml:

  ContractCreatedEvent:
...
      ? ""
      : properties:
          source:
            const: "/SDA-SE/insurance-contract/insurance-contract-stack/insurance-contract-service"
...

And 
@JsonSchemaInject(strings = @JsonSchemaString(path = "properties/source/const", value = "/SDA-SE/insurance-contract/insurance-contract-stack/insurance-contract-service"))

is not affecting the result
```

Maybe worth to investigate further.

christopher-cudennec: I switched to `JsonSchemaDefault`. It's nice to also have the original description! 👍

eikenilsknopp (approved):

PR #1529, opened 2022-03-30
fix(deps): bump jackson-databind to 2.13.2.2

christopher-cudennec:
Choosing a completely different approach to finding version conflicts.

What's wrong with the built in `failOnVersionConflict`?
* it does not seem to detect conflicts between different configurations/scopes
* it detects conflicts that are handled later by using `enforcedPlatform`

My feeling is that `enforcedPlatform` and `failOnVersionConflict` do not really play well together. I tried to get some help in [Gradle's forum](https://discuss.gradle.org/t/enforcedplatform-not-working-struggling-with-failonversionconflict/42516/2) but got no answers.

The version of `failOnVersionConflict` that I wrote here will:
* collect all effective versions from all (relevant) configurations/scopes, i.e. the version that was determined after constraints of `enforcedPlatform` were applied
* fail if different versions are used

I plugged the task into the build cycle. It will break the build if a version conflict is detected.

**Update**

The same behaviour should be supported by Gradle out-of-the-box. The feature is called "consistent resolution" and is described [here](https://docs.gradle.org/current/userguide/resolution_strategy_tuning.html#sec:java_consistency).

Unfortunately the feature does not work when activated in our project. See [bug report at Gradle](https://github.com/gradle/gradle/issues/20345) and our [pull request](https://github.com/SDA-SE/sda-dropwizard-commons/pull/1535).

### DIFF
--- a/failOnVersionConflict.gradle
+++ b/failOnVersionConflict.gradle
@@ -1,0 +1,137 @@
+/**
+ * This task will parse all dependency files in 'build/deps' and tries to find version conflicts.
+ * Our assumption is: Version should be aligned by enforcing versions from our platform module
+ * 'sda-commons-dependencies'. We don't care if we find different versions as long as Gradle is
+ * able to choose the right version from 'sda-commons-dependencies'.
+ */
+task failOnVersionConflict {
+  doLast {
+    // Only these scopes will be considered
+    var relevantScopes = new HashSet<String>(
+        Arrays.asList("api",
+        "compileClasspath",
+        "compileOnly",
+        "compileOnlyApi",
+        "implementation",
+        "runtimeClasspath",
+        "testCompileClasspath",
+        "testCompileOnly",
+        "testImplementation",
+        "testRuntimeClasspath",
+        "testRuntimeOnly",
+        ))
+
+    // Add some dependencies that can be ignored
+    var ignoredDuplicates = new HashSet<String>(
+        Arrays.asList("io.openapitools.hal:swagger-hal")
+        )
+
+    var baseDir = new File("${rootProject.buildDir}/deps")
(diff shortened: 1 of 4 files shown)


sda-commons-dependencies/build.gradle
```diff
@@ -21,25 +21,47 @@ ext {
   dbRiderVersion = '1.32.3'
   kotlinVersion = '1.6.10'
   kotlinxCoroutinesVersion = '1.6.0'
+  jacksonBomVersion = '2.13.2.20220328'
   jacksonVersion = '2.13.2'
+  jacksonDatabindVersion = '2.13.2.2'
```
JoergSiebahn, line 26: Hope we can join this into a single property soon.

failOnVersionConflict.gradle
```diff
@@ -0,0 +1,137 @@
+/**
+ * This task will parse all dependency files in 'build/deps' and tries to find version conflicts.
+ * Our assumption is: Version should be aligned by enforcing versions from our platform module
+ * 'sda-commons-dependencies'. We don't care if we find different versions as long as Gradle is
+ * able to choose the right version from 'sda-commons-dependencies'.
+ */
+task failOnVersionConflict {
```
JoergSiebahn, line 7: Should this be added somewhere in the GH-Actions?

christopher-cudennec: The task will be executed by the following lines of code at the end of the file:

```
// Make sure we can find the dependency files for all subprojects
failOnVersionConflict.dependsOn(subprojects.allDeps)

// execute task with 'check'
check.dependsOn failOnVersionConflict 
```

JoergSiebahn (approved):

PR #1552, opened 2022-04-07
Organize dependencies

JoergSiebahn:


sda-commons-client-jersey-wiremock-testing/build.gradle
```diff
@@ -1,8 +1,24 @@
...
 
   api "com.github.tomakehurst:wiremock-jre8", {
+    // conflict with jetty from dropwizard
     exclude group: 'com.fasterxml.jackson', module: 'jackson-bom'
+    exclude group: 'org.eclipse.jetty', module: 'jetty-bom'
+    exclude group: 'org.eclipse.jetty', module: 'jetty-util-ajax'
+    exclude group: 'org.eclipse.jetty', module: 'jetty-continuation'
+    exclude group: 'org.eclipse.jetty', module: 'jetty-xml'
+    exclude group: 'org.eclipse.jetty', module: 'jetty-client'
+    exclude group: 'org.eclipse.jetty', module: 'jetty-alpn-client'
```
christopher-cudennec, line 11: Could you exclude the whole group?

christopher-cudennec (commented):
Personally I now think that this whole dependency management with Gradle is totally over-engineered (by Gradle). I would try to keep it simple and get rid of "failOnVersionConflict". Since it's still might only be my personal opinion and not mainstream at SDA I will not block further attempts to square the circle.

sda-commons-client-jersey-wiremock-testing/build.gradle
JoergSiebahn: We should [declare the conflicting dependency](https://github.com/SDA-SE/sda-dropwizard-commons/pull/1552/files#diff-9ad82bad7e55e667f6af9c74e51638e140a09c9fa97d079c027172cf8b9e1199R15-R21) anyways to ensure that it uses the managed version. Looks a bit weird to me to exclude brute force and then select specific modules.

christopher-cudennec (approved):

PR #1569, opened 2022-04-19
chore: Upgrade to latest Fossa version

christopher-cudennec:
Updates to latest version of Fossa.

* Fossa analysis still works: https://app.fossa.com/projects/custom%2b8463%2fsda-dropwizard-commons/refs/branch/refs%2fheads%2ffeat%2ffossa-latest/24c460e769beb76ffd4eeb16ea3472fd1185be89 ✅ 
* it also refers to the correct branch! ✅ 
* it also refers to the correct commit SHA ✅ 

Open questions:
* do we really want to use "latest"? we could pin to exact versions instead: `https://github.com/fossas/fossa-cli/blob/v3.2.8/install-latest.sh`


eikenilsknopp:
> do we really want to use "latest"?

I would prefer to pin it to an exact version.

eikenilsknopp (approved):

PR #1581, opened 2022-04-26
Enable RSA and ECC with multiple signing alg

jeskosda:
This PR should try out to easily add EC as supported alg for token validation without big refactoring of existing code.

A breaking change would be to not accept empty alg in the jwks, because this needs to be distinguished now. 

I also need to add BouncyCastle if i want to easily resolve the Alg from the PEM file, at least i found no better solution. 
I we don't want to have the dependency, then we could still allow only RS265 for the PEM files. Are PEM files used somewhere? ALH? SI? Wondering if that is still a desired feature?


sda-commons-server-auth/src/main/java/org/sdase/commons/server/auth/service/AuthService.java
```diff
@@ -1,17 +1,117 @@
...
+    }
+  }
+
+  private Optional<DecodedJWT> verifyJwtSignature(
+      String authorizationToken, LoadedPublicKey loadedPublicKey) {
+    try {
+
+      Verification jwtVerification =
+          JWT.require(resolveAlgorithm(loadedPublicKey)).acceptLeeway(leeway);
+
```
jeskosda, line 101: This is just the new part.

sda-commons-server-auth/src/main/java/org/sdase/commons/server/auth/key/JwksKeySource.java
```diff
@@ -156,6 +204,9 @@ private static class Key {
     private String use;
     private String n;
     private String e;
+    private String x;
+    private String y;
+    private String crv;
```
christopher-cudennec, line 209: Are these common/good names for EC?

christopher-cudennec (commented):
Does your PR still contain a breaking change? Could you point me to the class/line of code?
Have you tried the PR in another application?

sda-commons-server-auth/src/main/java/org/sdase/commons/server/auth/key/JwksKeySource.java
jeskosda: https://datatracker.ietf.org/doc/html/rfc7517#section-3 that's how it's provided

sda-commons-server-auth/src/main/java/org/sdase/commons/server/auth/key/JwksKeySource.java
```diff
@@ -97,39 +111,73 @@ private boolean isForSigning(Key key) {
...
+    return SUPPORTED_KTY.contains(key.getKty());
+  }
+
+  private boolean isSupportedAlg(Key key) {
+    return SUPPORTED_ALG.contains(key.getAlg());
   }
 
-  private boolean isRsa256Key(Key key) {
-    // We only support RSA256, if blank we assume it to be RS256.
-    return StringUtils.isBlank(key.getAlg()) || "RS256".equals(key.getAlg());
```
jeskosda, line 106: The breaking change is actually here. So far we assume if the `"alg": "" ` is provided in the JWK, the RSA256 is used as default. Which is in my opinion a bad behavior. The "alg" should be "none" (also not supported so far) or have an existing alg. 

We would throw an exception if none of the supported algs couldnt be determined.

jeskosda: We could still fallback to `RSA256`, but this feels not right in my opinion. The parameter is OPTIONAL in the specification: https://datatracker.ietf.org/doc/html/rfc7517#section-4.4, which means for me that we are using a not signed JWT so not a JWS here. Which is the equivalent of "none". Which could be an proper default value then. 

See the list of algs here: https://www.iana.org/assignments/jose/jose.xhtml

jeskosda:
Regarding test: It could be a bit more difficult here. First of all we need to provide JWK with EC and provide a token which is signed with ES256. But i will investigate if our keycloak can do that 👍🏻

sda-commons-server-auth/src/main/java/org/sdase/commons/server/auth/key/JwksKeySource.java
JoergSiebahn: For security reasons we should not support `alg: "none"`.

christopher-cudennec: We do fix a non-compliant implementation by this "breaking change", don't we? We could argue that this is not a breaking change then. What do you think?

jeskosda:
@christopher-cudennec  I tested the SNAPSHOT against our keycloak with a token using signing alg ES256. Works fine 👍🏻

christopher-cudennec (approved):

PR #1610, opened 2022-05-13
feat: add support for opentelemetry

saadmar:
- [ ] Deploy a collector in the configured endpoint
- [ ] Test in one of the stacks


sda-commons-shared-otel-agent/README.md
```diff
@@ -0,0 +1,6 @@
+# SDA Commons Shared OpenTelemetry agent
```
christopher-cudennec, line 1: Why did you choose "shared" over "server" as part of the module name?

sda-commons-shared-otel-agent/README.md
```diff
@@ -0,0 +1,6 @@
...
+
+This provides an extension of the [OpenTelemetry Java agent](https://github.com/open-telemetry/opentelemetry-java-instrumentation) that redefines the default configuration that are set in the base agent to match requirements of Sda services.
+All configuration that can be used can be found in [OpenTelemetry SDK Autoconfigure](https://github.com/open-telemetry/opentelemetry-java/tree/main/sdk-extensions/autoconfigure). 
+
+> **_NOTE:_** This module _must_ be built as a fat jar.
```
christopher-cudennec, line 6: What does that mean for users?

sda-commons-shared-otel-agent/build.gradle
```diff
@@ -0,0 +1,57 @@
+plugins{
+  id "com.github.johnrengelman.shadow" version "7.1.2"
+}
+ext {
+  otelAgentVersion = '1.13.1'
+  otelAgentAlphaVersion = '1.13.1-alpha'
```
christopher-cudennec, line 6: * does the version need to match the open telemetry version defined in the dependencies module?
* I would also move these versions to the dependencies module

sda-commons-shared-otel-agent/build.gradle
```diff
@@ -0,0 +1,57 @@
...
+
+  autoservice= dependencies.create(group: 'com.google.auto.service', name: 'auto-service', version: '1.0')
```
christopher-cudennec, line 8: why do you need to create a dependency programmatically?

sda-commons-shared-otel-agent/build.gradle
```diff
@@ -0,0 +1,57 @@
...
+}
+
+configurations {
+  agents
+  // Interfaces that are used to compile the SPI are also needed for testing
+  testImplementation.extendsFrom compileOnly
```
christopher-cudennec, line 14: strange...

sda-commons-shared-otel-agent/src/main/java/org/sdase/commons/shared/otel/agent/SdaConfigPropertyProvider.java
```diff
@@ -0,0 +1,38 @@
...
+@AutoService(ConfigPropertySource.class)
+public final class SdaConfigPropertyProvider implements ConfigPropertySource {
+  private static final String JAEGER_SERVICE_NAME_ENV_VAR = "JAEGER_SERVICE_NAME";
+
+  @Override
+  public Map<String, String> getProperties() {
+    Map<String, String> properties = new HashMap<>();
+
+    // default otel collector gateway endpoint 4317 used for grpc
+    properties.put("otel.exporter.otlp.endpoint", "http://otel-collector-gateway.jaeger:4317");
```
christopher-cudennec, line 23: How does work for our customers or in different environments?

sda-commons-shared-otel-agent/src/main/java/org/sdase/commons/shared/otel/agent/SdaConfigPropertyProvider.java
```diff
@@ -0,0 +1,38 @@
...
+
+    // 'jaeger' header formats are used by the deprecated libraries.
+    // 'tracecontext' 'baggage' formats are recommended defaults and used by most external systems.
+    properties.put("otel.propagators", "tracecontext,baggage,jaeger");
+
+    // use environment legacy env variable defined by jaeger client by default
+    String jaegerServiceName = System.getenv(JAEGER_SERVICE_NAME_ENV_VAR);
+    if (jaegerServiceName != null && !jaegerServiceName.isEmpty()) {
+      properties.put("otel.service.name", jaegerServiceName);
+    }
```
christopher-cudennec, line 35: How can we set the `otel.service.name` without using the legacy env variable? Do we need a new one, e.g. `OTEL_SERVICE_NAME`?

sda-commons-shared-otel-instrumentation/README.md
```diff
@@ -0,0 +1,82 @@
...
+
+  @Override
+  public void initialize(Bootstrap<Configuration> bootstrap) {
+    // ... other bundles  
+    bootstrap.addBundle(InstrumentationBundle.builder().build());
+  }
+}
+```
+
+If the application is already using the [starter bundle](../sda-commons-server-starter), no changes are needed.
```
christopher-cudennec, line 35: The bundle you're referencing is deprecated. Please refer to [sda-commons-starter](https://github.com/SDA-SE/sda-dropwizard-commons/tree/master/sda-commons-starter).

sda-commons-shared-otel-agent/src/main/java/org/sdase/commons/shared/otel/agent/SdaConfigPropertyProvider.java
christopher-cudennec: Ah, I found documentation for that env variable in https://github.com/open-telemetry/opentelemetry-java/tree/main/sdk-extensions/autoconfigure#exporters

sda-commons-shared-otel-instrumentation/README.md
```diff
@@ -0,0 +1,82 @@
...
+
+## Service Name
+
+The service name is used to identify the source (the sender) of the received telemetry data in the monitoring backend. Therefore it's important that every service has a unique name.
+To provide the application with a custom name for each deployment, the `OTEL_SERVICE_NAME` environment variable must be used.
+
+## Basic configuration
+
+By default, the agent exports traces in the otlp format, to an [openTelemetry collector](https://opentelemetry.io/docs/collector/). The most basic configuration would be to provide the service name and a collector endpoint.
+- `OTEL_SERVICE_NAME=my-service`
```
christopher-cudennec, line 57: The JaegerBundle used the application name as default service name. What happens here if you don't specify OTEL_SERVICE_NAME?

sda-commons-shared-otel-instrumentation/README.md
```diff
@@ -0,0 +1,82 @@
...
+- `OTEL_EXPORTER_OTLP_ENDPOINT=http://otel-gateway-host:4317`.
+
+Or in case if the collector is deployed as a sidecar:
+- `OTEL_EXPORTER_OTLP_ENDPOINT=http://localhost:4317`
+
+## Using Jaeger exporter
+
+To exports traces in Jaeger format, to a [Jaeger collector](https://hub.docker.com/r/jaegertracing/jaeger-collector/). This can be very handy for less local setup overhead, where the [All in one Jaeger image](https://www.jaegertracing.io/docs/1.6/getting-started/#all-in-one-docker-image) can be enough.
```
christopher-cudennec, line 65: NIT "To export traces"

sda-commons-shared-otel-instrumentation/README.md
```diff
@@ -0,0 +1,82 @@
...
+- `OTEL_SERVICE_NAME=my-service`
+- `OTEL_EXPORTER_JAEGER_ENDPOINT=http://jaeger-collector-host:14250`
+
+> **_NOTE:_** It is recommended to not use this option in production even if the used monitoring backend is Jaeger. It is better to export the telemetry data in the _OTLP_ format to a collector, and configure it to handle the transformation.
+
+# Disabling instrumentation
+
+In order to disable tracing in the applications that are using this bundle, or the [starter bundle](../sda-commons-server-starter), the environment variable `OTEL_JAVAAGENT_ENABLED=false` can be used.
```
christopher-cudennec, line 73: Please refer to `sda-commons-starter`.

sda-commons-shared-otel-instrumentation/README.md
```diff
@@ -0,0 +1,82 @@
...
+Please note that setting `OTEL_JAVAAGENT_ENABLED` to false will force the bundle to completely skip loading the agent Jar into the JVM, so manual instrumentation like added in the [sda-commons-shared-otel-tracing-example](../sda-commons-shared-otel-tracing-example) will not be possible anymore.
+In this case manual tracing must be done differently (see other options for [manual instrumentation](https://opentelemetry.io/docs/instrumentation/java/manual/)).
+
+# Enable manual instrumentation only
+
+In order to suppress all auto instrumentation but still want to load the agent the option `OTEL_INSTRUMENTATION_COMMON_DEFAULT_ENABLED=false`.
```
christopher-cudennec, line 79: What could be a valid  scenario for manual instrumentation? When should we recommend it? If this is not desired, we should not mention it IMHO.

sda-commons-shared-otel-instrumentation/build.gradle
```diff
@@ -0,0 +1,17 @@
+ext {
+  otelContribVersion= '1.13.0'
+  otelContribAlphaVersion = '1.13.0-alpha'
```
christopher-cudennec, line 3: Should be moved to the dependency module to reduce duplication.

sda-commons-shared-otel-instrumentation/src/main/java/org/sdase/commons/shared/instrumentation/AgentLoader.java
```diff
@@ -0,0 +1,36 @@
...
+public class AgentLoader {
+  private AgentLoader() {
+    // utility class
+  }
+
+  private static final Logger LOG = LoggerFactory.getLogger(AgentLoader.class);
+
+  public static void load() {
+    String jaegerSamplerType = System.getenv("JAEGER_SAMPLER_TYPE");
+    String jaegerSamplerParam = System.getenv("JAEGER_SAMPLER_PARAM");
```
christopher-cudennec, line 20: You should use `new SystemPropertyAndEnvironmentLookup().lookup(...)` instead to make testing / configuration a little easier. It's not possible to replace env variables in tests starting from Java 16.

sda-commons-shared-otel-instrumentation/src/main/java/org/sdase/commons/shared/instrumentation/AgentLoader.java
```diff
@@ -0,0 +1,36 @@
...
+
+    // Skip loading the agent if tracing is disabled.
+    if ("const".equals(jaegerSamplerType) && "0".equals(jaegerSamplerParam)) {
+      LOG.warn("Tracing is disabled using deprecated configuration.");
+      LOG.info("Skipping instrumentation...");
+      return;
+    }
```
christopher-cudennec, line 27: Is `OTEL_JAVAAGENT_ENABLED` handled automatically?

sda-commons-shared-otel-tracing-example/README.md
```diff
@@ -0,0 +1,21 @@
+# SDA Commons Shared OpenTelemetry tracing Example
+
+This module is an example for a service with manual instrumentation using [OpenTelemetry](https://opentelemetry.io/).
+
+The usage of the modules [sda-commons-server-opentracing](../sda-commons-server-opentracing/README.md) and [sda-commons-shared-otel-agent](../sda-commons-shared-otel-agent/README.md) for custom manual instrumentation.
```
christopher-cudennec, line 5: `sda-commons-server-opentracing` was deprecated. Change documentation?

sda-commons-shared-otel-tracing-example/build.gradle
```diff
@@ -0,0 +1,15 @@
+dependencies {
+
+  api ("io.opentelemetry:opentelemetry-api")
+  // provides some conventional static attributes that are recommended for manual instrumentation.
+  api 'io.opentelemetry:opentelemetry-semconv:1.14.0-alpha'
```
christopher-cudennec, line 5: Extract version, please

sda-commons-shared-otel-tracing-example/src/test/java/org/sdase/shared/otel/tracing/example/OtelTracingApplicationIT.java
```diff
@@ -0,0 +1,92 @@
...
+  static final OpenTelemetryExtension otelTesting = OpenTelemetryExtension.create();
+
+  @BeforeEach
+  void setUp() throws IllegalAccessException {
+    // use InMemory Exporter for testing
+    FieldUtils.writeField(
+        APP.getApplication(),
+        "tracer",
+        otelTesting.getOpenTelemetry().getTracer(OtelTracingApplicationIT.class.getName()),
+        true);
```
christopher-cudennec, line 36: Can you find a better solution that does not use reflection to be ready for Java 17? What about providing a simple setter?

sda-commons-shared-otel-agent/README.md
saadmar: For the agent, it is just built as a standalone dependency that can be used for both servers and clients, if we consider also that some of the instrumented libraries are client dependencies.

sda-commons-shared-otel-agent/README.md
saadmar: You are right. that is only relevant to devs. I will add notes for users.

sda-commons-shared-otel-agent/build.gradle
saadmar: They are supposed to converge eventually, but can be different. The agent has the opentelemetry-api as a dependency and sometimes  takes a while to be updated.
I will move them.

sda-commons-shared-otel-agent/build.gradle
saadmar: True. At first most of the dependencies were only needed at compile time, this one was only added to supports tests as well, but that can be avoided. I will simplify it.

sda-commons-shared-otel-instrumentation/src/main/java/org/sdase/commons/shared/instrumentation/AgentLoader.java
saadmar: I changed it for this case, but for the agent module, maybe we can skip adding the dependency, and anyways internally it is using `System.getenv`, also replace env variables in tests is not an option, because the agent does most of the configuration at startup, and updates won't be picked up.

sda-commons-shared-otel-instrumentation/README.md
saadmar: the default will be `unknown_service:java`. not sure if this can be changed to keep the current behaviour, since it is set in a separate process.

sda-commons-shared-otel-agent/src/main/java/org/sdase/commons/shared/otel/agent/SdaConfigPropertyProvider.java
saadmar: This is just to define a default (still needs to be confirmed). This can be changed with the environment variable `OTEL_EXPORTER_OTLP_ENDPOINT` (documented in another module). We can add more documentation for deployment patterns... figured that this won't be the best place for  such docs.

sda-commons-shared-otel-instrumentation/README.md
saadmar: I also say that this is a very rare case. There would be many reasons for adding manual instrumentation (adding more business attributes, tracing some internal behaviour...), but right now I can not think of any for disabling automatic instrumentation. This was merely a continuation for the previous notes about disabling the agent and how to overcome such limitation
>In this case manual tracing must be done differently (see other options for [manual instrumentation](https://opentelemetry.io/docs/instrumentation/java/manual/)).

If such a case arises, this would describe an easier way to achieve this with the modules that we already have, without having to look for the other options.

sda-commons-shared-otel-agent/README.md
saadmar: added a reference to the other module, which describes steps to use.

JoergSiebahn:
Not needed any more, instrumentation this way is too slow on startup.

PR #1641, opened 2022-06-02
Use more restrictive CSP by default

JoergSiebahn:
Proposed release notes for these changes:

> If a service is serving HTML pages besides the API, this release requires enabling `withFrontendSupport()` when configuring the `SdaPlatformBundle` or the `SecurityBundle`. Otherwise, the frontend will not be able to use scripts, styles, images or web fonts.
> If a service is only serving APIs, no changes are needed and the service is more secure due to more restrictive CSP header.

### DIFF
--- a/sda-commons-server-security/src/main/java/org/sdase/commons/server/security/SecurityBundle.java
+++ b/sda-commons-server-security/src/main/java/org/sdase/commons/server/security/SecurityBundle.java
@@ -5,12 +5,15 @@ import io.dropwizard.ConfiguredBundle;
 import io.dropwizard.server.ServerFactory;
 import io.dropwizard.setup.Bootstrap;
 import io.dropwizard.setup.Environment;
-import org.sdase.commons.server.security.filter.WebSecurityHeaderFilter;
+import org.sdase.commons.server.security.filter.WebSecurityApiOnlyHeaderFilter;
+import org.sdase.commons.server.security.filter.WebSecurityFrontendSupportHeaderFilter;
 import org.sdase.commons.server.security.handler.ObscuringErrorHandler;
 import org.sdase.commons.server.security.validation.BufferLimitsAdvice;
 import org.sdase.commons.server.security.validation.CustomErrorHandlerSecurityAdvice;
 import org.sdase.commons.server.security.validation.HttpConnectorSecurityAdvice;
 import org.sdase.commons.server.security.validation.ServerFactorySecurityAdvice;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 /**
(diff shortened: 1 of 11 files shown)


sda-commons-server-security/src/main/java/org/sdase/commons/server/security/SecurityBundle.java
```diff
@@ -60,7 +68,14 @@ public void run(T configuration, Environment environment) {
 
     environment.getApplicationContext().setErrorHandler(createNewErrorHandler(environment));
     environment.getAdminContext().setErrorHandler(createNewErrorHandler(environment));
-    environment.jersey().register(WebSecurityHeaderFilter.class);
+    if (supportFrontend) {
+      LOG.info(
+          "Content-Security-Policy headers are configured to support frontends. "
+              + "Services that only serve APIs not enable this feature.");
```
christopher-cudennec, line 74: NIT: "Services ... don't enable this feature."

sda-commons-starter/src/main/java/org/sdase/commons/starter/builder/SecurityCustomizer.java
```diff
@@ -12,4 +12,14 @@
...
+
+  /**
+   * If a service is configured with frontend support, the {@code Content-Security-Policy} header
+   * allows the same domain as source for scripts, images, styles and fonts. Otherwise, only API
+   * endpoints can be served and {@code Content-Security-Policy} does not allow any sources.
+   *
+   * @return this builder instance
+   * @see <a href="https://en.wikipedia.org/wiki/Content_Security_Policy">CSP header</a>
+   */
+  PlatformBundleBuilder<C> withFrontendSupport();
```
christopher-cudennec, line 24: It would not be possible to use both options now, would it? Both methods return `PlatformBundleBuilder` and not `SecurityCustomizer`.

JoergSiebahn: [`PlatformBundleBuilder`](https://github.com/SDA-SE/sda-dropwizard-commons/blob/master/sda-commons-starter/src/main/java/org/sdase/commons/starter/builder/PlatformBundleBuilder.java#L13) extends `SecurityCustomizer`.

sda-commons-server-security/src/main/java/org/sdase/commons/server/security/SecurityBundle.java
JoergSiebahn: fixed

christopher-cudennec (approved):

PR #1643, opened 2022-06-03
feat(server-spring-data-mongo): New bundle for Spring Data Mongo

christopher-cudennec:
Adds spring-data-mongo as bundle

It should replace the morphia bundle since
* it's not compatible yet with Java 17
* it seems to rely on one single dev
* version 2.x introduced several breaking changes and removed important features


sda-commons-server-spring-data-mongo/src/main/java/org/sdase/commons/server/spring/data/mongo/SpringDataMongoBundle.java
```diff
@@ -0,0 +1,289 @@
...
+  private void registerHealthCheck(
+      HealthCheckRegistry healthCheckRegistry, MongoClient mongoClient, String database) {
+    healthCheckRegistry.register(
+        "mongo", new MongoHealthCheck(mongoClient.getDatabase((database))));
+  }
+
+  /**
+   * @return the {@link MongoOperations} that can be used to interact with your MongoDB instance
+   */
+  public MongoOperations getMongoOperations() {
```
JoergSiebahn, line 132: I would ensure, that we do not initialize the `mongoOperations` twice:

```suggestion
  public synchronized MongoOperations getMongoOperations() {
```

sda-commons-server-spring-data-mongo/src/main/java/org/sdase/commons/server/spring/data/mongo/SpringDataMongoBundle.java
```diff
@@ -0,0 +1,289 @@
...
+
+  private SpringDataMongoBundle<C> addCustomConverter(Converter<?, ?> converter) {
+    this.customConverters.add(converter);
+    return this;
+  }
+
+  private SpringDataMongoBundle<C> addCustomConverters(Converter<?, ?>... converters) {
+    Arrays.stream(converters).forEach(this::addCustomConverter);
+    return this;
+  }
```
JoergSiebahn, line 173: This private method is not used.

```suggestion
```

sda-commons-server-spring-data-mongo/src/main/java/org/sdase/commons/server/spring/data/mongo/SpringDataMongoBundle.java
```diff
@@ -0,0 +1,289 @@
...
+
+  private SpringDataMongoBundle<C> addCustomConverters(Collection<Converter<?, ?>> converters) {
+    converters.forEach(this::addCustomConverter);
+    return this;
+  }
```
JoergSiebahn, line 178: Imho we can simplify this:

```suggestion
  private SpringDataMongoBundle<C> addCustomConverters(Collection<Converter<?, ?>> converters) {
    this.customConverters.addAll(converters);
    return this;
  }
```

sda-commons-server-spring-data-mongo/src/main/java/org/sdase/commons/server/spring/data/mongo/SpringDataMongoBundle.java
christopher-cudennec: Not yet. The MorphiaBundle provided several methods for adding converters. Would be fine for me to clean them up a little bit.

sda-commons-server-spring-data-mongo/README.md
```diff
@@ -0,0 +1,121 @@
...
+
+```yaml
+mongo:
+  connectionString: "${MONGODB_CONNECTION_STRING:-}"
+  hosts: "${MONGODB_HOSTS:-}"
+  database: "${MONGODB_DATABASE:-}"
+  options: "${MONGODB_OPTIONS:-}"
+  username: "${MONGODB_USERNAME:-}"
+  password: "${MONGODB_PASSWORD:-}"
+  useSsl: ${MONGODB_USE_SSL:-true}
```
JoergSiebahn, line 83: Should we still support this kind of configuration?

Afaik every tool that accesses a MongoDB is configured with a connection string which is a well documented format of MongoDB. It would be a good opportunity to get rid of this custom configuration approach in v3.

christopher-cudennec: wasn't sure how many breaking changes we want to introduce / how easy it should be to migrate to the new bundle.

sda-commons-server-spring-data-mongo/src/main/java/org/sdase/commons/server/spring/data/mongo/SpringDataMongoBundle.java
JoergSiebahn: Imho that affects [the builder](https://github.com/SDA-SE/sda-dropwizard-commons/pull/1643/files#diff-f135cd5daaec8bded0717a881e3bf904904e78b8b00344e16c17688d017bb415R224-R235) only. At this point in the bundle there will be only one option.

sda-commons-server-spring-data-mongo/src/main/java/org/sdase/commons/server/spring/data/mongo/SpringDataMongoBundle.java
```diff
@@ -0,0 +1,289 @@
...
+  private static final Set<Converter<?, ?>> DEFAULT_CONVERTERS =
+      new LinkedHashSet<>(
+          List.of(
+              // LocalDate
+              new LocalDateToStringConverter(),
+              new StringToLocalDateConverter(),
+              // ZonedDateTime
+              new DateToZonedDateTimeConverter(),
+              new StringToZonedDateTimeConverter(),
+              new ZonedDateTimeToDateConverter()));
```
JoergSiebahn, line 60: Should we introduce `enableMorphiaCompatibility()` in the builder so that we can use it for upgrades from v2 to v3 but use the Spring Data defaults for new services? Imho that would pay out in the long term.

JoergSiebahn:
I did not see a test that ensures compatibility with the data structure Morphia produces in the various flavors (e.g. `noClassNameStored=true/false`, handling of generic subtypes). Imho we should add such a test based on known documents from existing services. I know there are some tricky variants in some of the consumers of v2.

sda-commons-server-spring-data-mongo/src/main/java/org/sdase/commons/server/spring/data/mongo/SpringDataMongoBundle.java
christopher-cudennec: Better safe than sorry. Although I don't expect the method to be invoked many times. It should be used on running your application when you initialize your services. Application's will usually only have one instance.

sda-commons-server-spring-data-mongo/src/main/java/org/sdase/commons/server/spring/data/mongo/SpringDataMongoBundle.java
christopher-cudennec: Feel free to change it 👍

sda-commons-server-spring-data-mongo/src/main/java/org/sdase/commons/server/spring/data/mongo/SpringDataMongoBundle.java
christopher-cudennec: Feel free to change it 👍

christopher-cudennec:
> I did not see a test that ensures compatibility with the data structure Morphia produces in the various flavors (e.g. `noClassNameStored=true/false`, handling of generic subtypes). Imho we should add such a test based on known documents from existing services. I know there are some tricky variants in some of the consumers of v2.

Can you create one or more subtasks for that? I haven't come so far with my changes. 😃 
It's really crucial for service migrations that the two bundles interoperate well together.

preethivenkatesh-sdase:
https://github.com/SDA-SE/sda-dropwizard-commons/blob/feat/spring-data-mongo/sda-commons-server-morphia-example/src/test/resources/test-config.yml

Do we need this file?

elbiocaetano:
> https://github.com/SDA-SE/sda-dropwizard-commons/blob/feat/spring-data-mongo/sda-commons-server-morphia-example/src/test/resources/test-config.yml

No, we do not, I will remove it

elbiocaetano (approved):

PR #1659, opened 2022-06-15
fix(deps): downgraded okhttp to 4.9.3 to fix CVE-2021-0341

vlad-geiger-msg-systems:


christopher-cudennec:
Going through the comments of https://github.com/square/okhttp/issues/6724 it seems like this is a false positive.
They created a ticket to fix the CVE: https://github.com/OSSIndex/vulns/issues/296

vlad-geiger-msg-systems:
vulnerability is actually fixed and its a false positive, therefore this PR is not required

PR #1680, opened 2022-06-24
Document the current state of fault tolerance in tests …

JoergSiebahn:
… for server and client implementations using the configured default `ObjectMapper`.

### DIFF
--- a/sda-commons-client-jersey/src/test/java/org/sdase/commons/client/jersey/test/ClientTestApp.java
+++ b/sda-commons-client-jersey/src/test/java/org/sdase/commons/client/jersey/test/ClientTestApp.java
@@ -1,6 +1,5 @@
 package org.sdase.commons.client.jersey.test;
 
-import com.fasterxml.jackson.databind.DeserializationFeature;
 import io.dropwizard.Application;
 import io.dropwizard.forms.MultiPartBundle;
 import io.dropwizard.setup.Bootstrap;
@@ -29,7 +28,6 @@ public class ClientTestApp extends Application<ClientTestConfig> {
 
   @Override
   public void initialize(Bootstrap<ClientTestConfig> bootstrap) {
-    bootstrap.getObjectMapper().disable(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES);
     bootstrap.addBundle(ConfigurationSubstitutionBundle.builder().build());
     bootstrap.addBundle(JacksonConfigurationBundle.builder().build());
     bootstrap.addBundle(TraceTokenBundle.builder().build());
--- a/sda-commons-client-jersey/src/test/java/org/sdase/commons/client/jersey/test/ClientTestApp.java
+++ b/sda-commons-client-jersey/src/test/java/org/sdase/commons/client/jersey/test/ClientTestApp.java
@@ -9,6 +9,7 @@ import io.opentracing.mock.MockTracer;
 import org.sdase.commons.client.jersey.JerseyClientBundle;
 import org.sdase.commons.client.jersey.oidc.OidcClient;
 import org.sdase.commons.server.dropwizard.bundles.ConfigurationSubstitutionBundle;
+import org.sdase.commons.server.jackson.JacksonConfigurationBundle;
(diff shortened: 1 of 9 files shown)


sda-commons-client-jersey/src/test/java/org/sdase/commons/client/jersey/test/ClientTestApp.java
```diff
@@ -30,6 +31,7 @@ public static void main(String[] args) throws Exception {
   public void initialize(Bootstrap<ClientTestConfig> bootstrap) {
     bootstrap.getObjectMapper().disable(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES);
```
christopher-cudennec, line 32: Do you know why the test application uses custom configuration for the `ObjectMapper`?

sda-commons-server-jackson/src/test/java/org/sdase/commons/server/jackson/JacksonConfigurationBundleIT.java
```diff
@@ -597,4 +597,80 @@ public void shouldGetErrorForMissingQueryParameter() {
...
+
+  @Test
+  void shouldNotAcceptSubtypeUndefinedSubtypeThree() {
+    try (Response response =
+        DW.client()
+            .target("http://localhost:" + DW.getLocalPort())
+            .path("subtypes")
+            .request(MediaType.APPLICATION_JSON)
+            .post(Entity.json(Collections.singletonMap("type", "THREE")))) {
+      assertThat(response.getStatus()).isEqualTo(422);
```
christopher-cudennec, line 635: That's also my expectation! 👍

sda-commons-server-jackson/src/test/java/org/sdase/commons/server/jackson/test/JacksonConfigurationTestApp.java
```diff
@@ -195,4 +196,22 @@ public Response createRequiredQueryValidationResource(
...
+  @Consumes(MediaType.APPLICATION_JSON)
+  public Response createSubType(@Valid @NotNull AbstractResource requestBody) {
+    return Response.created(URI.create("/subtypes/" + requestBody.getType())).build();
+  }
+
+  @POST
+  @Path("/subtypesTolerant")
+  @Consumes(MediaType.APPLICATION_JSON)
+  public Response createSubTypeAllowNull(@Valid AbstractResource requestBody) {
+    if (requestBody == null) {
```
christopher-cudennec, line 211: I find that a bit speculative to assume that the reasons was that the subtype is not supported. It's unfortunate, that we can't show a proper error message, e.g. `subtype 'three' is not invalid`. We should probably combine `@Valid` and `@NotNull` here, shouldn't we?

sda-commons-client-jersey/src/test/java/org/sdase/commons/client/jersey/test/ClientTestApp.java
JoergSiebahn: That's the OM for reading the config. Imho, we reuse the config for different test apps with different config classes and therefore have properties in the yaml that are not configurable in every test app. But actually, it seems to be the other app that's needs the tolerant reader. Should I try to remove the extra setting in this PR?

christopher-cudennec: Yeah. Let's see if the extra configuration is needed at all! 👍

christopher-cudennec (approved):

sda-commons-client-jersey/src/test/java/org/sdase/commons/client/jersey/test/ClientTestApp.java
JoergSiebahn: Removed: https://github.com/SDA-SE/sda-dropwizard-commons/pull/1680/commits/016003ee131237b4111d4972ca29f90b8f68d795

christopher-cudennec (approved):

PR #1835, opened 2022-09-27
feat(server-morphia): Support `connectionString` in configuration

christopher-cudennec:


sda-commons-server-morphia/README.md
```diff
@@ -91,6 +91,7 @@ The database connection is configured in the `config.yaml` of the application.
 Example config for **production** to be used with environment variables of the cluster configuration:
 ```yaml
 mongo:
+  connectionString: "${MONGODB_CONNECTION_STRING}"
```
JoergSiebahn, line 94: It may not be suitable to require connectionString and hosts in the `config.yml` any more. We may add validation annotations in `MongoConfiguration` so that one of both is defined. But that would be breaking, if anybody already uses the class in the AppConfiguration with `@Valid`. The easiest would be to suggest an empty default value for connection string and hosts in the `config.yml`.

sda-commons-server-mongo-testing/src/main/java/org/sdase/commons/server/mongo/testing/MongoDb.java
```diff
@@ -59,6 +59,9 @@ public interface MongoDb {
   /** @return the MongoDB options String without leading question mark */
   String getOptions();
 
+  /** @return the MongoDB connection String */
+  String getConnectionString();
+
```
JoergSiebahn, line 64: This is just a bonus feature for the new type of test setup shown in `MorphiaBundleWithConnectionStringIT`, correct?

sda-commons-server-morphia/src/main/java/org/sdase/commons/server/morphia/MongoConfiguration.java
```diff
@@ -103,6 +120,9 @@ public MongoConfiguration setDatabase(String database) {
   }
 
   public String getOptions() {
+    if (StringUtils.isBlank(options) && StringUtils.isNotBlank(connectionString)) {
+      // TODO what should be returned?
```
JoergSiebahn, line 124: The [mongodb-operator is doing some magic](https://github.com/SDA-SE/mongodb-operator/blob/master/src/main/java/com/sdase/k8s/operator/mongodb/controller/tasks/util/ConnectionStringUtil.java#L33-L39) with URI to get the query part. I wonder if we actually need a value here when the connection string is defined. Seems [like URI can handle](https://github.com/SDA-SE/mongodb-operator/blob/master/src/test/java/com/sdase/k8s/operator/mongodb/controller/tasks/util/ConnectionStringUtilTest.java#L11-L35) the comma in the host part.

sda-commons-server-morphia/src/main/java/org/sdase/commons/server/morphia/internal/ConnectionStringUtil.java
```diff
@@ -2,26 +2,42 @@
...
-  static String createConnectionString(MongoConfiguration configuration) {
+  static String createConnectionString(MongoConfiguration config) {
+    if (StringUtils.isNotBlank(config.getConnectionString())) {
+      LOGGER.info("Creating connection using connectionString");
+      if (StringUtils.isNotBlank(config.getHosts())
+          || StringUtils.isNotBlank(config.getDatabase())
+          || StringUtils.isNotBlank(config.getUsername())
+          || StringUtils.isNotBlank(config.getPassword())) {
+        LOGGER.warn(
+            "Please only use the 'connectionString' and no other properties to configure your database connection.");
```
JoergSiebahn, line 24:
```suggestion
            "Please only use the 'connectionString' and no other properties to configure your database connection. All other properties are ignored.");
```

sda-commons-server-mongo-testing/src/main/java/org/sdase/commons/server/mongo/testing/MongoDb.java
christopher-cudennec: Yes, correct.

sda-commons-server-morphia/src/main/java/org/sdase/commons/server/morphia/MongoConfiguration.java
christopher-cudennec: Isn't your example in the mongodb-operator prone to errors? The magic is not really clear to me.

sda-commons-server-morphia/src/main/java/org/sdase/commons/server/morphia/internal/ConnectionStringUtil.java
christopher-cudennec: 👍

sda-commons-server-morphia/README.md
christopher-cudennec: Wouldn't it be possible to stay backwards compatible here?
At the moment `hosts` and `database` are mandatory. I can add some custom validation "connectionString" and the other properties may not be set at the same time.
Why would it break things for clients?

christopher-cudennec: What option should we recommend in the documentation? Maybe it would be better to document two separate ways for the configuration: 1) Recommended option (connectionString?), 2) Legacy configuration

JoergSiebahn: It's not a backward compatibility issue with sda-dropwizard-commons for the services. But if services change there `config.yml` as suggested - without a default for hosts and connection string - it will be a breaking change for the deployments. When upgrading the service, the operations team must set `MONGODB_CONNECTION_STRING` to empty string, to be able to deploy as before. If they don't, the startup will fail when the config is substituted. I would suggest an empty default for both env's. Then both options are available. Imho a validation is not necessary. For me it doesn't matter if the service fails on reading the config or when connecting to the database. I assume both errors lead you to the cause in the logs.

Suggestion:

```yaml
mongo:
  connectionString: "${MONGODB_CONNECTION_STRING:-}"
    hosts: "${MONGODB_HOSTS:-}"
    database: "${MONGODB_DATABASE:-}"
    options: "${MONGODB_OPTIONS:-}"
```

christopher-cudennec: I changed the documented configuration using empty defaults for `hosts` and `connectionString`.

eikenilsknopp: But hosts and database are still required to be `@NotEmpty` in the MongoConfiguration class. I would prefer to remove that and either add some manual validation in the MongoConfiguration class or leave it out completely as Jörg mentioned as well. That would be maximum backwards compatibility, wouldn’t it?

JoergSiebahn (approved):

PR #1912, opened 2022-11-10
feat(server-morphia): Remove Morphia modules

christopher-cudennec:
The folders still exist; will be removed later


christopher-cudennec:
Build for Java 17 still fails due to the `EnvironmentRuleTest`. We should delete this specific rule in the next step. IMHO we should have a passing Java 17 build after that. 😃

elbiocaetano (approved):

PR #1921, opened 2022-11-14
fix(client-jersey): Remove deprecated methods `withConnectionTimeout`…

christopher-cudennec:
… and `withReadTimeout`

BREAKING CHANGE: Please provide a `HttpClientConfiguration` instead and set the `connectionTimeout` or `timeout` properties respectively.

### DIFF
--- a/sda-commons-client-jersey/src/main/java/org/sdase/commons/client/jersey/builder/AbstractBaseClientBuilder.java
+++ b/sda-commons-client-jersey/src/main/java/org/sdase/commons/client/jersey/builder/AbstractBaseClientBuilder.java
@@ -7,7 +7,6 @@ import io.dropwizard.client.JerseyClientBuilder;
 import io.dropwizard.setup.Environment;
 import io.opentracing.Tracer;
 import java.net.ProxySelector;
-import java.time.Duration;
 import java.util.ArrayList;
 import java.util.List;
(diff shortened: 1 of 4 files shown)


sda-commons-client-jersey/src/main/java/org/sdase/commons/client/jersey/builder/AbstractBaseClientBuilder.java
```diff
@@ -24,7 +23,7 @@
  *
  * @param <T> the type of the subclass
  */
-abstract class AbstractBaseClientBuilder<T extends AbstractBaseClientBuilder> {
+abstract class AbstractBaseClientBuilder<T extends AbstractBaseClientBuilder<?>> {
```
JoergSiebahn, line 26: Isn't this a `T` as well?

```suggestion
abstract class AbstractBaseClientBuilder<T extends AbstractBaseClientBuilder<T>> {
```

christopher-cudennec: I was afraid of the recursion, but yes 😆 
Still works when I use `T`. Updated the PR.

JoergSiebahn (approved):

PR #1967, opened 2022-11-25
Spring data mongo morphia compatibility

JoergSiebahn:


elbiocaetano:
Should we add a documentation on README.md file to explain the compatibility mode?

JoergSiebahn:
> Should we add a documentation on README.md file to explain the compatibility mode?

Imho we should not make it „a default“ by adding it to the README. I added a checklist item to the release notes task in the other story to describe the migration including the compatibility mode.

> The tests looks to be failing

I'll have a look … works on my machine.

elbiocaetano (approved):

PR #1973, opened 2022-11-30
feat(spring-data-mongo): adding option to disable saving class discminator

elbiocaetano:
…minator

### DIFF
--- a/sda-commons-server-spring-data-mongo/src/main/java/org/sdase/commons/server/spring/data/mongo/annotation/DocumentType.java
+++ b/sda-commons-server-spring-data-mongo/src/main/java/org/sdase/commons/server/spring/data/mongo/annotation/DocumentType.java
@@ -1,0 +1,13 @@
+package org.sdase.commons.server.spring.data.mongo.annotation;
+
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+@Retention(RetentionPolicy.RUNTIME)
+@Target(ElementType.TYPE)
+public @interface DocumentType {
+
+  String value() default "";
+}
--- a/sda-commons-server-spring-data-mongo/src/test/java/org/sdase/commons/server/spring/data/mongo/SpringDataMongoBundleClassDiscriminatorIT.java
+++ b/sda-commons-server-spring-data-mongo/src/test/java/org/sdase/commons/server/spring/data/mongo/SpringDataMongoBundleClassDiscriminatorIT.java
@@ -1,0 +1,324 @@
+package org.sdase.commons.server.spring.data.mongo;
+
+import static io.dropwizard.testing.ConfigOverride.config;
+import static io.dropwizard.testing.ConfigOverride.randomPorts;
+import static org.assertj.core.api.Assertions.assertThat;
+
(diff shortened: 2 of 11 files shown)


sda-commons-server-spring-data-mongo/src/test/java/org/sdase/commons/server/spring/data/mongo/SpringDataMongoBundleClassDiscriminatorIT.java
```diff
@@ -0,0 +1,324 @@
...
+    }
+  }
+
+  public static class WithoutClassDiscriminatorApp extends Application<MyConfiguration> {
+    private final SpringDataMongoBundle<MyConfiguration> springDataMongoBundle =
+        SpringDataMongoBundle.builder()
+            .withConfigurationProvider(MyConfiguration::getSpringDataMongo)
+            .withEntities(Dog.class, Cat.class)
+            .withoutClassDiscriminator(
+                new String[] {"org.sdase.commons.server.spring.data.mongo.example.model"})
```
eikenilsknopp, line 295: To me it feels very hacky to supply a String[] of packages. If it's not avoidable, couldn’t we simply use the applications base package or the package of the entities from `withEntities`? The loss of performance should be tolerable.

sda-commons-server-spring-data-mongo/src/main/java/org/sdase/commons/server/spring/data/mongo/annotation/DocumentType.java
```diff
@@ -0,0 +1,13 @@
+package org.sdase.commons.server.spring.data.mongo.annotation;
+
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+@Retention(RetentionPolicy.RUNTIME)
+@Target(ElementType.TYPE)
+public @interface DocumentType {
```
eikenilsknopp, line 10: To me, `DocumentType` does not convey the expected semantics. Maybe something like `SubType`?

JoergSiebahn: Spring data also knows about a `@TypeAlias` annotation. Could we just use that? Maybe we don't even have to care about the stored `_class` if we suggest and document to case `@TypeAlias`? I will play a little bit with that and try out.

JoergSiebahn:
> Personally, I'm not a big fan of annotations. If saving the subtype of the class is not avoidable, I wonder if we could do something in the direction of reflection (something compatible with Java 17) and find subclasses without the need for annotations etc?

In general I would agree with„I'm not a big fan of annotations“ besides some exceptions. If reflection is making it easier to understand is at least a matter to discuss ;) In the context of mapping annotations are actually how many frameworks (e.g. Jackson in an even more complicated way) are doing it for inheritance. So the developers should be used to this approach here. As shown in #1976 you see, that the `@TypeAlias` annotation of Spring Data follows the same pattern. The value is stored in the `_class` attribute instead of the class name in that case.

elbiocaetano:
Closed because https://github.com/SDA-SE/sda-dropwizard-commons/pull/1976

PR #1979, opened 2022-12-05
refactor: removing support for JUnit4

elbiocaetano:


sda-commons-server-auth-testing/src/main/java/org/sdase/commons/server/auth/testing/AuthClassExtension.java
```diff
@@ -14,7 +14,7 @@
 public class AuthClassExtension extends AbstractAuth
     implements BeforeAllCallback, AfterAllCallback {
 
-  private static final String AUTH_RULE_ENV_KEY = "AUTH_RULE";
+  public static final String AUTH_RULE_ENV_KEY = "AUTH_RULE";
```
eikenilsknopp, line 17: While we're at it: Is this naming still appropriate (rule -> extension)?

sda-commons-server-openapi/src/main/java/org/sdase/commons/server/openapi/OpenApiBundle.java
```diff
@@ -36,9 +36,9 @@
  * A bundle used to generate and serve API-first or code-first OpenAPI 3 files at the {@code
  * openapi.yaml} or {@code openapi.json} HTTP resources.
  *
- * <h3>Example Usage</h3>
+ * <p>Example Usage:
```
eikenilsknopp, line 39: Shouldn’t the paragraph be closed with </p>?

sda-commons-server-auth-testing/src/main/java/org/sdase/commons/server/auth/testing/AuthClassExtension.java
christopher-cudennec: I think it's a good opportunity to rename the variable name + value. Consequences for customers: Double check if you use that variable in any one of your `test-config.yaml`s and rename it accordingly.

sda-commons-server-openapi/src/main/java/org/sdase/commons/server/openapi/OpenApiBundle.java
christopher-cudennec: Javadoc is not HTML; a single `<p>` is correct according to [SO](https://stackoverflow.com/questions/33787753/javadoc-closing-tags) / the official documentation.

christopher-cudennec (commented):
3000+ lines of code removed 🥳

PR #2007, opened 2022-12-20
feat(server-kafka): provide option to set trace-header for records

eikenilsknopp:


sda-commons-server-kafka/src/main/java/org/sdase/commons/server/kafka/consumer/strategies/MessageListenerStrategy.java
```diff
@@ -105,4 +108,17 @@ public void commitOnClose(KafkaConsumer<K, V> consumer) {
...
   }
+
+  protected void startTraceContext(ConsumerRecord<K, V> consumerRecord) {
+    Header header = consumerRecord.headers().lastHeader(ConsumerTracing.TOKEN_HEADER);
+    if (header != null) {
+      TraceContext.storeTraceToken(new String(header.value()));
+    } else {
+      TraceContext.createNewTraceToken();
+    }
+  }
```
JoergSiebahn, line 119: Afaik, we should not keep the same context but make a connection between the received trace token and a new one for this new synchronous process.

PR #2026, opened 2023-01-03
fix: change variable name to tracing disabled

elbiocaetano:


sda-commons-server-opentelemetry/src/main/java/org/sdase/commons/server/opentelemetry/OpenTelemetryBundle.java
```diff
@@ -171,7 +171,7 @@ private static boolean isOpenTelemetryDisabled() {
       LOG.warn("Tracing is disabled using deprecated configuration.");
       return true;
     }
-    return "true".equals(getProperty(OTEL_DISABLED));
+    return "true".equals(getProperty(TRACING_DISABLED));
```
JoergSiebahn, line 174: NIT: Would it make sense to compare ignoring case?

JoergSiebahn (approved):

sda-commons-server-opentelemetry/src/main/java/org/sdase/commons/server/opentelemetry/OpenTelemetryBundle.java
elbiocaetano: Usually the environment variables are upper case only, right? I will write down, so we can discuss it later

JoergSiebahn: The variables are upper case, but the value may support `true` and `TRUE`.

PR #2042, opened 2023-01-09
feat(server-spring-data-mongo): remove spring application context

eikenilsknopp:


christopher-cudennec (commented):
Automatic validation being removed is another thing to add to the migration guide from Morphia to Spring Data Mongo. We need to keep that in mind.

christopher-cudennec (approved):
Please squash on merge. 👍

PR #2044, opened 2023-01-10
docs(server-spring-data-mongo): add migration guide

eikenilsknopp:


sda-commons-server-spring-data-mongo/docs/MIGRATION_GUIDE.md
```diff
@@ -0,0 +1,108 @@
...
+
+```java
+boolean isValid =
+  Validation.buildDefaultValidatorFactory()
+    .getValidator()
+    .validate(myEntity)
+    .isEmpty();
+```
+
+## Queries
```
christopher-cudennec, line 56: let's add a description how to replace Morphia's Datastore when initializing the application.

sda-commons-server-spring-data-mongo/docs/MIGRATION_GUIDE.md
```diff
@@ -0,0 +1,108 @@
...
+-    MorphiaBundle.builder()
+-      .withConfigurationProvider(AppConfiguration::getMongo)
+-      .withEntityScanPackageClass(ConsentConfigurationEntity.class)
+-      .withValidation()
+-      .build();
+
++  private final SpringDataMongoBundle<AppConfiguration> mongoBundle = 
++    SpringDataMongoBundle.builder()
++      .withConfigurationProvider(AppConfiguration::getMongo)
++      .build();
```
JoergSiebahn, line 22: I remember that there are some case, that require to register the entities directly. Not sure if that was about inheritance or generics or if we already found a way that it is not needed.

sda-commons-server-spring-data-mongo/docs/MIGRATION_GUIDE.md
```diff
@@ -0,0 +1,117 @@
...
+
+## Queries
+
+### Building queries with the datastore
+Simple operations can be realised by the MongoOperations directly.
+
+Saving an entity:
+```java
+public void save(SampleEntity entity) {
+    mongoOperations.save(entity);
```
JoergSiebahn, line 64: I would prefer `insert`. `save` is effectively an `upsert` and will also overwrite existing documents while `insert` is meant for creating new documents. As MongoDB has very little assertions in regards of transactions, you should avoid to overwrite full documents, because you may revert recent changes made in another request between the entity has been loaded and is then saved. The alternative would be to do an explicit update/upsert query for the fields that should be changed and don't even need to load the document before. This may not apply for ODS apps where it is ensured, that data is handled in sequence. Afaik Morphia does not have such a difference in the API.

MIGRATION_GUIDE.md
```diff
@@ -0,0 +1,364 @@
...
+bootstrap.addBundle(JaegerBundle.builder().build());
+bootstrap.addBundle(OpenTracingBundle.builder().build());
+```
+
+After:
+```java
+bootstrap.addBundle(
+        OpenTelemetryBundle.builder()
+            .withAutoConfiguredTelemetryInstance()
+            .withExcludedUrlsPattern(Pattern.compile(String.join("|", excludedTracingUrls)))
```
christopher-cudennec, line 19: @elbiocaetano : Is that mandatory? The setup now looks more complicated than for v2. What is the value of `excludedTracingUrls`?

MIGRATION_GUIDE.md
```diff
@@ -0,0 +1,364 @@
...
+```gradle
+testImplementation 'io.opentelemetry:opentelemetry-sdk-testing'
+```
+
+### Executors need to be instrumented to follow traces when parts of the request are handled asynchronously.
+
+Before:
+```java
+return new AsyncEmailSendManager(
+  new InMemoryAsyncTaskRepository<>(limits.getMaxAttempts()),
```
christopher-cudennec, line 40: Can we give some advice for `limits.getMaxAttempts()`?

MIGRATION_GUIDE.md
elbiocaetano: This is a list of excluded urls, e.g: healthcheck and metrics. it is not mandatory, but I liked to leave there, because I thought it would be useful when configuring manually. Also, in the dw3 it automatically excludes some patterns of urls, so to be the same behavior they should do the same in their app. Do you think I can put in the code the declaration of excludedTracingUrls?

MIGRATION_GUIDE.md
elbiocaetano: This is just a parameter for the creation of `InMemoryAsyncTaskRepository`. It has nothing to do with open telemetry. The important part is wrapping the async task inside a open telemetry Context

MIGRATION_GUIDE.md
```diff
@@ -0,0 +1,364 @@
...
+Morphia's `Datastore` can be replaced by Spring Data Mongo's `MongoOperations`:
+```java
+public MongoOperations getMongoOperations() {
+  return this.mongoBundle.getMongoOperations();
+}
+```
+
+#### Morphia compatibility
+When upgrading a service from 2.x.x a set of converters is provided to stay compatible with a
+database that was initialised with Morphia. To enable these converters, `.withMorphiaCompatibility()`
```
elbiocaetano, line 204: It shoud be `initialized`, instead of `initialised`, right?

MIGRATION_GUIDE.md
```diff
@@ -0,0 +1,364 @@
+# Migration Guide
```
elbiocaetano, line 1:
```suggestion
# Migration Guide from v2 to v3
```

MIGRATION_GUIDE.md
christopher-cudennec: What about adding the value that we use in the `SdaPlatformBundle`? Since that value was not needed in v2 people would not know what to use here.

MIGRATION_GUIDE.md
christopher-cudennec: Ah, sorry. I overread that this is the "Before" section. I can see it in the "After" section. 👍

MIGRATION_GUIDE.md
```diff
@@ -0,0 +1,364 @@
...
+JAEGER_SAMPLER_TYPE=0
+```
+
+After:
+```properties
+TRACING_DISABLED=true
+```
+
+### New environment variables
+In order to configure Open Telemetry, you have to set some environment variables:
```
christopher-cudennec, line 79: It sounds like setting these variables is mandatory. But I suppose it's optional, right?

MIGRATION_GUIDE.md
```diff
@@ -0,0 +1,364 @@
...
+```
+
+Morphia's `Datastore` can be replaced by Spring Data Mongo's `MongoOperations`:
+```java
+public MongoOperations getMongoOperations() {
+  return this.mongoBundle.getMongoOperations();
+}
+```
+
+#### Morphia compatibility
```
christopher-cudennec, line 202: let's use level 3 indentation for all subheaders.

MIGRATION_GUIDE.md
christopher-cudennec: I think we mostly use the US spelling with a `Z`.

MIGRATION_GUIDE.md
```diff
@@ -0,0 +1,364 @@
...
+
+Spring Data Mongo - `query.addCriteria(where("id").is(getId()));`
+
+##### Contains
+
+Morphia supports _contains()_ method. Use _regex()_ in Spring Data Mongo. For example
+
+Morphia - `query.criteria("fieldName").contains(entity.getFieldName());`
+
+Spring Data Mongo - `query.addCriteria(where("fieldName").regex(".*" + entity.getFieldName() + ".*"));`
```
christopher-cudennec, line 314: I'd use `Pattern.quote` ([see explanation](https://www.baeldung.com/java-pattern-quote#use-patternquote)) to make readers aware that this might be necessary.

MIGRATION_GUIDE.md
```diff
@@ -0,0 +1,364 @@
...
+### Indexes
+
+While all `@Indexes` can be created at the main entity class in Morphia, each field must be
+`@Indexed`, `@TextIndexed` or `@WildcardIndexed` where it is defined in Spring Data Mongo.
+A `@CompoundIndex` can be declared at the entity class.
+
+### Custom converters
+
+Custom converters can be added directly to the Entity class in Morphia by using annotation like `@Converters(value = {ExampleConverter.class})`. 
+In Spring Data Mongo the custom converters can be added to the builder in your `Application.class`. The entity class that uses the converters must also be registered in the builder like below:
```
christopher-cudennec, line 343: I would remove mentioning to register the entity class. That is misleading because registering the entity and adding a converter is unrelated. Converters will work globally for all places where a specific type is found.

MIGRATION_GUIDE.md
preethivenkatesh-sdase: updated

MIGRATION_GUIDE.md
preethivenkatesh-sdase: updated

MIGRATION_GUIDE.md
preethivenkatesh-sdase: updated

MIGRATION_GUIDE.md
preethivenkatesh-sdase: updated the query. Let me know if it required further explanation

sda-commons-server-spring-data-mongo/docs/MIGRATION_GUIDE.md
christopher-cudennec: Wasn't it only about performance? I think the initialization will be done if you save your first entity. If you don't want that you can register your entity when initializing your bundle.

MIGRATION_GUIDE.md
```diff
@@ -0,0 +1,348 @@
...
+```
+See the [official documentation](https://docs.spring.io/spring-data/jpa/docs/current/reference/html/#repositories.query-methods.details) for further details.
+
+### Query functions
+As Spring Data Mongo doesn't support/provide many query functions provided by Morphia, below are some replacements/alternatives.
+
+* #### _EqualIgnoreCase_
+  Morphia supports usage of _equalIgnoreCase()_. Use _regex()_ in Spring Data Mongo. For example
+  * Morphia - `query.criteria("fieldName").equalIgnoreCase(entity.getFieldName());`
+  * Spring Data Mongo - `query.addCriteria(where("fieldName").regex(Pattern.complie("^" + Pattern.quote(entity.getFieldName()), Pattern.CASE_INSENSITIVE)));`
```
christopher-cudennec, line 286: Small typo "complie" -> "compile"

sda-commons-server-spring-data-mongo/docs/MIGRATION_GUIDE.md
preethivenkatesh-sdase: updated

elbiocaetano (approved):

sda-commons-server-spring-data-mongo/docs/MIGRATION_GUIDE.md
JoergSiebahn: In case you store an abstract class or interface - effectively multiple implementations - as entity or field like in [this test app](https://github.com/SDA-SE/sda-dropwizard-commons/blob/feat%2Fv3/sda-commons-server-spring-data-mongo/src/test/java/org/sdase/commons/server/spring/data/mongo/inheritance/InheritanceMappingITest.java#L122) it is needed. Two of the tests fail if you remove the entity registration. If your first action is a read action or you never stored the type you read in the currently running instance before reading, it will fail because the type is not known.

PR #2055, opened 2023-01-17
feat:  provides api to register a callbacks

saadmar:
* Extends the `MessageProducer` api to register a callback when using a non blocking producer.
* Adds standard error logging for non blocking producers

We had a use case where a producer silently failed to produce a message (due to large size) and there was no trace for that as everything happens in a separate thread.

Another option for us could be blocking the thread and explicitly logging the exception, but we thought that could be good to have for all cases.

### DIFF
--- a/sda-commons-server-kafka/src/main/java/org/sdase/commons/server/kafka/producer/MessageProducer.java
+++ b/sda-commons-server-kafka/src/main/java/org/sdase/commons/server/kafka/producer/MessageProducer.java
@@ -28,7 +28,8 @@ public interface MessageProducer<K, V> {
    * @param key key to send
    * @param value value to send
    * @param headers headers to send
-   * @param callback A callback that will be triggered once the message is sent
+   * @param callback A callback that will be triggered once the request is complete (if the ACK is
+   *     received or an exception is raised).
    * @return The result of the send is a {@link RecordMetadata} specifying the partition the record
    *     was sent to, the offset it was assigned and the timestamp of the record.
    */


sda-commons-server-kafka/src/test/java/org/sdase/commons/server/kafka/producer/KafkaMessageProducerIT.java
```diff
@@ -0,0 +1,107 @@
...
+    // then
+    Awaitility.await()
+        .untilAsserted(
+            () -> {
+              assertThat(out.capturedLines())
+                  .anyMatch(l -> l.contains(statement))
+                  .anyMatch(
+                      l ->
+                          l.contains(
+                              "An error occurred while producing a message with key k to the topic test-topic."));
```
JoergSiebahn, line 81: [Above](https://github.com/SDA-SE/sda-dropwizard-commons/pull/2055/files#diff-ef9a7d90b70bb2c2e409f94bca64cad99025d1f5e78524565233017dc27a7fb6R31) it is documented, that the callback is called when the message is sent. In my understanding that means success, but here you verify that the callback is called and that an error occurred. That seems to be different to the documentation. Can you explain that?

saadmar: Right, the documentation is not accurate, this should be triggered in both cases "when the request is complete".
Changed

PR #2100, opened 2023-02-03
fix(server-kafka): removing deprecated methods

elbiocaetano:
BREAKING CHANGE:
removed the topicana classes in package `org.sdase.commons.server.kafka.topicana`;
removed method `checkTopicConfiguration` and changed the methods `forTopic(ExpectedTopicConfiguration topic)` to `forTopic(TopicConfig topic)` and `forTopics(Collection<ExpectedTopicConfiguration> topics)` to `forTopics(Collection<TopicConfig> topics)` on `org.sdase.commons.server.kafka.builder.MessageListenerRegistration` and `org.sdase.commons.server.kafka.builder.ProducerRegistration`.


sda-commons-server-kafka/src/test/java/org/sdase/commons/server/kafka/KafkaTopicIT.java
```diff
@@ -99,13 +83,7 @@ void checkTopicSuccessfulComplex() {
...
             MessageListenerRegistration.builder()
                 .withDefaultListenerConfig()
-                .forTopicConfigs(
-                    Collections.singletonList(
-                        TopicConfigurationBuilder.builder(topicName)
-                            .withPartitionCount(2)
-                            .withReplicationFactor(1)
-                            .build()))
-                .checkTopicConfiguration()
+                .forTopic(topicName)
```
christopher-cudennec, line 86: I like the simplification 😄

sda-commons-server-opentelemetry/src/test/java/org/sdase/commons/server/opentelemetry/servlet/TracingAsyncListenerIT.java
```diff
@@ -38,7 +39,10 @@ void shouldTraceAsyncServlets() {
 
     assertThat(response.getStatus()).isEqualTo(SC_OK);
 
+    // setting the pollDelay and maximum timeout, because it is an async call
```
christopher-cudennec, line 42: separate commit for this change?

MIGRATION_GUIDE.md
```diff
@@ -374,17 +374,31 @@ Implementing the converter changes from Morphia interface to using the Spring in
...
+on [MessageListenerRegistration](./sda-commons-server-kafka/src/main/java/org/sdase/commons/server/kafka/builder/MessageListenerRegistration.java)
+and [ProducerRegistration](./sda-commons-server-kafka/src/main/java/org/sdase/commons/server/kafka/builder/ProducerRegistration.java):
+
+### checkTopicConfiguration
+This method compared the current topics with the configured ones and threw a `org.sdase.commons.server.kafka.topicana.MismatchedTopicConfigException` in case they did not match.
+
+### forTopicConfigs
+These methods accepted a list of `org.sdase.commons.server.kafka.topicana.ExpectedTopicConfiguration` to compare against the current topic configuration.
+
+## forTopic(ExpectedTopicConfiguration topic)
```
christopher-cudennec, line 403: Last header should probably use three #, i.e. "###"? You could also make them bullet points.

sda-commons-server-kafka-example/src/main/java/org/sdase/commons/server/kafka/KafkaExampleConsumerApplication.java
```diff
@@ -58,7 +57,7 @@ private void createExampleMessageListener(ObjectMapper configuredObjectMapper) {
     kafka.createMessageListener(
         MessageListenerRegistration.builder()
             .withDefaultListenerConfig()
-            .forTopicConfigs(Collections.singletonList(kafka.getTopicConfiguration("example0")))
+            .forTopic("exampleTopic")
```
christopher-cudennec, line 60: Do we also need to change the [`config.yaml`](https://github.com/SDA-SE/sda-dropwizard-commons/blob/fe3f06fd5a09c49b1634e625e9f7537e92773b1d/sda-commons-server-kafka-example/src/test/resources/test-config-consumer.yml)? Can we now remove `kafka.topics` completely?

Can you double-check that the `config` that was defined in the `config.yaml` was not used / is not important?

sda-commons-server-kafka/src/main/java/org/sdase/commons/server/kafka/KafkaConfiguration.java
```diff
@@ -25,8 +24,6 @@ public class KafkaConfiguration {
 
   private Map<String, ConsumerConfig> consumers = new HashMap<>();
 
-  private Map<String, TopicConfig> topics = new HashMap<>();
```
christopher-cudennec, line 28: I think we should provide one preferred way how we want users of the library to define/configure topic names. Usually they would need to override the specific topic name used in their environment. Options that come to my mind:

* we keep this key but only give the opportunity to set the topic name
* we use a convention that we try to lookup "KAFKA_TOPIC_NAME_<topic-name>" in the environment

What do you think @elbiocaetano / @JoergSiebahn / @eikenilsknopp / @preethivenkatesh-sdase ?

sda-commons-server-kafka/src/main/java/org/sdase/commons/server/kafka/builder/MessageListenerRegistration.java
```diff
@@ -1,36 +1,31 @@
...
-  private Collection<ExpectedTopicConfiguration> topics;
-  private boolean checkTopicConfiguration;
   private MessageListenerStrategy<K, V> strategy;
 
   private ConsumerConfig consumerConfig;
   private String consumerConfigName;
   private String listenerConfigName;
   private ListenerConfig listenerConfig;
 
+  private List<String> topicsNames;
```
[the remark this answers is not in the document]
JoergSiebahn: As we are going for v3, we could do it right.

sda-commons-server-kafka/src/main/java/org/sdase/commons/server/kafka/KafkaConfiguration.java
JoergSiebahn: What should `KAFKA_TOPIC_NAME_` refer to? So, what should we lookup at the end of the property?

Although it would cause refactoring in the services when upgrading, I would prefer not to keep a map just for a single value and not to do some unknown magic with environment variables.

I would suggest for the producer services, to add dedicated string properties in their `config.yml` / `AppConfiguration` or to extend the `KafkaConfiguration` to add their specific properties like `myModelTopicName`. The environment variables in the `config.yml` can be the same as before but write to a new property. This way, the service will be backward compatible to existing deployments. We have various services where we use inheritance for configuring Api clients. We also tried [to migrate that into the lib](https://github.com/SDA-SE/sda-dropwizard-commons/pulls?q=is%3Apr+is%3Aclosed+dedicated+api+client+configuration), but never finished it.

But to make the migration smoother, we could deprecate all properties but `name` in `TopicConfig` in v2 and keep this single property in v3 and think about a long term solution when v3 is released.

sda-commons-server-kafka/src/main/java/org/sdase/commons/server/kafka/KafkaBundle.java
```diff
@@ -107,29 +98,27 @@ public void run(C configuration, Environment environment) {
...
+   * Provides a {@link TopicConfig} that is generated from the values within the configuration yaml
    *
    * @param name the name of the topic
    * @return the configured topic configuration
    * @throws ConfigurationException if no such topic exists in the configuration
    */
-  public ExpectedTopicConfiguration getTopicConfiguration(String name)
-      throws ConfigurationException { // NOSONAR
-    if (topics.get(name) == null) {
+  public TopicConfig getTopicConfiguration(String name) throws ConfigurationException { // NOSONAR
```
christopher-cudennec, line 107: Change it to "key" or "topicKey" and also adapt the documentation?

sda-commons-server-kafka/src/test/java/org/sdase/commons/server/kafka/KafkaTopicIT.java
```diff
@@ -78,8 +64,8 @@ void checkTopicSuccessful() {
         bundle.createMessageListener(
             MessageListenerRegistration.builder()
                 .withDefaultListenerConfig()
-                .forTopic(topicName)
-                .checkTopicConfiguration()
+                .forTopicConfigs(
```
christopher-cudennec, line 67: easier: just `forTopic`?

sda-commons-server-kafka/README.md
```diff
@@ -338,22 +334,13 @@ kafka:
         acks: all
         retries: 1000
   # Map with topic configurations. The key is the name of the topic and is also used to address the configuration within the code
-  # Topic descriptions can be used to validate the brokers topic configuration or to create new topics
   topics:
     # id of the topic configuration
-    topic1:
+    topic1: # topic key
```
JoergSiebahn, line 339: 👍

sda-commons-server-kafka/src/main/java/org/sdase/commons/server/kafka/KafkaBundle.java
```diff
@@ -107,29 +98,27 @@ public void run(C configuration, Environment environment) {
...
-    return topics.get(name);
+    return kafkaConfiguration.getTopics().get(key);
   }
 
   /**
    * Creates a number of message listeners with the parameters given in the {@link
    * MessageListenerRegistration}.
    *
-   * <p>The depricated fields of the {@link ListenerConfig} from {@link MessageListenerRegistration}
+   * <p>The deprecated fields of the {@link ListenerConfig} from {@link MessageListenerRegistration}
```
JoergSiebahn, line 121: I don't see any deprecated fields in [`ListenerConfig`](https://github.com/SDA-SE/sda-dropwizard-commons/blob/PLP-642-remove-check-topic-configuration-and-topicana-classes-for-v-3/sda-commons-server-kafka/src/main/java/org/sdase/commons/server/kafka/config/ListenerConfig.java). Is this sentence still needed?
Thanks for raising awareness by fixing the typo. 🤣

PR #2111, opened 2023-02-07
fix(server-s3-testing): fix `SdkClientException` in `S3MockExtension` during test execution

vladimirsvicevicsrb:
Since S3Mock accepts any authentication provider without validating it, this PR is setting up Anonymous authentication mechanism to avoid unnecessary stack traces like:
`com.amazonaws.SdkClientException: Failed to connect to service endpoint`

### DIFF
--- a/sda-commons-server-s3-testing/src/main/java/org/sdase/commons/server/s3/testing/S3Mock.java
+++ b/sda-commons-server-s3-testing/src/main/java/org/sdase/commons/server/s3/testing/S3Mock.java
@@ -1,5 +1,7 @@
 package org.sdase.commons.server.s3.testing;
 
+import com.amazonaws.auth.AWSStaticCredentialsProvider;
+import com.amazonaws.auth.AnonymousAWSCredentials;
 import com.amazonaws.client.builder.AwsClientBuilder;
 import com.amazonaws.regions.Regions;
 import com.amazonaws.services.s3.AmazonS3;
@@ -101,6 +103,7 @@ public class S3Mock {
             new AwsClientBuilder.EndpointConfiguration(
                 getEndpoint(), Regions.DEFAULT_REGION.getName()))
         .withPathStyleAccessEnabled(true)
+        .withCredentials(new AWSStaticCredentialsProvider(new AnonymousAWSCredentials()))
         .build();
   }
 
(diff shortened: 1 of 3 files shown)


christopher-cudennec:
Can you verify that the change has the desired effect? You can use the PR snapshot in your service to give it a try.
Otherwise it's hard to decide what the consequences of your changes are.
Is it possible to reproduce the error in a test case?

vladimirsvicevicsrb:
> 

Hey @christopher-cudennec, changes does not have effect on functionality of `s3mock`, since the tests for `S3ClassExtensionTest` and `S3MockRuleTest` worked before as well. The only difference is to avoid having this weird error stack trace for `SdkClientException`.
That can also be confirmed by running old version of above-mentioned test classes and new versions and compare that the error messages are not popping up anymore in the fixed version.

JoergSiebahn:
Thanks for the fix. I have a few questions regarding behavior and communication:

- Will existing tests still work without removing [the credential settings](https://github.com/SDA-SE/sda-dropwizard-commons/pull/2111/files#diff-5c308f5edc94ea19ea2706454ebf6ced17790268a725ccd267e8937d80d64e59L36-L41)?
- Is the change from a user perspective mainly „no exception“ or is it more positive „no need to setup credentials in S3 tests any more“. In the latter case, I would prefer a _feat_ commit explaining the advantage.
- If consumers have to or should remove credential settings in their tests, we should have an additional note prepared that we can add to the release notes.

vladimirsvicevicsrb:
Hey @JoergSiebahn,

> Will existing tests still work without removing [the credential settings](https://github.com/SDA-SE/sda-dropwizard-commons/pull/2111/files#diff-5c308f5edc94ea19ea2706454ebf6ced17790268a725ccd267e8937d80d64e59L36-L41)?

Yes, I ran it locally and I don't see any effect on the existing tests. They still work. I am not even sure what we need [the credential settings](https://github.com/SDA-SE/sda-dropwizard-commons/pull/2111/files#diff-5c308f5edc94ea19ea2706454ebf6ced17790268a725ccd267e8937d80d64e59L36-L41) when they are not set on the referenced/linked anywhere. 🤔 

> Is the change from a user perspective mainly „no exception“ or is it more positive „no need to setup credentials in S3 tests any more“. In the latter case, I would prefer a feat commit explaining the advantage.

I guess the first one. On the other hand, from what I saw in the official library that we use for `S3`, [authentication was never supported/validated](https://github.com/findify/s3mock/blob/master/README.md?plain=1#L21), so I don't see a reason of passing any credential provider for `S3` testing.

> If consumers have to or should remove credential settings in their tests, we should have an additional note prepared that we can add to the release notes.

I am wondering who/how they will provide custom `CredentialProvider`, when if they call [getS3Client](https://github.com/SDA-SE/sda-dropwizard-commons/pull/2111/files#diff-cfab490829da5f640e6c3c5d702745614bba48756560baca5d49cd8a46c63962L98) it is already built and there is no way to modify it anymore. Am I right?

JoergSiebahn (approved):

PR #2115, opened 2023-02-08
Provide handling of a `MetadataContext` to track metadata across services for synchronous, asynchronous and interrupteds business processes

JoergSiebahn:


sda-commons-server-dropwizard/src/main/java/org/sdase/commons/server/dropwizard/metadata/DetachedMetadataContext.java
```diff
@@ -0,0 +1,34 @@
+package org.sdase.commons.server.dropwizard.metadata;
+
+import java.util.ArrayList;
+import java.util.LinkedHashMap;
+
+/**
+ * A {@link MetadataContext} representation which internals <strong>do not reflect changes</strong>
```
christopher-cudennec, line 7: I don't understand the word "internals" here. Would also be "does not" grammatically.

sda-commons-server-dropwizard/src/main/java/org/sdase/commons/server/dropwizard/metadata/MetadataContext.java
```diff
@@ -0,0 +1,119 @@
+package org.sdase.commons.server.dropwizard.metadata;
+
+import java.util.List;
+import java.util.Set;
+import java.util.concurrent.Callable;
+
+/**
+ * The {@code MetadataContext} stores information about a long-running business processes that is
+ * independent of specific APIs or business rules of a single service. A {@code MetaContext} is
+ * built from the metadata fields of input APIs like HTTP of Kafka message headers. It is added
```
christopher-cudennec, line 10: ... HTTP **or** Kafka message headers ...

sda-commons-server-dropwizard/src/main/java/org/sdase/commons/server/dropwizard/metadata/MetadataContext.java
```diff
@@ -0,0 +1,119 @@
...
+ * the {@code KafkaMessageProducer}.
+ *
+ * <p>Services that claim to support {@code MetadataContext} must take care that it is also kept and
+ * reloaded when the business process is interrupted and proceeded later. An interrupted business
+ * process in these terms is for example asynchronous processing in a new {@link Thread} or
+ * finishing the current step of the process by saving the current state to a database.
+ *
+ * <p>The metadata fields (e.g. header names) that are put into the {@code MetadataContext} must be
+ * configured by {@linkplain System#getenv(String) environment variable} or {@linkplain
+ * System#getProperty(String) system property} {@code METADATA_HEADERS} as comma separated list of
```
christopher-cudennec, line 21: Reading your description I would prefer `METADATA_FIELDS` as name.

sda-commons-server-dropwizard/src/main/java/org/sdase/commons/server/dropwizard/metadata/MetadataContext.java
```diff
@@ -0,0 +1,119 @@
...
+    return MetadataContextUtil.current();
+  }
+
+  /**
+   * @return the current metadata context, never {@code null}. Changes in the returned instance will
+   *     not affect the {@link MetadataContext} of the current {@link Thread}.
+   */
+  static DetachedMetadataContext detachedCurrent() {
+    return DetachedMetadataContext.of(MetadataContextUtil.current());
+  }
```
christopher-cudennec, line 47: We could add `@NotNull` to the return type. Would make it easier for IDEs to find possible bugs. We've never discussed if we want to make use of that approach AFAIK.

sda-commons-server-dropwizard/src/main/java/org/sdase/commons/server/dropwizard/metadata/MetadataContext.java
```diff
@@ -0,0 +1,119 @@
...
+   *     in a new thread.
+   */
+  static <V> Callable<V> transferMetadataContext(Callable<V> callable) {
+    return MetadataContextUtil.transferMetadataContext(callable);
+  }
+
+  /**
+   * @return all available keys in the metadata context.
+   */
+  Set<String> metadataKeys();
```
christopher-cudennec, line 86: I'd prefer `keys()` as method name. Does it clash with implementing another interface?

sda-commons-server-dropwizard/src/main/java/org/sdase/commons/server/dropwizard/metadata/MetadataContext.java
```diff
@@ -0,0 +1,119 @@
...
+
+  /**
+   * {@link #valuesByConfiguredMetadataKey(String)} should be preferred to be independent of
+   * different environments.
+   *
+   * @param key a key in the metadata context
+   * @return the values stored in the metadata context by this key
+   */
+  List<String> valuesByMetadataKey(String key);
```
christopher-cudennec, line 95: I'd prefer `valuesByKey` (see above).

sda-commons-server-dropwizard/src/main/java/org/sdase/commons/server/dropwizard/metadata/MetadataContext.java
```diff
@@ -0,0 +1,119 @@
...
+   * own assumptions about the naming.
+   *
+   * @param environmentOrPropertyName the name of a {@linkplain System#getProperty(String) system
+   *     property} or {@linkplain System#getenv(String) environment variable} that defines the
+   *     actual key for a specific need of the service.
+   * @return the values stored in the metadata context by this key
+   * @throws KeyConfigurationMissingException if the given {@code environmentOrPropertyName} does
+   *     not resolve to a metadata context key
+   */
+  default List<String> valuesByConfiguredMetadataKey(String environmentOrPropertyName)
```
christopher-cudennec, line 115: I'd prefer `valuesByEnvKey` instead of "configured". It would make it clearer IMHO that where we look up the key. I guess you don't mean that the value comes from the configuration file.

sda-commons-server-dropwizard/src/main/java/org/sdase/commons/server/dropwizard/metadata/InternalMetadataContext.java
```diff
@@ -0,0 +1,28 @@
+package org.sdase.commons.server.dropwizard.metadata;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Set;
+
+class InternalMetadataContext implements MetadataContext {
```
christopher-cudennec, line 7: alternative name `UnmodifiableMetadataContext` comparable to Java collections?

christopher-cudennec (commented):
Can you add a test that shows how clients would create store the current context to a database and restore the values? (don't really use a MongoDB for the test but as showcase) It would help me to see how the API feels for them.

JoergSiebahn:
> Can you add a test that shows how clients would create store the current context to a database and restore the values? (don't really use a MongoDB for the test but as showcase) It would help me to see how the API feels for them.

Yes, I will also ensure with an IT that the object can be stored and loaded to the database. But before investing too much time into tests and other implementation (e.g. HTTP/Kafka), I wanted to discuss the API.

sda-commons-server-dropwizard/src/main/java/org/sdase/commons/server/dropwizard/metadata/MetadataContext.java
JoergSiebahn: That would match the main production use case. We are also looking it up in the system properties (mainly to be able to test). As the parameter name makes that clear as well, I'm fine with renaming. What do you think about `valuesByKeyFromEnvironment(String environmentOrPropertyName) `. Imho that is more exact.

sda-commons-server-dropwizard/src/main/java/org/sdase/commons/server/dropwizard/metadata/MetadataContext.java
JoergSiebahn: I'll try.

sda-commons-server-dropwizard/src/main/java/org/sdase/commons/server/dropwizard/metadata/MetadataContext.java
JoergSiebahn: I'm not a big fan of annotating the code that way.

sda-commons-server-dropwizard/src/main/java/org/sdase/commons/server/dropwizard/metadata/MetadataContext.java
JoergSiebahn: Sounds good 👍

sda-commons-server-dropwizard/src/main/java/org/sdase/commons/server/dropwizard/metadata/DetachedMetadataContext.java
```diff
@@ -0,0 +1,36 @@
+package org.sdase.commons.server.dropwizard.metadata;
+
+import java.util.ArrayList;
+import java.util.LinkedHashMap;
+import java.util.List;
+
+/**
+ * A {@link MetadataContext} representation which is detached from the {@linkplain
+ * MetadataContext#current() current} {@link MetadataContext}. Changes in this class do not
+ * propagate to the {@linkplain MetadataContext#current() current} {@link MetadataContext} vice
```
christopher-cudennec, line 10: nit: ... *and* vice versa.

sda-commons-server-dropwizard/src/main/java/org/sdase/commons/server/dropwizard/metadata/MetadataContext.java
christopher-cudennec: Then let's not start with it. 😄

sda-commons-server-dropwizard/src/main/java/org/sdase/commons/server/dropwizard/metadata/MetadataContext.java
christopher-cudennec: I like `...ByKeyFromEnvironment` 👍

sda-commons-server-dropwizard/src/main/java/org/sdase/commons/server/dropwizard/metadata/MetadataContext.java
```diff
@@ -0,0 +1,119 @@
...
+  Set<String> keys();
+
+  /**
+   * {@link #valuesByKeyFromEnvironment(String)} should be preferred to be independent of different
+   * environments.
+   *
+   * @param key a key in the metadata context
+   * @return the values stored in the metadata context by this key
+   */
+  List<String> valuesByKey(String key);
```
christopher-cudennec, line 95: Do we really want to store multiple values for one key? Are you thinking about the same semantics like for HTTP headers? In most cases clients will probably only expect one value, e.g. `tenant-id`.

JoergSiebahn: Imho the library should not restrict that and should not decide what happens, if multiple headers are received.

sda-commons-server-dropwizard/src/main/java/org/sdase/commons/server/dropwizard/metadata/MetadataContext.java
```diff
@@ -0,0 +1,119 @@
...
+   * @return the current metadata context, never {@code null}. Changes in the returned instance will
+   *     not affect the {@link MetadataContext} of the current {@link Thread}.
+   */
+  static DetachedMetadataContext detachedCurrent() {
+    return DetachedMetadataContext.of(MetadataContextUtil.current());
+  }
+
+  /**
+   * Creates a new {@link MetadataContext} for the current {@link Thread}. Any existing {@link
+   * MetadataContext} in the current {@link Thread} will be replaced.
```
saadmar, line 51: Could we add warning here for the replacement behaviour? or maybe away how to suppress it and passively set it if the user needs it? 
From what I understood, depending on the specific service implementation, restoring the current context from a `DetachedMetadataContext` for e.g from a database could lead to some unexpected behaviour if in the case of the process triggered from a http request with predefined metadata, the user could then define different `createContext` strategies (replace, skip, fail...) in multiple locations in the implementation

sda-commons-server-dropwizard/src/main/java/org/sdase/commons/server/dropwizard/metadata/MetadataContextFilter.java
```diff
@@ -0,0 +1,64 @@
...
+
+  private final Set<String> metadataFields;
+
+  public MetadataContextFilter(Set<String> metadataFields) {
+    this.metadataFields = metadataFields;
+  }
+
+  @Override
+  public void filter(ContainerRequestContext requestContext) throws IOException {
+    var metadataContext = buildContext(requestContext.getHeaders());
```
christopher-cudennec, line 33: Better safe than sorry: Should we also add `MetadataContextHolder.clear();` here?

sda-commons-server-dropwizard/src/main/java/org/sdase/commons/server/dropwizard/metadata/MetadataContext.java
JoergSiebahn: > Especially if the current service does not know all the metadata upfront as they could evolve over time

That should not happen. All fields that are considered must be specified by [`METADATA_FIELDS`](https://github.com/SDA-SE/sda-dropwizard-commons/pull/2115/files#diff-4c7fc6f534d55a8e5faba904dda14f3e525d90d61d1b26c735c49764b1a56a31R39-R46). Only these fields will be picked up and forwarded.

> The question is if all business services involved in a business processes should know all the metadata keys that should be transported, or is it only consumers that need to act on it?

Every service involved in the process - also the ones not interested in the metadata - must know the fields. Otherwise we have to forward all headers which could leak sensitive information or break requests. Services that don't need the metadata must not implement anything. But the deployment must define `METADATA_FIELDS`. We defined that as environment variable instead of a regular configuration, so that uninterested services don't need to add properties to their `config.yml`. However, services that interrupt the process by storing data and using it later must add the context the data, even if they are not interested in it themselves.

> in the second case a nice addition could be an explicitly set parametrised behaviour (Implemented internally in the library) for only for the intersection part(commons keys), this way keeping the rest of the metadata could assured by the library, as opposed to leaving much freedom to users to build the desired metadata from the current and the DetachedMetadataContext and use it in the createContext which does not guarantee transporting everything to upstream services which could assume that downstream are just dumb transporters for the metadata.

I'm not sure if I understand this correctly. A service can't just create any context. The metadata context is something global for all services involved. So it is highly affected by configuration. The idea is like this example:

A service that is sending emails claims, that it supports to choose a template based on the metadata context.

When the feature should be used in a specific environment/cluster based on the _tenant_, all services must be configured with `METADATA_FIELDS=tenant-id,optional-other-fields` and the email service is additionally configured with `METADATA_FIELD_TEMPLATE_HINT=`tenant-id`. `METADATA_FIELD_TEMPLATE_HINT` should be part of its deployment documentation!

When choosing the template, the email service get's the hint via `MetadataContext.current().valuesByKeyFromEnvironment("METADATA_FIELD_TEMPLATE_HINT")` and will receive the `tenant-id` from the context.

The same applies for a producer. It can't just put anything to the context but must have in its own business context a property that is mapped by configuration to a specific metadata field. In some cases you may also have no explicit service setting the meta data. A tenant or otherwise specific use of a service could just get the header from an Ingress proxy based on the Domain where the service is exposed. In such case, no developer has to set fields of the context in their service.

sda-commons-server-dropwizard/src/main/java/org/sdase/commons/server/dropwizard/metadata/MetadataContextFilter.java
JoergSiebahn: Hmmm, either we jump out with an exception here and the context is not used anymore or `set` in the next line will overwrite it. I'm not sure how expensive thread local operations are.

sda-commons-server-dropwizard/src/main/java/org/sdase/commons/server/dropwizard/metadata/MetadataContext.java
saadmar: Okay, I got you, and it is logical and can be already seen in the implementation and the concern should be also fixed by the proposed `merge`. 
Just that if the process is interrupted (metadata saved to a database when `METADATA_FIELDS=key-1,key-2`) then at a future state, if a new configuration is introduced that adds a new `key-3` which could be not so relevant to the current service itself but to the next one, configuring all services with `METADATA_FIELDS=key-1,key-2,key-3` should still get all new values and `key-3` that the new service will start to receive should not be lost when restoring the context.
But I guess services interacting with the context should also be aware of such case and should be able to handle this themselves with the helpers, even if the behaviour is not imposed by the library.

JoergSiebahn: I hope these additions will help:

- [`isEffectivelyEmpty()`](https://github.com/SDA-SE/sda-dropwizard-commons/pull/2115/commits/cceef85599a8ccb13c638766c8d5e38300d246fe)
- [`mergeContext(newContextData, mergeStrategy)`](https://github.com/SDA-SE/sda-dropwizard-commons/pull/2115/commits/5bcbafc6d3cc224e2488122036cd900cac67d458)

christopher-cudennec (approved):

PR #2122, opened 2023-02-13
Send metadata context with Kafka producers

JoergSiebahn:


christopher-cudennec (commented):
Using Kafka's interceptor probably is the cleanest approach.

When I first looked at it I thought we would change our custom [KafkaMessageProducer](https://github.com/SDA-SE/sda-dropwizard-commons/blob/feat/v3/sda-commons-server-kafka/src/main/java/org/sdase/commons/server/kafka/producer/KafkaMessageProducer.java#L32). Would that be easier?

JoergSiebahn:
> When I first looked at it I thought we would change our custom [KafkaMessageProducer](https://github.com/SDA-SE/sda-dropwizard-commons/blob/feat/v3/sda-commons-server-kafka/src/main/java/org/sdase/commons/server/kafka/producer/KafkaMessageProducer.java#L32). Would that be easier?

I tried that first. For me it was quite unclear, what I'm doing there. The downside of all this approaches is, that you must come across our implementation or do something yourself. Implementing it in our code directly would make it impossible to use it elsewhere. The interceptor may be added to any producer. So if we implement it in a producer, it must be some kind of a wrapper that can be wrapped around anywhere. I threw code from here to there a few times but it did not feel good. But I'll give it another try because the current configuration is also not clean code.

JoergSiebahn:
Discussed with @christopher-cudennec that #2125 is the better approach.

PR #2125, opened 2023-02-14
feat(server-kafka): keep metadata context when producing Kafka messages

JoergSiebahn:
BREAKING CHANGE: All `KafkaBundle.createProducer` methods return a `Producer` instead of its implementation `KafkaProducer` now. The public API of the `KafkaProducer` is exactly the same as in the interface.

### DIFF
--- a/sda-commons-server-kafka/src/main/java/org/sdase/commons/server/kafka/KafkaBundle.java
+++ b/sda-commons-server-kafka/src/main/java/org/sdase/commons/server/kafka/KafkaBundle.java
@@ -12,15 +12,18 @@ import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
+import java.util.Set;
 import java.util.function.Function;
 import javax.validation.constraints.NotNull;
 import org.apache.kafka.clients.consumer.KafkaConsumer;
 import org.apache.kafka.clients.producer.KafkaProducer;
+import org.apache.kafka.clients.producer.Producer;
 import org.apache.kafka.common.Metric;
 import org.apache.kafka.common.MetricName;
 import org.apache.kafka.common.errors.InterruptException;
 import org.apache.kafka.common.serialization.Deserializer;
 import org.apache.kafka.common.serialization.Serializer;
+import org.sdase.commons.server.dropwizard.metadata.MetadataContext;
 import org.sdase.commons.server.kafka.builder.MessageListenerRegistration;
 import org.sdase.commons.server.kafka.builder.ProducerRegistration;
 import org.sdase.commons.server.kafka.config.ConsumerConfig;
@@ -33,6 +36,7 @@ import org.sdase.commons.server.kafka.health.ExternalKafkaHealthCheck;
 import org.sdase.commons.server.kafka.health.KafkaHealthCheck;
 import org.sdase.commons.server.kafka.producer.KafkaMessageProducer;
 import org.sdase.commons.server.kafka.producer.MessageProducer;
+import org.sdase.commons.server.kafka.producer.MetadataContextAwareKafkaProducer;
 import org.sdase.commons.server.kafka.prometheus.ConsumerTopicMessageHistogram;
 import org.sdase.commons.server.kafka.prometheus.KafkaConsumerMetrics;
 import org.sdase.commons.server.kafka.prometheus.ProducerTopicMessageCounter;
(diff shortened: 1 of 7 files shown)


sda-commons-server-kafka/src/main/java/org/sdase/commons/server/kafka/KafkaBundle.java
```diff
@@ -279,15 +282,18 @@ public <K, V> KafkaConsumer<K, V> createConsumer(
...
 
     if (producerConfig != null) {
       producerProperties.putAll(producerConfig.getConfig());
     }
 
-    return new KafkaProducer<>(producerProperties, keySerializer, valueSerializer);
+    KafkaProducer<K, V> kafkaProducer =
+        new KafkaProducer<>(producerProperties, keySerializer, valueSerializer);
+
+    return new MetadataContextAwareKafkaProducer<>(kafkaProducer, MetadataContext.metadataFields());
```
christopher-cudennec, line 296: Only wrap if any metadata fields were configured?

sda-commons-server-kafka/src/main/java/org/sdase/commons/server/kafka/KafkaBundle.java
```diff
@@ -279,15 +282,18 @@ public <K, V> KafkaConsumer<K, V> createConsumer(
    * @param <V> Value object type
    * @return a new kafka producer
    */
-  public <K, V> KafkaProducer<K, V> createProducer(
+  public <K, V> Producer<K, V> createProducer(
```
christopher-cudennec, line 285: Sorry, can't mark the right line here: Can you fix the documentation in L271? It says "Creates a new Kafka **Consumer**..."

sda-commons-server-kafka/src/main/java/org/sdase/commons/server/kafka/producer/MetadataContextAwareKafkaProducer.java
```diff
@@ -0,0 +1,145 @@
...
+
+  private ProducerRecord<K, V> withMetadataContext(ProducerRecord<K, V> producerRecord) {
+    if (metadataFields.isEmpty()) {
+      return producerRecord;
+    }
+    var headers = new RecordHeaders(producerRecord.headers());
+    MetadataContext metadataContext = MetadataContext.current();
+    for (String metadataField : metadataFields) {
+      List<String> valuesByKey = metadataContext.valuesByKey(metadataField);
+      if (valuesByKey == null) {
```
christopher-cudennec, line 127: What about `valuesByKey.isEmpty()`?
Do we want to support that case at all? We could also change the API for the metadata context and only store keys that have at least one value.

sda-commons-server-kafka/src/test/java/org/sdase/commons/server/kafka/KafkaBundleMetadataContextProducerIntegrationTest.java
```diff
@@ -0,0 +1,191 @@
...
+  @BeforeEach
+  @AfterEach
+  void clearMetadata() {
+    MetadataContext.createContext(new DetachedMetadataContext());
+  }
+
+  @Test
+  void shouldSendMetadataContext() {
+    DetachedMetadataContext given = new DetachedMetadataContext();
+    given.put("tenant-id", List.of("t-1"));
```
christopher-cudennec, line 87: Can we provide a fluent syntax to our clients?
ie. `new DetachedMetadataContext().put("foo", "bar")`

This example could make use of a varargs method. Would make it easier to use.

sda-commons-server-kafka/src/main/java/org/sdase/commons/server/kafka/producer/MetadataContextAwareKafkaProducer.java
JoergSiebahn: > Do we want to support that case at all? We could also change the API for the metadata context and only store keys that have at least one value.

Better safe than sorry. In the end it is an interface that may be implemented different.

> What about valuesByKey.isEmpty()?

That's just to avoid NPEs. Is it so much more expensive to iterate over an empty stream than checking for empty? In the current implementation, the code has less branches is is imho easier to read.

sda-commons-server-kafka/src/test/java/org/sdase/commons/server/kafka/KafkaBundleMetadataContextProducerIntegrationTest.java
JoergSiebahn: We extend a `LinkedHashMap` here. And that's the API of `Map`. Imho, we loose a lot of other benefits if we implement all that on our own.

sda-commons-server-kafka/src/main/java/org/sdase/commons/server/kafka/KafkaBundle.java
JoergSiebahn: Documentation updated

sda-commons-server-kafka/src/main/java/org/sdase/commons/server/kafka/KafkaBundle.java
JoergSiebahn: Updated.

christopher-cudennec (approved):

PR #2206, opened 2023-03-14
docs: Update catalog-info.yaml

JLeine:
docs: adding medium blog and removing redundant GH Link

### DIFF
--- a/catalog-info.yaml
+++ b/catalog-info.yaml
@@ -15,9 +15,8 @@ metadata:
     SDA Dropwizard Commons is a set of libraries to bootstrap services easily that follow
     the patterns and specifications promoted by the SDA SE.
   links:
-    - url: https://github.com/SDA-SE/sda-dropwizard-commons
-      title: GitHub
-      icon: code
+    - url: https://medium.com/sda-se/sda-dropwizard-commons-v3-448302578112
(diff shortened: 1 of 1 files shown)


catalog-info.yaml
```diff
@@ -15,9 +15,8 @@ metadata:
     SDA Dropwizard Commons is a set of libraries to bootstrap services easily that follow
     the patterns and specifications promoted by the SDA SE.
   links:
-    - url: https://github.com/SDA-SE/sda-dropwizard-commons
-      title: GitHub
-      icon: code
+    - url: https://medium.com/sda-se/sda-dropwizard-commons-v3-448302578112
+      title: SDA Dropwizard Commons v3
```
JoergSiebahn, line 19: Do we need this information in the internal backstage? And if so, why don't we need [this article](https://medium.com/sda-se/how-we-released-our-first-open-source-project-9ee0c9bdf720)?

christopher-cudennec: I don't think the article is crucial... but I'd say it's interesting/helpful. 😃 

I tried to find out if we can just all our articles using the "dropwizard" tag but Medium does not seem to support that.

> And if so, why don't we need this article?

I don't mind linking both. 😄

JoergSiebahn (approved):

PR #2233, opened 2023-03-22
docs: replace deprecated OTEL_EXPORTER_JAEGER_ENDPOINT info with OTEL…

mathieu-keller:
…_EXPORTER_OTLP_ENDPOINT

the env var OTEL_EXPORTER_JAEGER_ENDPOINT is deprecated and will be removed on July 2023 see: https://opentelemetry.io/docs/reference/specification/sdk-environment-variables/#jaeger-exporter
Instead we should use OTLE_EXPORTER_OTLP_ENDPOINT see: https://opentelemetry.io/docs/reference/specification/protocol/exporter/#configuration-options


mathieu-keller:
No't working with our current Jaeger setup

PR #2248, opened 2023-03-30
feat(client-jersey): Remove dependency to sda-commons-server-opentele…

christopher-cudennec:
…metry

This change prevents that applications that don't use our "sda-commons-starter" activate collection of Telemetry collection.

## Why was OTEL activated before?

sda-commons-server-opentelemetry defines a dependency to `io.opentelemetry:opentelemetry-sdk-extension-autoconfigure`. This causes `GlobalOpenTelemetry` to return an active instance of `OpenTelemetry` instead of the `noop` implementation.

## What is the result of this change?

* Applications that use "sda-commons-starter" will still get an active `OpenTelemetry` instance
* Applications that only use the jersey-client bundle or other bundles will not pull in "sda-commons-server-opentelemetry" anymore - thus use the `noop` implementation of `OpenTelemetry`

## BREAKING CHANGE?

Strictly speaking this is a breaking change because I changed the dependencies and moved one class to "sda-commons-client-jersey". As we had several complaints in the past: Should we make this a BREAKING CHANGE commit?

### DIFF
--- a/sda-commons-server-auth/src/main/java/org/sdase/commons/server/auth/AuthBundle.java
+++ b/sda-commons-server-auth/src/main/java/org/sdase/commons/server/auth/AuthBundle.java
@@ -3,11 +3,13 @@ package org.sdase.commons.server.auth;
 import io.dropwizard.Configuration;
 import io.dropwizard.ConfiguredBundle;
 import io.dropwizard.auth.AuthDynamicFeature;
+import io.dropwizard.client.HttpClientBuilder;
 import io.dropwizard.client.JerseyClientBuilder;
 import io.dropwizard.setup.Bootstrap;
 import io.dropwizard.setup.Environment;
 import io.opentelemetry.api.GlobalOpenTelemetry;
 import io.opentelemetry.api.OpenTelemetry;
+import io.opentelemetry.instrumentation.apachehttpclient.v4_3.ApacheHttpClientTelemetry;
 import java.net.ProxySelector;
 import java.net.URI;
 import java.net.URISyntaxException;
@@ -30,7 +32,6 @@ import org.sdase.commons.server.auth.key.PublicKeyLoader;
 import org.sdase.commons.server.auth.service.AuthService;
 import org.sdase.commons.server.auth.service.JwtAuthenticator;
 import org.sdase.commons.server.auth.service.TokenAuthorizer;
-import org.sdase.commons.server.opentelemetry.client.TracedHttpClientInitialBuilder;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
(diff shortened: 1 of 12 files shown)


JoergSiebahn:
> will not pull in "sda-commons-server-opentelemetry" anymore - thus use the noop implementation of OpenTelemetry

We should add at test for that.

> Strictly speaking this is a breaking change because I changed the dependencies and moved one class to "sda-commons-client-jersey". As we had several complaints in the past: Should we make this a BREAKING CHANGE commit?

Would you then leave or for the next major version or just create a new major release? Maybe we should just do major releases more often but with less impact?

INoTime:
Hi, when would the PR be approved so we can implement it?

JoergSiebahn:
> Strictly speaking this is a breaking change because I changed the dependencies and moved one class to "sda-commons-client-jersey". As we had several complaints in the past: Should we make this a BREAKING CHANGE commit?

I changed it to keep the class in the previous location and marked it as deprecated. A copy exists in client-jersey.
I'm not sure, if the change of the dependency hierarchy should be considered as breaking. If consumers don't use the starter which claims to be used as single entry point for the server features of the library, consumers should add explicit dependencies to the bundles they use.

I don't like that server-auth depends on client-jersey now. This means, that you get all the client dependencies with the starter bundle now. That was not the case previously. Options would be to duplicate the `TracedHttpClientInitialBuilder` again or to introduce a new module `sda-commons-client-???`. @christopher-cudennec what do you think?

sda-commons-server-s3/build.gradle
```diff
@@ -1,7 +1,9 @@
 dependencies {
   api project(':sda-commons-server-dropwizard')
   api project(':sda-commons-server-healthcheck')
+
   api 'com.amazonaws:aws-java-sdk-s3'
+  api 'io.opentelemetry:opentelemetry-api'
```
JoergSiebahn, line 6: Is this really needed?

sda-commons-server-auth/src/main/java/org/sdase/commons/server/opa/OpaBundle.java
```diff
@@ -175,7 +176,12 @@ private Client createClient(
 
     JerseyClientBuilder jerseyClientBuilder = new JerseyClientBuilder(environment);
     jerseyClientBuilder.setApacheHttpClientBuilder(
-        new TracedHttpClientInitialBuilder(environment).usingTelemetryInstance(openTelemetry));
+        new HttpClientBuilder(environment) {
```
JoergSiebahn, line 179: Needs a test to ensure, Noop-Tracer is used.

sda-commons-server-auth/src/main/java/org/sdase/commons/server/auth/AuthBundle.java
```diff
@@ -112,7 +113,12 @@ private Client createKeyLoaderClient(
     JerseyClientBuilder jerseyClientBuilder = new JerseyClientBuilder(environment);
     // should be set as soon as creating the builder
     jerseyClientBuilder.setApacheHttpClientBuilder(
-        new TracedHttpClientInitialBuilder(environment).usingTelemetryInstance(openTelemetry));
+        new HttpClientBuilder(environment) {
```
JoergSiebahn, line 116: Needs a test to ensure, Noop-Tracer is used.

christopher-cudennec (commented):
Thanks for your changes! Looks good to me! 👍

elbiocaetano (approved):

PR #2256, opened 2023-04-03
fix: logging trace id

elbiocaetano:


sda-commons-server-trace/build.gradle
```diff
@@ -1,6 +1,7 @@
 dependencies {
   api project(':sda-commons-server-dropwizard')
   api project(':sda-commons-shared-tracing')
+  api project (':sda-commons-server-opentelemetry')
```
JoergSiebahn, line 4: We should not spread this dependency into more modules, see https://github.com/SDA-SE/sda-dropwizard-commons/pull/2248

sda-commons-server-trace/src/main/java/org/sdase/commons/server/trace/filter/TraceTokenServerFilter.java
```diff
@@ -63,6 +63,7 @@ private Optional<String> extractTokenFromRequest(ContainerRequestContext request
   private void addTokenToMdc(String token) {
     if (MDC.getMDCAdapter() != null) {
       MDC.put(RequestTracing.TOKEN_MDC_KEY, token);
+      MDC.put("TraceID", token);
```
JoergSiebahn, line 66: I would add inline comments here to make clear what Trace Token/ID is from which context. Could we name it `OtelTraceId` or is the name important for matching in Grafana?

PR #2280, opened 2023-04-12
feat(server-kafka): implement send with callback for kafka producers

eikenilsknopp:


sda-commons-server-kafka/src/main/java/org/sdase/commons/server/kafka/KafkaBundle.java
```diff
@@ -193,7 +197,17 @@ public <K, V> MessageProducer<K, V> registerProducer(ProducerRegistration<K, V>
...
+        @Override
+        public Future<RecordMetadata> send(K key, V value, Headers headers) {
+          return null;
+        }
+
+        @Override
+        public Future<RecordMetadata> send(K key, V value, Headers headers, Callback callback) {
+          return null;
+        }
+      };
```
eikenilsknopp, line 210: It's no longer a functional interface since we now have 2 non-default abstract methods :/

JoergSiebahn: Hmmm, isn't it a breaking change then?

eikenilsknopp: Hm. Strictly speaking it would be if someone is either implementing the interface in their own class or if they are declaring a lambda, both is not really intended in my opinion. We intend that a user of the library will use the `KafkaMessageProducer`, dont we?

sda-commons-server-kafka/src/main/java/org/sdase/commons/server/kafka/producer/MessageProducer.java
```diff
@@ -37,6 +38,37 @@ default Future<RecordMetadata> send(K key, V value) {
    */
   Future<RecordMetadata> send(K key, V value, Headers headers);
```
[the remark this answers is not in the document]
eikenilsknopp: Yeah that would work, sounds good

sda-commons-server-kafka/src/main/java/org/sdase/commons/server/kafka/KafkaBundle.java
JoergSiebahn: > We intend that a user of the library will use the KafkaMessageProducer, dont we?

The [bundle returns the interface](https://github.com/SDA-SE/sda-dropwizard-commons/blob/master/sda-commons-server-kafka/src/main/java/org/sdase/commons/server/kafka/KafkaBundle.java#L189). So the interface is what consumers use in their classes as dependency. To unit test the classes, they need an instance of the interface. When using Mockito, there is no issue. But in some cases it may be easier to „mock“ desired behavior by providing a lambda to the instance in test.

eikenilsknopp: I updated the PR and added a breaking change message

elbiocaetano (dismissed):

JoergSiebahn (approved):

PR #2288, opened 2023-04-13
fix(mongodb): derive ssl usage from connection string

JoergSiebahn:


sda-commons-server-spring-data-mongo/src/main/java/org/sdase/commons/server/spring/data/mongo/MongoConfiguration.java
```diff
@@ -115,7 +117,7 @@ public MongoConfiguration setHosts(String hosts) {
   @Deprecated(forRemoval = true)
   public String getDatabase() {
     if (StringUtils.isBlank(database) && StringUtils.isNotBlank(connectionString)) {
-      return String.join(",", new ConnectionString(connectionString).getDatabase());
+      return new ConnectionString(connectionString).getDatabase();
```
christopher-cudennec, line 120: If creating the `ConnectionString` is expensive, we could also store it as a field of this class.

sda-commons-server-spring-data-mongo/src/test/java/org/sdase/commons/server/spring/data/mongo/MongoConfigurationTest.java
```diff
@@ -80,4 +80,32 @@ void shouldNotBeValidIfOnlyDatabaseWasSet() {
     MongoConfiguration config = new MongoConfiguration().setDatabase("records");
     assertThat(config.isValid()).isFalse();
   }
+
```
christopher-cudennec, line 83: We could also add a test without `ssl` or `tls` in the connection string to ensure that `isUseSsl` is `false`.

sda-commons-server-spring-data-mongo/src/test/java/org/sdase/commons/server/spring/data/mongo/SpringDataMongoBundleSSLConfigurationIT.java
```diff
@@ -44,7 +45,7 @@ class SpringDataMongoBundleSSLConfigurationIT {
           null,
           randomPorts(),
           config("springDataMongo.connectionString", mongo::getConnectionString),
-          config("springDataMongo.useSsl", "false"),
+          config("springDataMongo.useSsl", "true"),
```
christopher-cudennec, line 48: This value does not match the use case, does it? The app extension's name is `DW_WITHOUT_SSL` but here you set the value to `true`.

I would prefer to not set this deprecated property anymore at all. If the connection string is configured it should override all other settings IMHO.

JoergSiebahn: > If the connection string is configured it should override all other settings IMHO.

That's why I changed the test. But I'm not sure if this assumption is correct in all cases as I don't know if the mongo client will upgrade to SSL automatically, if it is not set in the connection string, but the server only offers SSL.

sda-commons-server-spring-data-mongo/src/main/java/org/sdase/commons/server/spring/data/mongo/MongoConfiguration.java
JoergSiebahn: That only happens on startup. Not sure if it's worth it.

sda-commons-server-spring-data-mongo/src/test/java/org/sdase/commons/server/spring/data/mongo/MongoConfigurationTest.java
JoergSiebahn: That is tested [here](https://github.com/SDA-SE/sda-dropwizard-commons/pull/2288#discussion_r1166287322), isn't it?.

PR #2315, opened 2023-04-24
fix(server-dropwizard): log level not limited by appender threshold

sliesche:
Actual log output level is depended on "logging.level" and "logging.appenders.threshold". This fix will set appender threshold to "TRACE". This will allow full control of log output using logging.level. All logging default values can be overwritten.


christopher-cudennec (commented):
Can we provide a test that sets system property `dw.logging.level` to `DEBUG` when starting an application to test if the changes work?

christopher-cudennec (approved):
👑

PR #2394, opened 2023-05-30
fix(server-kafka): support case insensitive headers when propagating metadata

saadmar:
Provides same level of support as in [REST servers](https://github.com/SDA-SE/sda-dropwizard-commons/blob/d3fd0757211134187de43590e1e838117452f6c2/sda-commons-starter/src/test/java/org/sdase/commons/starter/MetadataContextIntegrationTest.java#L47).
Currently Kafka headers are checked for exact equality, which results in some metadata being ignored and not added to the current context.


saadmar:
> Do we need to consider the headers/values from REST endpoints as well? @JoergSiebahn : Can I get your opinion here?

From what I saw in [this test](https://github.com/SDA-SE/sda-dropwizard-commons/blob/d3fd0757211134187de43590e1e838117452f6c2/sda-commons-starter/src/test/java/org/sdase/commons/starter/MetadataContextIntegrationTest.java#L47), it is already considered

JoergSiebahn:
Afaik, the rest headers are case insensitive by design.

eikenilsknopp:
> Afaik, the rest headers are case insensitive by design.

Yes, see the [specification for message headers](https://www.w3.org/Protocols/rfc2616/rfc2616-sec4.html#sec4.2): 
> Field names are case-insensitive

But is it necessary to add a new class, couldn’t we just make sure to call `toLowerCase()` when creating the headers?

saadmar:
> But is it necessary to add a new class, couldn’t we just make sure to call `toLowerCase()` when creating the headers?

No, that is just what I did (I can delete the extra class, I just wanted to avoid changes [here](https://github.com/SDA-SE/sda-dropwizard-commons/pull/2394/files#diff-edaaf46805af51af53409732edbf29192d3a24ce3364b7b92b9627318fcaabe2R133)), the only important thing is that we need convert all headers to lowercase before matching also.

sda-commons-server-kafka/src/main/java/org/sdase/commons/server/kafka/consumer/strategies/MessageListenerStrategy.java
```diff
@@ -143,6 +141,14 @@ private MetadataContextCloseable createMetadataContext(ConsumerRecord<K, V> cons
     return MetadataContext.createCloseableContext(newContext);
   }
 
+  private Headers normalizeHeaders(Headers headers) {
+    return new RecordHeaders(
+        Arrays.stream(headers.toArray())
+            .filter(Objects::nonNull)
+            .map(CaseInsensitiveHeader::new)
+            .collect(Collectors.toList()));
+  }
```
JoergSiebahn, line 150: Could we go for a simpler approach like the following instead of introducing a new class?

```suggestion
  private Headers normalizeHeaders(Headers headers) {
    return new RecordHeaders(
        Arrays.stream(headers.toArray())
            .filter(Objects::nonNull)
            .map(h -> new RecordHeader(StringUtils.lowerCase(h.key()), h.value()))
            .collect(Collectors.toList()));
  }
```

saadmar: done

sda-commons-server-kafka/src/main/java/org/sdase/commons/server/kafka/consumer/strategies/MessageListenerStrategy.java
```diff
@@ -2,11 +2,7 @@
 
 import static java.nio.charset.StandardCharsets.UTF_8;
 
-import java.util.Collections;
-import java.util.HashMap;
-import java.util.Map;
-import java.util.Optional;
-import java.util.Set;
+import java.util.*;
```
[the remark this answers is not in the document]
saadmar: Done

JoergSiebahn:
Thanks @saadmar ❤️

PR #2456, opened 2023-06-23
Reduce number of tests that cause reports about changed tests in PRs

JoergSiebahn:
image


christopher-cudennec (approved):
That behaviour has also bugged me for some time 👍

PR #2484, opened 2023-06-29
fix(prometheus): remove registry

sliesche:
fixes metrics test issues 
see [https://github.com/SDA-SE/partner-ods/pull/747](https://github.com/SDA-SE/partner-ods/pull/747)

### DIFF
--- a/sda-commons-server-prometheus/src/test/java/org/sdase/commons/server/prometheus/PrometheusBundleTest.java
+++ b/sda-commons-server-prometheus/src/test/java/org/sdase/commons/server/prometheus/PrometheusBundleTest.java
@@ -18,11 +18,14 @@ import javax.ws.rs.client.Invocation;
 import javax.ws.rs.core.Response;
 import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.extension.ExtendWith;
 import org.junit.jupiter.api.extension.RegisterExtension;
+import org.sdase.commons.server.prometheus.helper.MicrometerTestExtension;
 import org.sdase.commons.server.prometheus.test.PrometheusTestApplication;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+@ExtendWith(MicrometerTestExtension.class)
 class PrometheusBundleTest {
   private static final Logger LOGGER = LoggerFactory.getLogger(PrometheusBundleTest.class);
 
@@ -199,6 +202,8 @@ class PrometheusBundleTest {
     assertThat(sampleList.get(0).labelValues).contains("testTagValue");
 
     assertThat(sampleList.get(0).value).isEqualTo(2);
+
+    Metrics.globalRegistry.close();
   }
 
   private Invocation.Builder prepareResourceRequest() {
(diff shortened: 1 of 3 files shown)


christopher-cudennec (commented):
Looks good - is it possible to test/verify the behaviour?

sliesche:
> Looks good - is it possible to test/verify the behaviour?

@christopher-cudennec 
I added a test that checks if the Metrics.globalRegistry.getRegistries() are empty after shutdown.

sda-commons-server-prometheus/src/test/java/org/sdase/commons/server/prometheus/PrometheusBundleTest.java
```diff
@@ -199,6 +202,8 @@ void micrometerMetricsAvailableInPrometheus() {
     assertThat(sampleList.get(0).labelValues).contains("testTagValue");
 
     assertThat(sampleList.get(0).value).isEqualTo(2);
+
+    Metrics.globalRegistry.close();
```
christopher-cudennec, line 206: Do we need to do this in a `finally` block or `afterEach`?

JoergSiebahn, line 206: Would it also be possible to close the registry in the Bundle? Or is [`removeRegistry`](https://github.com/SDA-SE/sda-dropwizard-commons/pull/2484/files#diff-c7d888d64ab52c9b42637100cad05ce6be7c9d9b6d9e81811ad2317c9e8e5f23R94) doing exactly that?
Would it make sense (and is it possible) to also close (or reset) the `CollectorRegistry` on shutdown?

sliesche: No closing, clearing and removing have different effects:

- Metrics.globalRegistry.clear();
   Clears all meters (e.g. Counter) but keeps them. registries unaffected

- Metrics.globalRegistry.close();
   Keeps registries but does not accept new meters and does not publish new values

- Metrics.removeRegistry(meterRegistry);
   actually removes the registry

Currently doing some tests to see the effect of doing both removingRegistry and closing

sliesche: Although it should be possible to use `CollectorRegistry.defaultRegistry.clear();` the side effects are hard to predict. We are using CollectorRegistry.defaultRegistry.unregister(collector) in multiple places. If we clear the registry first and then unregister there will be a null pointer exceptions.

Given the impact of this issue i suggest we analyse the clearing of the prometheus registry separately.

sliesche: I decided to use `Metrics.globalRegistry.clear()` in `afterEach` see comments above. For some tests it might be necessary to use removeRegistry

JoergSiebahn: But if we have to clear it here in tests, consumers must do that as well in their tests, don't they?

sliesche: I thinks that depends on the test.

If the metrics tests are in different classes the shutdown behaviour should kick in and clearing should be unnecessary. If the same metric should be tested multiple times in the same class then registry needs to be cleared in between the tests.

JoergSiebahn (approved):

PR #2500, opened 2023-07-04
Exclude weld-probe-core from weld-bundle

chrste90:
The probe bundle is responsible for most vulnerabilities in the weld bundle. However, it is not needed at runtime as it is a developer tool.

See
https://docs.jboss.org/weld/reference/3.1.9.Final/en-US/html/developmentmode.html
https://github.com/weld/core/pull/2792

weld-probe-core is also responsible for multiple breaking builds inside SIGNAL-IDUNA due to its multiple cves.

### DIFF
--- a/sda-commons-server-weld/build.gradle
+++ b/sda-commons-server-weld/build.gradle
@@ -31,6 +31,12 @@ dependencies {
      * We keep the version managed by Dropwizard.
      */
     exclude group: 'org.jboss.spec.javax.annotation', module: 'jboss-annotations-api_1.3_spec'
+
+    /**
+     * The probe bundle is responsible for most vulnerabilities in the weld bundle.
+     * However, it is not needed at runtime as it is a developer tool.
+     */
+    exclude group: 'org.jboss.weld.probe', module: 'weld-probe-core'
   }
   api 'org.jboss.weld.servlet:weld-servlet-core', {
     /**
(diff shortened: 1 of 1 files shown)


JoergSiebahn:
Hi @chrste90, thanks for looking into the weld bundle and providing a fix for vulnerabilities.

Just out of curiosity: Can you post links to affected CVEs? We recently enabled a scanner and got no information about this vulnerabilities.

One more important thing: Can you rephrase the first line of the commit message to `fix(server-weld): exclude weld-probe-core from weld-bundle to avoid vulnerabilities`? Without [the prefix](https://github.com/SDA-SE/sda-dropwizard-commons/blob/master/CONTRIBUTING.md#changelog-and-versioning), the change will not create a new release.

chrste90:
Hi @JoergSiebahn ,

i updated the commit message. 

Regarding the CVEs, the two most important (ranked as high) are:

weld-probe-core-3.1.9.Final.jar: probe.js | CVE-2017-18214
weld-probe-core-3.1.9.Final.jar: probe.js | CVE-2022-24785

JoergSiebahn (approved):

JoergSiebahn:
Thanks a lot for your contribution @chrste90. It will be merged once the build passes again. I'll prepare a PR to backport the fix to the 2.x.x release.

PR #2503, opened 2023-07-04
Another try to update snakeyaml

JoergSiebahn:
- [x] AsyncApi generation fails and produces an invalid AsyncAPI on embedding definitions (failure in [JsonSchemaEmbedder](https://github.com/SDA-SE/sda-dropwizard-commons/blob/security/deps/snakeyaml-jackson/sda-commons-shared-asyncapi/src/main/java/org/sdase/commons/shared/asyncapi/internal/JsonSchemaEmbedder.java), visible in [JsonSchemaEmbedderTest](https://github.com/SDA-SE/sda-dropwizard-commons/blob/security/deps/snakeyaml-jackson/sda-commons-shared-asyncapi/src/test/java/org/sdase/commons/shared/asyncapi/internal/JsonSchemaEmbedderTest.java))
- [x] needs a check with the SNAPSHOT in a real consumer (withFailOnVersionConflict enabled) to verify that the dependency resolution is consistent not only within the library
- [x] seems like the new Jackson version deserializes `ZonedDateTime` differently. Our format results in offset `Z` and zone id `UTC` before. Now both is `Z`. That means the same but fails in tests when the Java value is compared with `equals`. As it will fail many (not ideal written) API tests, the upgrade could be considered as breaking change. ([new regression tests](https://github.com/SDA-SE/sda-dropwizard-commons/commit/49dfb94156f454dc47085dcae51f8e64217e4d79) worked in master, fail here, see [required change here](https://github.com/SDA-SE/sda-dropwizard-commons/pull/2503/commits/75ea75aec6b77a87837a0da8de476ba2333db2ea))
- [x] check if above mentioned change - affecting test assertions - affects other areas as well and should be declared as a breaking change
- [x] make it a breaking change

### DIFF
--- a/sda-commons-dependencies/build.gradle
+++ b/sda-commons-dependencies/build.gradle
@@ -74,8 +74,9 @@ dependencies {
     api "org.jboss.logging:jboss-logging:3.5.3.Final", {
       because "conflict between org.hibernate.validator:hibernate-validator and org.jboss.weld.se:weld-se-core"
     }
-    api "org.yaml:snakeyaml:1.33", {
+    api "org.yaml:snakeyaml:2.0", {
       because "conflict between com.github.ftrossbach:club-topicana-core and com.github.database-rider:rider-core and com.fasterxml.jackson.dataformat:jackson-dataformat-yaml"
+      because "vulnerability CVE-2022-1471 in 1.33 and below"
     }
     api "net.javacrumbs.json-unit:json-unit-core:$jsonUnitVersion", {
       because "conflict between com.github.tomakehurst:wiremock-jre8 and net.javacrumbs.json-unit:json-unit-assertj"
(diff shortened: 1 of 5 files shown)


sda-commons-dependencies/build.gradle
```diff
@@ -65,8 +74,9 @@ dependencies {
     api "org.jboss.logging:jboss-logging:3.5.3.Final", {
       because "conflict between org.hibernate.validator:hibernate-validator and org.jboss.weld.se:weld-se-core"
     }
-    api "org.yaml:snakeyaml:1.33", {
+    api "org.yaml:snakeyaml:2.0", {
       because "conflict between com.github.ftrossbach:club-topicana-core and com.github.database-rider:rider-core and com.fasterxml.jackson.dataformat:jackson-dataformat-yaml"
```
christopher-cudennec, line 78: This line still refers to "topicana" but I'm pretty sure we don't use it anymore. We could remove the comment and the version definition further down in another PR.

christopher-cudennec (approved):

PR #2534, opened 2023-07-14
feat(server-openapi): sort output

christopher-cudennec:
This should make the output more reliable and fix random failing tests for OpenAPI comparison because the order of attributes changed.


JoergSiebahn (approved):

JoergSiebahn:
> … and fix random failing tests …

The tests should not compare with expecting a specific order. But reviews should be easier to read.

christopher-cudennec:
> > … and fix random failing tests …
> 
> The tests should not compare with expecting a specific order. But reviews should be easier to read.

I think the tests don't fail but the output of the `openapi.yaml` still changes. This is at least confusing.

PR #2585, opened 2023-08-02
feat: add micrometer JVM and system metrics

frederic-huebner:


sda-commons-server-prometheus/src/main/java/org/sdase/commons/server/prometheus/PrometheusBundle.java
```diff
@@ -91,13 +95,22 @@ private void createPrometheusRegistry(Environment environment) {
 
     Metrics.addRegistry(meterRegistry);
 
+    // JVM and System Metrics
```
sliesche, line 98: binding JVM metrics to micrometer and exposing micrometer metrics through prometheus are logically separated one step is defining what goes in the other how it comes out.

my suggestion would be to put them in different methods

sliesche:
I noticed that some libraries put `deprecated` in the label(tag) name of an outdated metric
Could you try to add a `deprecated` label to the old jvm metrics?

e.g.
mappers.add(createMapperConfig("jvm.gc.*.count", "jvm_gc_total", "step", "deprecated"));

https://github.com/SDA-SE/sda-dropwizard-commons/blob/e4cfbee322b4aaa50e2bc493577bc07029ee357f/sda-commons-server-prometheus/src/main/java/org/sdase/commons/server/prometheus/PrometheusBundle.java#L127

sliesche (approved):

PR #2685, opened 2023-09-06
Collect configuration from environment variables dynamically

JoergSiebahn:
In Dropwizard it is not very comfortable to apply configuration if the `config.yml` does not define a specific environment variable for certain configuration options. Usually, we only prepare the most common settings to be configurable. But applications need to be deployed in different environments and sometimes unexpected configuration is needed.

So far the only option is to set forgotten configuration properties with JVM properties. There is only one environment variable for all JVM properties in typical dockerized applications. And it is mostly used do define technical configuration like resource settings or GC tuning. Mixing required application configuration in the same variable makes the deployment hardly readable and creates the risk, that important resource settings are overwritten when one needs to configure something not defined in the `config.yml`.

This PR adds a dynamic approach to define environment variables based on the structure of the actual configuration class. It is created as an idea and a starting point for discussion. It is not fully implemented yet. There are various types in the standard Dropwizard configuration classes that are not mapped yet. Also primitive types are not handled correctly yet.


sda-commons-server-dropwizard/src/main/java/org/sdase/commons/server/dropwizard/bundles/GenericEnvironmentMapper.java
```diff
@@ -0,0 +1,129 @@
...
+
+public class GenericEnvironmentMapper<T> implements ConfigurationSourceProvider {
+
+  private static final Logger LOG = LoggerFactory.getLogger(GenericEnvironmentMapper.class);
+
+  private final ConfigurationSourceProvider delegate;
+  private final Class<T> configurationClass;
+  // as in io.dropwizard.configuration.YamlConfigurationFactory
+  private final YAMLFactory yamlFactory = new YAMLFactory();
+  private final ObjectMapper mapper =
```
christopher-cudennec, line 29: I noticed during profiling that it's quite expensive to create new instances of the ObjectMapper. Would it be possible to re-use an existing one?

sda-commons-server-dropwizard/src/main/java/org/sdase/commons/server/dropwizard/bundles/JacksonTypeScanner.java
```diff
@@ -0,0 +1,113 @@
...
+
+    public String propertyTypeSimpleName() {
+      if (propertyType instanceof Class) {
+        return ((Class<?>) propertyType).getSimpleName();
+      }
+      return propertyType.getTypeName();
+    }
+
+    public String environmentVariableName() {
+      return jsonPathToProperty.stream().map(String::toUpperCase).collect(Collectors.joining("_"));
```
christopher-cudennec, line 110: Init value in constructor? The method seems to be invoked several times.

christopher-cudennec (commented):
First impression of your changes: I like the improvement. This would support the same way to set/override configuration values like Spring.

I must admit that I imagined the solution/implementation to be easier. But without really going hands-on/into the code I can't really say if there is an easier way to do what you're trying.

I wonder if there would be an advantage to use a `DW_` prefix for all environment variables that are meant as configuration overrides. That would match the Dropwizard approach somehow for system properties. But it would be different from the Spring approach. And I don't think there's a high probability that we pick up system environment variables as application configuration values.

sda-commons-server-dropwizard/src/main/java/org/sdase/commons/server/dropwizard/bundles/GenericEnvironmentMapper.java
JoergSiebahn: I don't think so. Dropwizard creates their own as well for parsing the config. And that is hidden in the config reader. I think it is intentional to not use the OM that is prepared for the REST API. I don't think it is an issue because it will only happen once for each app start.

sda-commons-server-dropwizard/src/main/java/org/sdase/commons/server/dropwizard/bundles/JacksonTypeScanner.java
JoergSiebahn: fixed

JoergSiebahn:
> I must admit that I imagined the solution/implementation to be easier.

Yes, I expected the same. Especially, I expected that producing the nested structure as in the scanner is something you get, maybe in another format, directly from Jackson. Also identifying the actual (Json) type of a property is kind of impossible. That is somehow hidden and not published in the serializers. Even if you get the serializer for a property (I found no way to identify it so far), you won't get information about the Json type.

sda-commons-server-dropwizard/src/main/java/org/sdase/commons/server/dropwizard/bundles/GenericLookupYamlConfigurationFactory.java
```diff
@@ -0,0 +1,105 @@
...
+            "{} of type {} picked from lookup.",
+            mappableField.getEnvironmentVariableName(),
+            mappableField.getPropertyTypeSimpleName());
+      } else {
+        LOG.info(
+            "{} of type {} not set.",
+            mappableField.getEnvironmentVariableName(),
+            mappableField.getPropertyTypeSimpleName());
+      }
+    }
```
JoergSiebahn, line 81: Maybe remove logging before merging, users can use the `config` command to get information.

sda-commons-starter/src/main/java/org/sdase/commons/starter/SdaPlatformBundle.java
```diff
@@ -47,6 +47,7 @@ public class SdaPlatformBundle<C extends Configuration> implements ConfiguredBun
   private final OpaBuilder<C> opaBundleBuilder;
   private final CorsBundle.FinalBuilder<C> corsBundleBuilder;
   private final OpenApiBundle.FinalBuilder openApiBundleBuilder;
+  private final ConfigurationSubstitutionBundle.Builder configurationSubstitutionBundleBuilder;
```
JoergSiebahn, line 50: This class might be reverted. Change is most likely not needed.

sda-commons-server-dropwizard/src/main/java/org/sdase/commons/server/dropwizard/bundles/scanner/MappableField.java
```diff
@@ -0,0 +1,117 @@
...
+import java.lang.reflect.Type;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+import java.util.stream.Collectors;
+
+public class MappableField {
```
christopher-cudennec, line 12: I wonder if it would make sense and be possible to make this class package private as it's only an implementation detail. We would need to move around some classes, I guess.

sda-commons-server-dropwizard/src/main/java/org/sdase/commons/server/dropwizard/bundles/scanner/MappableField.java
```diff
@@ -0,0 +1,117 @@
...
+   * @return the {@code MappableField}s that are available in the current context, may be empty.
+   *     <strong>Note that the type is just copied and might be wrong for the actual
+   *     property.</strong>
+   */
+  public List<MappableField> expand(Set<String> availableKeysInContext) {
+    if (availableKeysInContext.contains(getEnvironmentVariableName())) {
+      return List.of(new MappableField(this.jsonPathToProperty, this.propertyType));
+    }
+    if (getEnvironmentVariableName()
+        .contains(JacksonTypeScanner.MAP_STRING_KEY_NAME.toUpperCase())) {
```
christopher-cudennec, line 85: Why not store the result of `MAP_STRING_KEY_NAME.toUpperCase()` somewhere instead of doing it every time?

sda-commons-server-dropwizard/src/main/java/org/sdase/commons/server/dropwizard/bundles/scanner/MappableField.java
```diff
@@ -0,0 +1,117 @@
...
+          .collect(Collectors.toList());
+    }
+    return List.of();
+  }
+
+  private Pattern createExpandPattern(String environmentVariableName) {
+    return Pattern.compile(
+        "^"
+            + environmentVariableName.replace(
+                JacksonTypeScanner.MAP_STRING_KEY_NAME.toUpperCase(), "(.*)")
```
christopher-cudennec, line 114: one more place that invokes `MAP_STRING_KEY_NAME.toUpperCase()`

sda-commons-server-dropwizard/src/main/java/org/sdase/commons/server/dropwizard/bundles/scanner/MappableField.java
```diff
@@ -0,0 +1,117 @@
...
+                  for (int j = 0; j < newPath.size(); j++) {
+                    if (JacksonTypeScanner.MAP_STRING_KEY_NAME.equals(newPath.get(j))) {
+                      newPath.set(j, realKeyPart);
+                      break;
+                    }
+                  }
+                }
+                return new MappableField(newPath, newEnvironmentVariableName, getPropertyType());
+              })
+          .collect(Collectors.toList());
```
christopher-cudennec, line 105: It would be helpful to have an example in the documentation to easily understand what you're doing here. You could extract the block to a new method and add some documentation.

sda-commons-server-dropwizard/src/test/java/org/sdase/commons/server/dropwizard/bundles/scanner/JacksonTypeScannerTest.java
```diff
@@ -0,0 +1,197 @@
...
+        .contains("OPA_OPACLIENT_TLS_VERIFYHOSTNAME (boolean)")
+        .contains("OPA_OPACLIENT_TLS_CERTALIAS (String)")
+        .contains("OPA_OPACLIENT_TLS_VALIDKEYSTOREPASSWORD (boolean)")
+        .contains("OPA_OPACLIENT_TLS_VALIDTRUSTSTOREPASSWORD (boolean)")
+        .contains("OPA_DISABLEOPA (boolean)")
+        .contains("OPA_BASEURL (String)")
+        .contains("OPA_POLICYPACKAGE (String)")
+        .contains("KAFKA_DISABLED (boolean)")
+        .contains("KAFKA_CONFIG_<KEY> (Map)")
+        .contains("KAFKA_PRODUCERS_<KEY>_CLIENTID (String)")
```
christopher-cudennec, line 113: We need some documentation/instruction for users. According to the examples in the other test, the key is case sensitive while the rest of the env variable needs to be upper cased.

sda-commons-server-dropwizard/src/main/java/org/sdase/commons/server/dropwizard/bundles/GenericLookupYamlConfigurationFactory.java
```diff
@@ -0,0 +1,100 @@
...
+    return super.build(node, path);
+  }
+
+  private HashSet<String> findKeysInEnvironmentContext() {
+    // TODO this is probably too related to the given lookup so both may be a separate kind of
+    //      properties resolve mechanism just for this purpose
+    var contextKeys = new HashSet<>(System.getenv().keySet());
+    contextKeys.addAll(
+        System.getProperties().keySet().stream()
+            .map(Object::toString)
```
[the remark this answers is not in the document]
JoergSiebahn: `getProperties()` returns Object keys. But in the real world, I've never seen anything else than String - especially for the keys we are looking for. `toString` is just a shorter variant of class casting.

sda-commons-server-dropwizard/src/main/java/org/sdase/commons/server/dropwizard/bundles/scanner/JacksonTypeScanner.java
```diff
@@ -0,0 +1,140 @@
...
+      JavaType propertyType, List<String> mapRootPath) {
+    try {
+      var newRootPath = new ArrayList<>(mapRootPath);
+      newRootPath.add(MAP_STRING_KEY_NAME);
+      if (isTypedMap(propertyType, String.class, String.class)) {
+        return List.of(new MappableField(newRootPath, propertyType.getRawClass()));
+      }
+      if (isTypedMap(propertyType, String.class, Object.class)) {
+        return scan(newRootPath, propertyType.findTypeParameters(Map.class)[1]);
+      }
```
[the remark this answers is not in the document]
JoergSiebahn: Imho there is no value in such a log. You would inform about something that is likely not used at all and just not supported. Imho that would distract more people than it helps.

sda-commons-server-dropwizard/src/main/java/org/sdase/commons/server/dropwizard/bundles/GenericLookupConfigCommand.java
```diff
@@ -0,0 +1,66 @@
...
+          "  - MiB: mebibytes",
+          "  - GB: gigabytes",
+          "  - GiB: gibibytes",
+          "  - T: terabytes",
+          "  - TiB: tebibytes",
+          "  - PB: petabytes",
+          "  - PiB: pebibytes",
+          "",
+          "The type Map supports String values and uses <KEY> in the environment variable.",
+          "<KEY> must be changed to the desired key. <KE> may occur in other types to fill",
```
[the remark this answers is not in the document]
JoergSiebahn: fixed

sda-commons-server-dropwizard/src/main/java/org/sdase/commons/server/dropwizard/bundles/scanner/MappableField.java
JoergSiebahn: It is the return type of the `JacksonTypeScanner`'s public `scan(…)` method. I put both in a separate package, because they both work together quite independent of the implementation that applies the configuration somewhere. I was also thinking about removing any Dropwizard relation and add the [specific types](https://github.com/SDA-SE/sda-dropwizard-commons/pull/2685/files#diff-c77741edf0b134673367e86c14cd853540daaf3ea5fd8210366d6c416c507dd2R23-R24) in the constructor.

JoergSiebahn: Just saw, that I could make the constructor package private. That should clarify enough that it is used internally.

sda-commons-server-dropwizard/src/main/java/org/sdase/commons/server/dropwizard/bundles/GenericLookupConfigCommand.java
```diff
@@ -0,0 +1,66 @@
...
+          "  - MB: megabytes",
+          "  - MiB: mebibytes",
+          "  - GB: gigabytes",
+          "  - GiB: gibibytes",
+          "  - T: terabytes",
+          "  - TiB: tebibytes",
+          "  - PB: petabytes",
+          "  - PiB: pebibytes",
+          "",
+          "The type Map supports String values and uses <KEY> in the environment variable.",
```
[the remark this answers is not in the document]
JoergSiebahn: If such configuration exists, you will find the real world example in the appended list of configuration keys. That are the examples you can actually use.

sda-commons-server-dropwizard/src/main/java/org/sdase/commons/server/dropwizard/bundles/GenericLookupYamlConfigurationFactory.java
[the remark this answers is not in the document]
JoergSiebahn: System properties use `Properties` which defines `Object` for `key` and `value` (no generics in `Properties`). But the API of `System` regarding properties only deals with `String` on all methods but on `getProperties()`, which returns the bare `Properties`.

sda-commons-server-dropwizard/src/test/java/org/sdase/commons/server/dropwizard/bundles/scanner/JacksonTypeScannerTest.java
JoergSiebahn: Clarified [here](https://github.com/SDA-SE/sda-dropwizard-commons/pull/2685/commits/4ff51cfd8f9dfe112019b977544be1929db2c5fb).

sda-commons-server-dropwizard/src/main/java/org/sdase/commons/server/dropwizard/bundles/scanner/MappableField.java
JoergSiebahn: Introduced a second constant.

sda-commons-server-dropwizard/src/main/java/org/sdase/commons/server/dropwizard/bundles/scanner/MappableField.java
JoergSiebahn: Introduced a second constant.

sda-commons-server-dropwizard/src/main/java/org/sdase/commons/server/dropwizard/bundles/scanner/MappableField.java
JoergSiebahn: Extracted a method and added JavaDoc.

JoergSiebahn:
/closes #590 
Thanks @dhenneke for the investigation regarding config overrides long time ago.

JoergSiebahn:
<details>
  <summary>Tested with snapshot release in partner-ods locally:</summary>
<pre><code>
$ docker run partner-ods:test-config config     
WARNING: The requested image's platform (linux/amd64) does not match the detected host platform (linux/arm64/v8) and no specific platform was requested
Configuration can be manipulated dynamically from environment variables.
Some types use a special syntax for the value:
- Duration: '&lt;int>&lt;unit>' with unit as:
  - ns: nanoseconds
  - us: microseconds
  - ms: milliseconds
  - s: seconds
  - m: minutes
  - h: hours
  - d: days
- DataSize: &lt;int>&lt;unit>, with unit as
  - B: bytes
  - KB: kilobytes
  - KiB: kibibytes
  - MB: megabytes
  - MiB: mebibytes
  - GB: gigabytes
  - GiB: gibibytes
  - T: terabytes
  - TiB: tebibytes
  - PB: petabytes
  - PiB: pebibytes

The type Map supports String values and uses &lt;KEY> in the environment variable.
&lt;KEY> must be changed to the desired key of the Map using the character case as needed.
&lt;KEY> may occur in other types to fill a Map of well defined objects.

All supported environment variable keys:

ADMIN_HEALTHCHECKS_MAXTHREADS (int)
ADMIN_HEALTHCHECKS_MINTHREADS (int)
ADMIN_HEALTHCHECKS_SERVLETENABLED (boolean)
ADMIN_HEALTHCHECKS_WORKQUEUESIZE (int)
ADMIN_TASKS_PRINTSTACKTRACEONERROR (boolean)
HEALTH_DELAYEDSHUTDOWNHANDLERENABLED (boolean)
HEALTH_ENABLED (boolean)
HEALTH_INITIALOVERALLSTATE (boolean)
HEALTH_RESPONDER_CACHECONTROLENABLED (boolean)
HEALTH_RESPONDER_CACHECONTROLVALUE (String)
HEALTH_SHUTDOWNWAITPERIOD (Duration)
LOGGING_LEVEL (String)
METRICS_FREQUENCY (Duration)
METRICS_REPORTONSTOP (boolean)
SERVER_ADMINCONTEXTPATH (String)
SERVER_ADMINMAXTHREADS (int)
SERVER_ADMINMINTHREADS (int)
SERVER_APPLICATIONCONTEXTPATH (String)
SERVER_DUMPAFTERSTART (boolean)
SERVER_DUMPBEFORESTOP (boolean)
SERVER_ENABLETHREADNAMEFILTER (boolean)
SERVER_GROUP (String)
SERVER_GZIP_BUFFERSIZE (DataSize)
SERVER_GZIP_DEFLATECOMPRESSIONLEVEL (int)
SERVER_GZIP_ENABLED (boolean)
SERVER_GZIP_GZIPCOMPATIBLEINFLATION (boolean)
SERVER_GZIP_MINIMUMENTITYSIZE (DataSize)
SERVER_GZIP_SYNCFLUSH (boolean)
SERVER_IDLETHREADTIMEOUT (Duration)
SERVER_MAXQUEUEDREQUESTS (int)
SERVER_MAXTHREADS (int)
SERVER_METRICPREFIX (String)
SERVER_MINTHREADS (int)
SERVER_RESPONSEMETEREDLEVEL (enum ResponseMeteredLevel)
SERVER_ROOTPATH (String)
SERVER_SERVERPUSH_ASSOCIATEPERIOD (Duration)
SERVER_SERVERPUSH_ENABLED (boolean)
SERVER_SERVERPUSH_MAXASSOCIATIONS (int)
SERVER_SHUTDOWNGRACEPERIOD (Duration)
SERVER_UMASK (String)
SERVER_USER (String)
</code></pre>
</details>

christopher-cudennec (approved):

PR #2695, opened 2023-09-08
text(sda-commons-server-spring-data-mongo): test against MongoDB 4.4 …

frederic-huebner:
**Missing tests**
- [ ] SpringDataMongoBundleTracingIT


JoergSiebahn (approved):

JoergSiebahn:
> Missing Tests: SpringDataMongoBundleTracingIT

Would you expect a failure there? If not, I would say, you already converted more tests than I expected.

frederic-huebner:
> > Missing Tests: SpringDataMongoBundleTracingIT
> 
> Would you expect a failure there? If not, I would say, you already converted more tests than I expected.

Probably not, I just haven't thought of a way to get the `OTEL` instance in there yet.

PR #2724, opened 2023-09-20
chore: use Gradle's test-report-aggregation

christopher-cudennec:
Try to use Gradle's [report aggregation plugin](https://docs.gradle.org/current/userguide/test_report_aggregation_plugin.html).

I think it illustrates that we need to refactor our Gradle setup and write [custom Gradle plugins](https://docs.gradle.org/current/userguide/custom_plugins.html) to reuse build logic.


christopher-cudennec:
Fixed by #2728

PR #2729, opened 2023-09-21
chore: handle deprecations of `destination` and `buildDir`

christopher-cudennec:
See [deprecation note](https://docs.gradle.org/current/javadoc/org/gradle/api/reporting/ConfigurableReport.html#setDestination-java.io.File-):

```
@Deprecated
void setDestination​(java.io.File file)

Deprecated.
Use [Report.getOutputLocation()](https://docs.gradle.org/current/javadoc/org/gradle/api/reporting/Report.html#getOutputLocation--).set() instead. This method will be removed in Gradle 9.0.
```

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -311,7 +311,7 @@ task codeCoverageReport(type: JacocoReport) {
 
   reports {
     xml.required = true
-    xml.destination file("${buildDir}/reports/jacoco/report.xml")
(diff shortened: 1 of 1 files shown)


christopher-cudennec:
FYI: I could verify manually that files are still present in `build/reports/jacoco` 😃

eikenilsknopp (approved):

PR #2731, opened 2023-09-21
fix(openapi): embed parameter modifier for search operations

pathec:
Hi, the response model of search operations is currently restricted to exactly one list in the `EmbedParameterModifier` when using HAL links. But regarding the [REST guide](https://sda.dev/core-concepts/communication/restful-api-guide/#RESTfulAPIGuide-SHOULD%3AProvidecomplexqueriesusingsearchresources) normally a search operation contains a list with filters that is also a part of the response besides the result list. So the filter should also handle more than one list to get the correct `embed` query parameter documented in the OpenAPI

### DIFF
--- a/sda-commons-server-openapi/src/main/java/org/sdase/commons/optional/server/openapi/parameter/embed/EmbedParameterModifier.java
+++ b/sda-commons-server-openapi/src/main/java/org/sdase/commons/optional/server/openapi/parameter/embed/EmbedParameterModifier.java
@@ -118,10 +118,21 @@ public class EmbedParameterModifier implements ReaderListener {
             .filter(Objects::nonNull)
             .collect(Collectors.toList());
 
-    // only when there is a single list entry that supports embedding
-    if (nestedRefs.size() == 1) {
-      // get the model definition from the array
-      return getSchemaDefinition(definitions, nestedRefs.get(0));
+    // normally a search result model contains besides other meta information a list with the
+    // initially passed filters and a list with the search results therefore the correct list must
+    // be filtered
+    var schemaDefinition =
+        nestedRefs.stream()
+            // map refs to model definition
+            .map(schemaRef -> getSchemaDefinition(definitions, schemaRef))
+            // filter the model definition with embedded properties
+            .filter(
+                schema ->
+                    schema.getProperties() != null
+                        && schema.getProperties().containsKey(EMBEDDED_PROPERTY))
+            .findFirst();
+    if (schemaDefinition.isPresent()) {
+      return schemaDefinition.get();
     }
 
     return definition;
--- a/sda-commons-server-openapi/src/test/java/org/sdase/commons/server/openapi/apps/test/HouseSearchResource.java
+++ b/sda-commons-server-openapi/src/test/java/org/sdase/commons/server/openapi/apps/test/HouseSearchResource.java
@@ -10,6 +10,9 @@ import java.util.List;
 @Schema(name = "HouseSearchResource")
 public class HouseSearchResource {
 
+  @ArraySchema(arraySchema = @Schema(description = "The filters to apply", required = true))
+  private List<AnimalFilter> filters = new ArrayList<>();
+
   @ArraySchema(arraySchema = @Schema(description = "A list of found houses", required = true))
   private List<HouseResource> houses = new ArrayList<>();
 
@@ -18,11 +21,17 @@ public class HouseSearchResource {
 
   @JsonCreator
   public HouseSearchResource(
+      @JsonProperty("filters") List<AnimalFilter> filters,
       @JsonProperty("houses") List<HouseResource> houses,
       @JsonProperty("totalCount") int totalCount) {
 
+    this.filters.addAll(filters);
     this.houses.addAll(houses);
     this.totalCount = totalCount;
+  }
+
+  public List<AnimalFilter> getFilters() {
+    return filters;
   }
 
   public List<HouseResource> getHouses() {
--- a/sda-commons-server-openapi/src/test/java/org/sdase/commons/server/openapi/OpenApiBundleIT.java
+++ b/sda-commons-server-openapi/src/test/java/org/sdase/commons/server/openapi/OpenApiBundleIT.java
@@ -171,6 +171,7 @@ class OpenApiBundleIT {
             NATURAL_PERSON_DEFINITION,
             PARTNER_DEFINITION,
             "HALLink",
+            "AnimalFilter",
             "Animal",
             "House",
             "HouseSearchResource",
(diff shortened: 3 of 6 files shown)


christopher-cudennec:
Hi @pathec ! Nice to see you again!

I guess you talk about this section of the REST guide the includes an example request and response, right?

```
POST /carSearches
{
    "skip": 0,
    "limit": 2,
    "filters": [
        {
            "filterType": "seatsFilter",
            "seatsMin": 4,
            "seatsMax": 10
        }
    ],
    "orderBy": ["+sign"]
}

200 OK
{
    "skip": 0,
    "limit": 2,
    "totalResults": 25,
    "filters": [
        {
            "filterType": "seatsFilter",
            "seatsMin": 4,
            "seatsMax": 10
        }
    ],
    "results": [
        {
            "id": 711,
            "manufacturer": "bmw",
            "model": "X5",
            "seats": 5,
            "sign": "HH-MM-1234"
        },
        {
            "id": 712,
            "manufacturer": "bmw",
            "model": "X5",
            "seats": 5,
            "sign": "M-CV-777"
        }
    ]
}
```

Could you add an example for the current behaviour here in the PR?
If I understand you right, the problem occurs when you add the `embed` parameter to your search query?

pathec:
Hi @christopher-cudennec,

exactly the section in the guide I mean.

At runtime the code works fine, but the `embed` query parameter is not documented in OpenAPI currently.

The filter works fine already if the response model of an operation contains directly the HAL links, e.g. GET cars/1. But for search operations the car resource with the HAL links in defined in a list of results. Therefore the `EmbedParameterModifier` should search for HAL links in the properties of a response model and the important thing not restricted to a response model with exactly one list property.

I want for example something like this:

```yaml
  /cars:
    post:
      parameters:
      - description: "Select linked resources that should be resolved and embedded\
          \ into the response"
        in: "query"
        name: "embed"
        schema:
          type: "array"
          items:
            type: "string"
            enum:
            - "a"
            - "b"
            - "c"
     # ...
```

sda-commons-server-openapi/src/test/java/org/sdase/commons/server/openapi/apps/test/HouseSearchResource.java
```diff
@@ -18,13 +21,18 @@ public class HouseSearchResource {
 
   @JsonCreator
   public HouseSearchResource(
+      @JsonProperty("filters") List<Filter> filters,
```
christopher-cudennec, line 24: For completeness, you should also set `this.filters = filters` in the constructor.

sda-commons-server-openapi/src/main/java/org/sdase/commons/optional/server/openapi/parameter/embed/EmbedParameterModifier.java
```diff
@@ -118,10 +118,18 @@ private Schema<?> getResolvedSchemaDefinition(
...
-      return getSchemaDefinition(definitions, nestedRefs.get(0));
+    // normally a search result model contains besides other meta information a list with the
+    // initially passed filters and a list with the search results therefore the correct list must
+    // be filtered
+    var schemaDefinition =
+        nestedRefs.stream()
+            // map refs to model definition
+            .map(schemaRef -> getSchemaDefinition(definitions, schemaRef))
+            // filter the model definition with embedded properties
+            .filter(schema -> schema.getProperties().containsKey(EMBEDDED_PROPERTY))
```
christopher-cudennec, line 129: Please add a `null` check to `schema.getProperties()`

sda-commons-server-openapi/src/test/java/org/sdase/commons/server/openapi/apps/test/Filter.java
```diff
@@ -0,0 +1,15 @@
+package org.sdase.commons.server.openapi.apps.test;
+
+public class Filter {
```
christopher-cudennec, line 3: I wonder if we should extend the example to a more common use case because we usually support multiple filter types. In that case `Filter` is abstract and defines `@JsonSubTypes`. Your class would be a specific `AnimalFilter`. Any opinion, @JoergSiebahn ?

sda-commons-server-openapi/src/test/java/org/sdase/commons/server/openapi/apps/test/HouseSearchResource.java
pathec: Sure, I forgot this. Done

sda-commons-server-openapi/src/main/java/org/sdase/commons/optional/server/openapi/parameter/embed/EmbedParameterModifier.java
pathec: Good point. I added a check

sda-commons-server-openapi/src/test/java/org/sdase/commons/server/openapi/apps/test/Filter.java
JoergSiebahn: As I understand the implementation and the changes (it's actually very little understanding), that detail should not affect the case that is solved in the changed modifier. I think that introducing inheritance here is just more overhead.

christopher-cudennec:
We discussed internally that it would be nice to see the actual impact on an `openapi.yaml`. That's why I created #2734. Let's wait for this PR to be merged and rebase your changes after that. 😄

christopher-cudennec:
@pathec : Could you please rebase your PR? 😄

pathec:
@christopher-cudennec any blocker for this PR. I already rebased the branch twice, but before the build is ready the branch is already out-of-date :) How do you handle this?

christopher-cudennec:
> @christopher-cudennec any blocker for this PR. I already rebased the branch twice, but before the build is ready the branch is already out-of-date :) How do you handle this?

I'll let @elbiocaetano know to pause his maintenance task and stop merging PRs until your PR was merged. 😄

christopher-cudennec:
I wonder why the new test in [OpenApiBundleIT](https://github.com/SDA-SE/sda-dropwizard-commons/blob/1c57afb9c8344d210bbc40e854d13732c2b70980/sda-commons-server-openapi/src/test/java/org/sdase/commons/server/openapi/OpenApiBundleIT.java#L62) has not changed the expected output. I would've expected that it should fail once locally because you've changed `HouseSearchResource`. What happens when you execute it locally?

pathec:
> I wonder why the new test in [OpenApiBundleIT](https://github.com/SDA-SE/sda-dropwizard-commons/blob/1c57afb9c8344d210bbc40e854d13732c2b70980/sda-commons-server-openapi/src/test/java/org/sdase/commons/server/openapi/OpenApiBundleIT.java#L62) has not changed the expected output. I would've expected that it should fail once locally because you've changed `HouseSearchResource`. What happens when you execute it locally?

You're right, the test fails once locally. I added the change of expected spec now to the PR. Strange that the test has passed previously

pathec:
@christopher-cudennec  is there any manual step required for the fossa check? This check is still in progress for 3 hours

christopher-cudennec (approved):

christopher-cudennec:
I will remove the check for this PR. There seems to be an issue for PRs from forks. Not sure yet what the issue is.
But since you did not change / introduce any dependencies we can safely ignore the license check.

christopher-cudennec:
Thanks for your contribution! 😃

PR #2749, opened 2023-09-26
Fix checkout branch

frederic-huebner:


.github/workflows/cve-check.yml
```diff
@@ -29,6 +29,8 @@ jobs:
 
       - name: Checkout code
         uses: actions/checkout@3df4ab11eba7bda6032a0b82a6bb43b11571feac # v3
+        with:
+          ref: release/2.x.x
```
JoergSiebahn, line 33: It seems like scheduled workflows only run from [the default branch](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#schedule). Do we need a second workflow in master for the CVE checks of 2.x.x?

christopher-cudennec: > Scheduled workflows run on the latest commit on the default or base branch. The shortest interval you can run scheduled workflows is once every 5 minutes.

I guess it would be more transparent if we have a second workflow 👍

PR #2752, opened 2023-09-27
docs: adding documentation to backstage

elbiocaetano:


JoergSiebahn (approved):
So much content. Seems like we need a follow up to give some more guidance in the docs and shorten it a bit. @christopher-cudennec what do you think?

PR #2825, opened 2023-10-24
feat(metrics): add Kafka consumer metrics

sliesche:


sda-commons-server-kafka/src/test/resources/test-config-default.yml
```diff
@@ -18,6 +18,8 @@ kafka:
       config:
         key.deserializer: "org.apache.kafka.common.serialization.LongDeserializer"
         value.deserializer: "org.apache.kafka.common.serialization.LongDeserializer"
+        metric.reporters: "org.sdase.commons.server.kafka.helper.KafkaMetricsReporter"
+        metrics.sample.window.ms: 5000
```
JoergSiebahn, line 22: Is this a config that is needed in services as well?

sda-commons-server-kafka/src/main/java/org/sdase/commons/server/kafka/KafkaBundle.java
```diff
@@ -272,8 +282,17 @@ public <K, V> KafkaConsumer<K, V> createConsumer(
           org.apache.kafka.clients.consumer.ConsumerConfig.CLIENT_ID_CONFIG,
           consumerConfig.getClientId() + "-" + instanceId);
     }
+    KafkaConsumer<K, V> consumer =
+        new KafkaConsumer<>(consumerProperties, keyDeSerializer, valueDeSerializer);
+
+    if (micrometerConsumerListener != null) {
+      micrometerConsumerListener.consumerAdded(
+          consumerProperties.getProperty(
+              org.apache.kafka.clients.consumer.ConsumerConfig.CLIENT_ID_CONFIG),
```
JoergSiebahn, line 291: Why don't we use the `clientIdOptional` from above?

sda-commons-server-kafka/src/main/java/org/sdase/commons/server/kafka/KafkaBundle.java
```diff
@@ -272,8 +282,17 @@ public <K, V> KafkaConsumer<K, V> createConsumer(
           org.apache.kafka.clients.consumer.ConsumerConfig.CLIENT_ID_CONFIG,
           consumerConfig.getClientId() + "-" + instanceId);
     }
+    KafkaConsumer<K, V> consumer =
+        new KafkaConsumer<>(consumerProperties, keyDeSerializer, valueDeSerializer);
+
+    if (micrometerConsumerListener != null) {
```
JoergSiebahn, line 288: It seems like the `micrometerConsumerLister` is created unconditionally in `run`. Shouldn't we fail if it is `null`?
In the [client bundle we also fail](https://github.com/SDA-SE/sda-dropwizard-commons/blob/master/sda-commons-client-jersey/src/main/java/org/sdase/commons/client/jersey/JerseyClientBundle.java#L61) when the _create client_ calls are made too early.
Not sure if failing here would be breaking or if the code would fail anyways, when this is called before run.

However, in the case `micrometerConsumerListener` is `null` here, I assume, you won't get any metrics. Finding out why would be hard to find out. Would it make sense to log a warning in that case. At least it seems like the bundle API is used wrong by the consuming service, when `micrometerConsumerListener` is `null`.

sda-commons-server-kafka/src/main/java/org/sdase/commons/server/kafka/KafkaBundle.java
sliesche: The `clientIdOptional` is not available at this point. Actually the `clientIdOptional` is just a complicated way to read out the value that is set here. Sadly the KafkaConsumer has no method to access that value. There is most likely a more elegant way to handle that value but i have not put effort in looking into this without changing public method signatures

sda-commons-server-kafka/src/main/java/org/sdase/commons/server/kafka/KafkaBundle.java
sliesche: Good idea

add init check

```
if (!initialized) {
      throw new IllegalStateException("KafkaBundle not yet initialized!");
}
```

and logging

```
if (micrometerProducerListener == null) {
  LOGGER.error("MicrometerProducerListener is not initialized! Metrics will not be recorded.");
} else {
  micrometerProducerListener.producerAdded(clientId, producer);
}
```

sda-commons-server-kafka/src/main/java/org/sdase/commons/server/kafka/KafkaBundle.java
```diff
@@ -285,7 +300,9 @@ public <K, V> KafkaConsumer<K, V> createConsumer(
     KafkaConsumer<K, V> consumer =
         new KafkaConsumer<>(consumerProperties, keyDeSerializer, valueDeSerializer);
 
-    if (micrometerConsumerListener != null) {
+    if (micrometerConsumerListener == null) {
+      LOGGER.error("MicrometerConsumerListener is not initialized! Metrics will not be recorded.");
+    } else {
```
JoergSiebahn, line 305: I guess, with the initialization check, the `micrometerConsumerListener` can't be `null` here.

sda-commons-server-kafka/src/main/java/org/sdase/commons/server/kafka/KafkaBundle.java
```diff
@@ -204,6 +207,10 @@ public <K, V> List<MessageListener<K, V>> createMessageListener(
   public <K, V> MessageProducer<K, V> registerProducer(ProducerRegistration<K, V> registration)
       throws ConfigurationException {
 
+    if (!initialized) {
+      throw new IllegalStateException("KafkaBundle not yet initialized!");
```
JoergSiebahn, line 211: Maybe give some more hints?

```suggestion
      throw new IllegalStateException("KafkaBundle not yet initialized! registerProducer can be called earliest in run of the application.");
```

Maybe the sentence could be shorter and clearer than my suggestion.

- Same for [createConsumer](https://github.com/SDA-SE/sda-dropwizard-commons/pull/2825/files/4d885145a0325966decf2e5e4afe1ecec1e28712..32894c1a1bab7f60bb10fab9850ce1fdb6796a89#diff-786e204b7cd6a5ecc49de07ceb6a254aacf7290e9144a5b95d2c4ee6622a0207R290).
- What about the other create… methods? Some use the already available function [checkInit()](https://github.com/SDA-SE/sda-dropwizard-commons/pull/2825/files/4d885145a0325966decf2e5e4afe1ecec1e28712..32894c1a1bab7f60bb10fab9850ce1fdb6796a89#diff-786e204b7cd6a5ecc49de07ceb6a254aacf7290e9144a5b95d2c4ee6622a0207R470).

JoergSiebahn: As discussed internally, we should stay with a warning log for now, to avoid potentially breaking services. We should be stricter in the next major version.

sliesche: Did not see the `checkInit()`, i have removed my init additions and just log when listeners can not be used

sliesche: updated the checkInit() method with additional check for micometer listeners, that should not be breaking and just extend the functionality imho

```
  private void checkInit() {
    if (kafkaConfiguration == null
        || micrometerProducerListener == null
        || micrometerConsumerListener == null) {
      throw new IllegalStateException("KafkaConfiguration not yet initialized!");
    }
  }
```

sliesche: The `checkInit()` is used in `createMessageListener()` and `registerProducer()` methods which seem to be the main entry points for users of this bundle.

There are a couple of `createConsumer()`, `createProducer()` methods which are public were init checks could be added. I added the question to the [feature next major version](https://github.com/SDA-SE/sda-dropwizard-commons/discussions/2435#discussioncomment-7388886) discussion.

sda-commons-server-kafka/src/main/java/org/sdase/commons/server/kafka/KafkaBundle.java
```diff
@@ -212,9 +222,12 @@ public <K, V> MessageProducer<K, V> registerProducer(ProducerRegistration<K, V>
         new KafkaMessageProducer<>(
             registration.getTopic().getName(), producer, topicProducerCounterSpec, clientId);
 
-    if (micrometerProducerListener != null) {
+    if (micrometerProducerListener == null) {
+      LOGGER.error("MicrometerProducerListener is not initialized! Metrics will not be recorded.");
```
JoergSiebahn, line 226: NIT: imho this should be `warn`. The service is still functional without the metrics, isn't it?

JoergSiebahn (dismissed):

sda-commons-server-kafka/src/main/java/org/sdase/commons/server/kafka/KafkaBundle.java
JoergSiebahn:
```suggestion
      LOGGER.warn("MicrometerProducerListener is not initialized! Metrics will not be recorded.");
```

sliesche: updated for producer and consumer listener

JoergSiebahn (approved):

PR #2864, opened 2023-11-07
feat(metrics): deprecates SDA specific Kafka metrics, adds circuit br…

sliesche:
…eaker micrometer summary metric


sda-commons-server-circuitbreaker/build.gradle
```diff
@@ -7,6 +7,8 @@ dependencies {
     // version clash with latest client 0.9.0 -> 0.16.0
     exclude group: 'io.prometheus', module: 'simpleclient_common'
   }
+  implementation 'io.micrometer:micrometer-core'
```
christopher-cudennec, line 10: We haven't put much effort into the distinction between "api" and "implementation". Will users of dw-commons be able to track metrics? Or is that dependency necessary?

sda-commons-server-kafka/src/main/java/org/sdase/commons/server/kafka/prometheus/KafkaConsumerMetrics.java
```diff
@@ -15,16 +15,21 @@
...
  */
+@Deprecated(since = "5.8.1", forRemoval = true)
 public class KafkaConsumerMetrics extends Collector {
 
   private static final Logger LOGGER = LoggerFactory.getLogger(KafkaConsumerMetrics.class);
 
   /** The prefix for the metrics name as they are published to Prometheus */
   private static final String METRIC_NAME_PREFIX = "kafka_consumer_";
 
+  private static final String METRIC_DESCRIPTION_PREFIX = "**DEPRECATED** ";
```
christopher-cudennec, line 31: Is it worth to extract the value as a constant? I would prefer to name it "DEPRECATION_PREFIX" if you want to keep it.

christopher-cudennec (commented):
Do you add a "deprecated" label to the metrics at all? That's what we wrote down as acceptance criteria.

sda-commons-server-kafka/src/main/java/org/sdase/commons/server/kafka/prometheus/KafkaConsumerMetrics.java
sliesche: removed constant

sda-commons-server-circuitbreaker/build.gradle
sliesche: My understanding is that using the `implementation` configuration users of dw-commons will not be able to use micrometer-core code unless they add the library explicitly.

discussed this a couple of months ago with @JoergSiebahn and decided then that we don't want to expose micrometer to users of our library but just use it internally till further discussions happened.
Since we decided to move forward with micrometer we could now also expose the library to make it available to users but should then do it consistently 
https://github.com/SDA-SE/sda-dropwizard-commons/blob/c301f7806f9fbf72444271cba9f6b8169f751e1b/sda-commons-server-kafka/build.gradle#L9



@JoergSiebahn any objections using `api` here and make micrometer available to users of our library?

christopher-cudennec: I verified that we still see the implementation in the generated pom, but with `runtime` scope:
```
<dependency>
      <groupId>io.micrometer</groupId>
      <artifactId>micrometer-core</artifactId>
      <scope>runtime</scope>
      …
```
So, using `implementation` here would be ok for me. 😄

JoergSiebahn: Imho we only should use API where consumers need to interact with the 3rd party code. As this was not handled very consistently in the past, I would not change everything but start going in the right direction when possible. As we already use implementation here, I would keep it like that. Afaik you won't use micrometer for custom service specific metrics but stick with the Dropwizard metrics in that case?

sliesche:
> Do you add a "deprecated" label to the metrics at all? That's what we wrote down as acceptance criteria.

No i did not, a label/tag has a name and a value, i decided to add `deprecated` to the corresponding classes and in the metric help message. From my point of view this makes more sense than adding a label/tag deprecated=true.

With the exception of `resilience4j_circuitbreaker_calls_bucket` the circuit breaker changes stay the same we just use a micrometer library instead of pushing them to prometheus directly.

sda-commons-server-circuitbreaker/build.gradle
sliesche: fair enough then we leave it as is

christopher-cudennec:
PR looks good to me; any comments from your side, @JoergSiebahn ?

sda-commons-server-circuitbreaker/src/main/java/org/sdase/commons/server/circuitbreaker/metrics/SdaCircuitBreakerMetricsCollector.java
```diff
@@ -0,0 +1,55 @@
...
+ *
+ * <p>Prometheus and Micrometer handle histogram metrics differently. Prometheus adds a
+ * resilience4j_circuitbreaker_calls_bucket histogram while micrometer adds a
+ * resilience4j_circuitbreaker_calls_count summary.
+ *
+ * <p>This class is adding only the prometheus histogram metric to not introduce a breaking changes.
+ * The rest of the metrics are added by the resilience4j micrometer library {@link
+ * io.github.resilience4j.micrometer.tagged.TaggedCircuitBreakerMetrics}.
+ *
+ * @deprecated class should be removed in the next major release. This will introduce breaking
```
JoergSiebahn, line 21: Is there a micrometer alternative or will we loose inside for circuit breakers?

docs/server-prometheus.md
```diff
@@ -9,27 +9,27 @@ The endpoint is available at the applications admin port at `/metrics/prometheus
...
+|                                             | _`resource_path`_       | The mapped path of the request with path param placeholders.                                                | Request Context                           |               |
+|                                             | _`status_code`_         | The HTTP status code sent with the response.                                                                | Response Context                          |               |
+|                                             | _`consumer_name`_       | Name of the consumer that started the request.                                                              | Request Context Property `Consumer-Name`* |               |
+| **`kafka_consumer_records_lag`**            |                         | **DEPRECATED** See [Kafka Documentation](https://kafka.apache.org/documentation/#consumer_fetch_monitoring) | Bridged from Kafka                        |               |
```
JoergSiebahn, line 20: Could we mention the Micrometer alternative for the deprecated metrics?

sda-commons-server-circuitbreaker/src/main/java/org/sdase/commons/server/circuitbreaker/metrics/SdaCircuitBreakerMetricsCollector.java
sliesche: All circuit breaker metric stay the same except for `resilience4j_circuitbreaker_calls_bucket` histogram which will be replaced by  `resilience4j_circuitbreaker_calls_seconds_count` summary.

With this PR both will be available till `SdaCircuitBreakerMetricsCollector` is removed

docs/server-prometheus.md
sliesche: Added the replacements for `kafka_consumer_records_lag` and `kafka_producer_topic_message_total`

However there is no equivalent for `kafka_consumer_topic_message_duration` histogram since this is not an internal Kafka metric. What can be used is the `kafka_consumer_fetch_manager_records_consumed_rate` gauge which is inverse to above message duration metric.

JoergSiebahn (approved):

PR #2901, opened 2023-11-21
feat: upgrading to dropwizard 4.0.4 and fixing imports

elbiocaetano:


sda-commons-client-jersey-example/src/main/java/org/sdase/commons/client/jersey/JerseyClientExampleApplication.java
```diff
@@ -1,10 +1,12 @@
...
-import io.dropwizard.Application;
-import io.dropwizard.Configuration;
-import io.dropwizard.setup.Bootstrap;
-import io.dropwizard.setup.Environment;
+import io.dropwizard.core.Application;
+import io.dropwizard.core.Configuration;
+import io.dropwizard.core.setup.Bootstrap;
+import io.dropwizard.core.setup.Environment;
+import jakarta.ws.rs.client.Client;
+import jakarta.ws.rs.core.MediaType;
```
christopher-cudennec, line 9: What about lines 12 + 13 now?

elbiocaetano: It was some missed lines that I forgot to remove. But it will be fixed in the next PR

JoergSiebahn: Should we take over [this test](https://github.com/SDA-SE/sda-spring-boot-commons/blob/master/sda-commons-dependency-check/src/test/java/org/sdase/commons/spring/boot/dependency/check/AvoidJavaxTest.java) from Spring Boot Commons, when we are a few steps further?

elbiocaetano: Good idea, I will add in the next PR that includes Jakarta

christopher-cudennec (approved):

PR #2902, opened 2023-11-21
feat: moving from javax.* to jakarta.*

elbiocaetano:


sda-commons-server-weld/build.gradle
```diff
@@ -27,10 +27,10 @@ dependencies {
     exclude group: 'org.jboss.spec.javax.interceptor', module: 'jboss-interceptors-api_1.2_spec'
     /**
      * Dropwizard manages jakarta.annotation:jakarta.annotation-api while Weld brings
-     * org.jboss.spec.javax.annotation:jboss-annotations-api_1.3_spec. Both have identical contents.
+     * org.jboss.spec.jakarta.annotation:jboss-annotations-api_1.3_spec. Both have identical contents.
      * We keep the version managed by Dropwizard.
      */
-    exclude group: 'org.jboss.spec.javax.annotation', module: 'jboss-annotations-api_1.3_spec'
+    exclude group: 'org.jboss.spec.javakarta.annotation', module: 'jboss-annotations-api_1.3_spec'
```
christopher-cudennec, line 33: search+replace error? `jakarta` instead of `javakarta`?

sda-commons-dependency-check/src/test/java/org/sdase/commons/dependency/check/AvoidJavaxTest.java
```diff
@@ -0,0 +1,58 @@
...
+
+class AvoidJavaxTest {
+
+  private static final Logger LOG = LoggerFactory.getLogger(AvoidJavaxTest.class);
+
+  final Set<String> ALLOWED_FILE_PARTS =
+      Stream.of(
+              // Wiremock repacked classes. A Jakarta variant is not available yet. This will not
+              // influence production code.
+              "/com.github.tomakehurst/wiremock-jre8-standalone/",
```
christopher-cudennec, line 25: Is this still the correct path? I noticed that you switched to `org.wiremock:wiremock:3.3.1`.

JoergSiebahn: If we don't use the standalone variant any more, this exclude should not be needed.

elbiocaetano: It was removed, indeed

christopher-cudennec (approved):

PR #2927, opened 2023-11-29
Drop support for Java 11

christopher-cudennec:


sda-commons-client-jersey-example/src/main/java/org/sdase/commons/client/jersey/clients/apia/Car.java
```diff
@@ -1,24 +1,3 @@
...
-  public String getColor() {
-    return color;
-  }
-
-  public Car setColor(String color) {
-    this.color = color;
-    return this;
-  }
-}
+public record Car(String sign, String color) {}
```
JoergSiebahn, line 3: I guess, there is more version related code smell like this? Any idea how we want to migrate the code?

JoergSiebahn (approved):

sda-commons-client-jersey-example/src/main/java/org/sdase/commons/client/jersey/clients/apia/Car.java
christopher-cudennec: It's possible to use InteeliJ's feature "Analyze > Inspect code" and look for "Java > Java language level migration aids". Let me check what I can find there.

PR #2933, opened 2023-11-30
ci: upgrade to java 21 on gh actions

elbiocaetano:


christopher-cudennec (commented):
I would only add 21 to the matrix in `java-ci.yaml` and keep the other workflows as they are. Compare our acceptance criteria.

JoergSiebahn:
> I would only add 21 to the matrix in java-ci.yaml and keep the other workflows as they are. Compare our acceptance criteria.

Yes. If we test with Java 17 to ensure compatibility, we can not publish with Java 21. It would break the tested compatibility.

PR #2935, opened 2023-11-30
test: upgrade to wiremock 3

elbiocaetano:


sda-commons-dependency-check/src/test/java/org/sdase/commons/dependency/check/UnwantedDependenciesTest.java
```diff
@@ -126,6 +126,21 @@ void checkForApacheHttpClientV4() {
     }
   }
 
+  /**
+   * This test finds the library `com.github.tomakehurst` in our classpath. We don't want to use it,
```
christopher-cudennec, line 130: nit: the library is missing `wiremock` at the end. You can see it below in your implementation.

PR #2991, opened 2023-12-13
Update AWS SDK to v2

christopher-cudennec:


sda-commons-server-s3/src/main/java/org/sdase/commons/server/s3/S3Bundle.java
```diff
@@ -111,11 +102,90 @@ public void run(C configuration, Environment environment) throws Exception {
...
+        .build()
+        .newExecutionInterceptor();
+  }
+
+  private Signer createSigner(S3Configuration s3Configuration) {
+    String signerOverride = s3Configuration.getSignerOverride();
+    if (signerOverride == null || signerOverride.isBlank()) {
+      return AwsS3V4Signer.create();
+    }
+    // I don't know how to do that more elegantly
```
sliesche, line 163: Did you have a look at
`SignerFactory#getSignerByTypeAndService(String signerType, final String serviceName)`

sda-commons-server-s3/src/main/java/org/sdase/commons/server/s3/S3Bundle.java
```diff
@@ -111,11 +102,90 @@ public void run(C configuration, Environment environment) throws Exception {
...
+   * @return the region to use for the S3 client, default is 'eu-central-1'
+   */
+  private Region getRegion() {
+    String region = s3Configuration.getRegion();
+    if (region != null && !region.isBlank()) {
+      return Region.of(region);
+    }
+
+    // default
+    return Region.EU_CENTRAL_1;
```
sliesche, line 145: Does it make sense to put the default value in `S3Configuration` similar to the default value of `signerOverride`

sda-commons-server-s3/src/main/java/org/sdase/commons/server/s3/S3Bundle.java
christopher-cudennec: SignerFactory is part of the v1 SDK. I could not find something similar for v2.
(the testing module still ships v1)

sda-commons-server-s3-testing/src/main/java/org/sdase/commons/server/s3/testing/S3ClassExtension.java
```diff
@@ -1,63 +1,125 @@
...
 
 /**
- * JUnit 5 extension for running a AWS S3-compatible object storage alongside the (integration)
+ * JUnit 5 extension for running an AWS S3-compatible object storage alongside the (integration)
  * tests. Use {@link #getEndpoint()} to retrieve the endpoint URL to connect to.
  *
  * <p>Example usage:
  *
  * <pre>
+ * &#64;LocalS3
```
christopher-cudennec, line 31: Question to the reviewers: Do you like this approach?
* good: less code because we extend the Junit 5 extension from local-s3
* bad: `LocalS3` annotation is forced on all our consumers

We could completely re-write the extension by copying the code from local-s3 and adapting it to our needs. Consumers would not need to add the annotation in that case.

sda-commons-server-s3/src/main/java/org/sdase/commons/server/s3/S3Bundle.java
christopher-cudennec: Why not. That makes it more transparent. The current documentation of the property in `S3Configuration` says "can be left empty" which might even be misleading. I've changed the default from `us-west-2` (v1) to `eu-central-1`. Usually customers will set their region anyway, e.g. when they host their own S3 server.

sda-commons-server-s3-testing/src/main/java/org/sdase/commons/server/s3/testing/S3ClassExtension.java
frederic-huebner: I would be fine with the annotation.

sda-commons-server-s3/src/main/java/org/sdase/commons/server/s3/S3Bundle.java
```diff
@@ -111,11 +102,90 @@ public void run(C configuration, Environment environment) throws Exception {
...
+  }
+
+  /**
+   * Using basic credentials. If no credentials are provided, anonymous credentials are used.
+   *
+   * @return credentials provider
+   */
+  private AwsCredentialsProvider createCredentialsProvider(String acccessKey, String secretKey) {
+    return isBlank(acccessKey) && isBlank(secretKey)
+        ? AnonymousCredentialsProvider.create()
```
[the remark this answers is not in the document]
christopher-cudennec: I've implemented that as a new feature for the S3Bundle 👍

sda-commons-server-s3-testing/src/main/java/org/sdase/commons/server/s3/testing/S3ClassExtension.java
```diff
@@ -1,63 +1,125 @@
...
 
-  @Override
-  public void stop() {
-    super.stop();
+  /**
+   * @return a new S3 client
+   * @deprecated Use {@link #newClient()} instead.
+   */
+  @Deprecated
+  public S3Client getClient() {
```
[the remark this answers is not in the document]
christopher-cudennec: Maybe it's just a matter of taste. It helps a little with the migration because it's one step less that users of the s3 testing module need to adapt right away. I've flagged it with `forRemoval` anyway. (I wonder who annotates code with `Deprecated` but does not want to remove it 🤔 )

christopher-cudennec:
> I associate annotation based approaches more with spring boot but thats no reason not to do it.

It's also quite common in simple JUnit 5 tests, e.g. https://junit-pioneer.org/docs/system-properties/

frederic-huebner (approved):

PR #3025, opened 2023-12-20
chore: use windows-2019 in "Java CI" workflow

christopher-cudennec:
according to the following source it should improve the performance: https://github.com/actions/runner-images/issues/7320


christopher-cudennec:
not a big improvement 😞 
closing the PR.

PR #3026, opened 2023-12-20
fix(server-mongo-testing): remove deprecated methods from MongoDb int…

christopher-cudennec:
…erface

BREAKING CHANGE: remove deprecated methods from MongoDb interface

Removed getters from the `MongoDb` interface which affects the `MongoDbClassExtension`. You can retrieve information about the database, username or password from the `ConnectionString` that is provided by `MongoDbClassExtension#getMongoConnectionString()`.

### DIFF
--- a/sda-commons-server-mongo-testing/src/main/java/org/sdase/commons/server/mongo/testing/MongoDb.java
+++ b/sda-commons-server-mongo-testing/src/main/java/org/sdase/commons/server/mongo/testing/MongoDb.java
@@ -3,6 +3,7 @@ package org.sdase.commons.server.mongo.testing;
 import static de.flapdoodle.embed.mongo.distribution.Version.Main.V4_4;
 import static java.util.concurrent.TimeUnit.MINUTES;
 
+import com.mongodb.ConnectionString;
 import com.mongodb.MongoClient;
 import com.mongodb.client.MongoDatabase;
 import de.flapdoodle.embed.mongo.distribution.IFeatureAwareVersion;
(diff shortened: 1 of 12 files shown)


sda-commons-server-mongo-testing/src/main/java/org/sdase/commons/server/mongo/testing/UseExistingMongoDb.java
```diff
@@ -75,14 +46,15 @@ public String getServerVersion() {
...
           "Testing as '{}' {} in MongoDB '{}' at '{}' with options '{}'",
-          username,
+          mongoConnectionString.getUsername(),
           withPasswordText,
-          getDatabase(),
-          getHosts(),
-          getOptions());
+          mongoConnectionString.getDatabase(),
+          mongoConnectionString.getHosts(),
+          optionsFromMongoUrl(connectionString));
```
elbiocaetano, line 57: Could we keep onlu mongoConnectionString variable and instead of the `connectionString` variable we use the `mongoConnectionString.getConnectionString()` method?

christopher-cudennec: Yes, that was possible. 👍 
I've updated `UseExistingMongoDb` and `StartLocalMongoDb`.

elbiocaetano (approved):

sda-commons-server-mongo-testing/src/main/java/org/sdase/commons/server/mongo/testing/MongoDb.java
```diff
@@ -43,59 +44,32 @@ public interface MongoDb {
...
-   * @return the password that must be used to connect to the database.
-   * @deprecated Please use 'getConnectionString()' instead.
-   */
-  @Deprecated(forRemoval = true)
-  String getPassword();
-
-  /**
-   * @return the initialized database
-   */
-  String getDatabase();
```
christopher-cudennec, line 70: I wonder why this method was not deprecated 🤔

PR #3056, opened 2024-01-02
CDI: Undo deprecation of `WeldAppExtension` and `WeldTestSupport`

JoergSiebahn:
To reduce maintenance effort, we deprecated the [WeldAppExtension](https://github.com/SDA-SE/sda-dropwizard-commons/releases/tag/5.16.0) for removal. 
So far we did not show an alternative way for testing applications using Weld. This PR is meant to close this gap.

To have better coverage, the test of the example module is extended to verify a REST response ~~and then [migrated](https://github.com/SDA-SE/sda-dropwizard-commons/commit/77a38f37dcda0a3f5e28af650299f40259895440)~~. We decided to undo the deprecation and have test support for Weld in the future. The official extension requires too much overhead for a working integration test of a full Dropwizard application.

**Disclaimer**

We suggest to use pure Java with this library instead of an additional layer of magic like CDI. You may want to consider [Spring Boot for SDA compliant services](https://github.com/SDA-SE/sda-spring-boot-commons) that use dependency injection.

### DIFF
--- a/sda-commons-server-weld-example/build.gradle
+++ b/sda-commons-server-weld-example/build.gradle
@@ -14,6 +14,7 @@ dependencies {
   api project(':sda-commons-server-dropwizard')
   api project(':sda-commons-server-weld')
 
-  testImplementation project(':sda-commons-server-weld-testing')
+  testImplementation project(':sda-commons-server-testing')
+  testImplementation 'org.jboss.weld:weld-junit5:2.0.2.Final'
 }
 
--- a/sda-commons-server-weld-example/build.gradle
+++ b/sda-commons-server-weld-example/build.gradle
@@ -14,7 +14,6 @@ dependencies {
   api project(':sda-commons-server-dropwizard')
   api project(':sda-commons-server-weld')
 
-  testImplementation project(':sda-commons-server-testing')
-  testImplementation 'org.jboss.weld:weld-junit5:2.0.2.Final'
+  testImplementation project(':sda-commons-server-weld-testing')
 }
 
--- a/sda-commons-server-weld-example/src/test/java/org/sdase/commons/server/weld/WeldExampleApplicationITest.java
+++ b/sda-commons-server-weld-example/src/test/java/org/sdase/commons/server/weld/WeldExampleApplicationITest.java
@@ -1,11 +1,13 @@
 package org.sdase.commons.server.weld;
 
 import static io.dropwizard.testing.ConfigOverride.randomPorts;
+import static org.assertj.core.api.Assertions.assertThat;
 
 import io.dropwizard.Configuration;
-import org.assertj.core.api.Assertions;
+import javax.ws.rs.core.MediaType;
 import org.junit.jupiter.api.Test;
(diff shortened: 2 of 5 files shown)


sda-commons-server-weld-example/src/test/java/org/sdase/commons/server/weld/WeldExampleApplicationITest.java
```diff
@@ -1,22 +1,52 @@
...
 import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.TestInstance;
 import org.junit.jupiter.api.extension.RegisterExtension;
-import org.sdase.commons.server.weld.testing.WeldAppExtension;
+import org.sdase.commons.server.weld.beans.UsageBean;
 
+@EnableAutoWeld
+@AddPackages(WeldExampleApplication.class)
+@AddExtensions(CdiComponentProvider.class)
+@TestInstance(TestInstance.Lifecycle.PER_CLASS)
```
christopher-cudennec, line 24: `PER_CLASS` seems to be necessary because the [Weld JUnit 5 extension](https://github.com/weld/weld-testing/blob/master/junit5/src/main/java/org/jboss/weld/junit5/WeldJunit5Extension.java#L130) will not start the Weld container otherwise. Default Junit 5 lifecycle is `PER_METHOD`.

christopher-cudennec (dismissed):

sda-commons-server-weld-example/build.gradle
```diff
@@ -14,6 +14,7 @@ dependencies {
   api project(':sda-commons-server-dropwizard')
   api project(':sda-commons-server-weld')
 
-  testImplementation project(':sda-commons-server-weld-testing')
+  testImplementation project(':sda-commons-server-testing')
+  testImplementation 'org.jboss.weld:weld-junit5:2.0.2.Final'
```
christopher-cudennec, line 18: We've introduced "weld-junit5" to [server-weld-testing in version 6](https://github.com/SDA-SE/sda-dropwizard-commons/blob/6.0.0-rc.1/sda-commons-server-weld-testing/build.gradle). I'd to the same in this PR to prepare the change.

Our customer has specifically asked for maintaining a version of this library as convenience for the users of sda-dw-commons.

sda-commons-server-weld-example/src/test/java/org/sdase/commons/server/weld/WeldExampleApplicationITest.java
JoergSiebahn: „not start“ is not exactly true. But the Weld container is only available for each test method, which is not suitable for the Dropwizard extension.

sda-commons-server-weld-example/build.gradle
JoergSiebahn: We will add a managed version of weld-junit5 in the next major release of sda-dropwizard-commons.

christopher-cudennec (approved):

PR #3123, opened 2024-01-30
chore: add all submodules as dependency of dependency-check

christopher-cudennec:


sda-commons-dependency-check/build.gradle
```diff
@@ -5,41 +5,50 @@ dependencies {
 
   // Add all new modules here to verify that there are no dependency clashes.
   api project(':sda-commons-client-jersey')
-  api project(':sda-commons-shared-wiremock-testing')
+  api project(':sda-commons-client-jersey-example')
```
christopher-cudennec, line 8: @JoergSiebahn : Do you know if there was a reason to not include the example modules?

christopher-cudennec:
Closing as:
* example modules are not relevant for external projects consuming sda-dropwizard-commons
* the Gradle build fails and I don't want to invest time in fixing it.

sda-commons-dependency-check/build.gradle
JoergSiebahn: Because they are not provided for the consumers. But maybe they are a good indicator if dependency resolution will fail when the library is used in a service?

PR #3163, opened 2024-02-12
feat(server-hibernate): add flyway-postgresql dependency

eikenilsknopp:


sda-commons-server-hibernate/build.gradle
```diff
@@ -16,6 +16,7 @@ dependencies {
   api 'org.flywaydb:flyway-core'
   api 'jakarta.annotation:jakarta.annotation-api'
   api 'org.flywaydb:flyway-core'
+  api 'org.flywaydb:flyway-database-postgresql'
```
JoergSiebahn, line 19: Would it be enough to just add it to the dependency management and check if it works without version in an example module?

sda-commons-dependencies/build.gradle
```diff
@@ -119,7 +120,8 @@ dependencies {
 
     // sda-commons-server-hibernate
     api "org.postgresql:postgresql:42.7.1"
-    api "org.flywaydb:flyway-core:10.7.2"
+    api "org.flywaydb:flyway-core:${flywayVersion}"
+    api "org.flywaydb:flyway-database-postgresql:${flywayVersion}"
```
JoergSiebahn, line 124: Afaik SI is also using DB2 in base services. Should we manage that as well?

sda-commons-server-hibernate/build.gradle
eikenilsknopp: As Steve pointed out, the module is documented to [work specifically with PostgreSQL](https://github.com/SDA-SE/sda-dropwizard-commons/blob/ba635ebac3dd4d5885234a533d5c3ae07ff70095/docs/server-hibernate.md?plain=1#L6), so in my opinion it should provide everything needed to work out of the box.

sda-commons-dependencies/build.gradle
eikenilsknopp: We can, then I would add that to the documentation as well. At the moment it's just documented to work with postgre.

sda-commons-server-hibernate/build.gradle
JoergSiebahn: Hmmm, ok. Then let's keep it as it is here.

sda-commons-dependencies/build.gradle
JoergSiebahn: Let's just wait until we get a specific request for DB2. I hope, that you can pull in the Flyway DB2 (or other) module in addition to the Postgres module.

But regarding the docs: Maybe [this](https://github.com/SDA-SE/sda-dropwizard-commons/blob/ba635ebac3dd4d5885234a533d5c3ae07ff70095/docs/server-hibernate.md?plain=1#L13) is too specific. Could you add the flyway postgres module and remove the versions?

JoergSiebahn (approved):

PR #3184, opened 2024-02-19
chore: use opentelemetry httpclient 5.2 instrumentation

eikenilsknopp:


sda-commons-server-auth/src/test/java/org/sdase/commons/server/auth/key/OpenIdProviderDiscoveryKeySourceIT.java
```diff
@@ -142,7 +143,11 @@ void shouldLoadKeyWithTracing() {
               assertThat(spans)
                   .hasSize(2)
                   .extracting(
-                      s -> s.getAttributes().asMap().get(stringKey("http.url")), SpanData::getName)
+                      s ->
+                          s.getAttributes()
+                              .asMap()
+                              .get(stringKey(SemanticAttributes.URL_FULL.getKey())),
```
JoergSiebahn, line 149: What's the difference between `"http.url"` and `SemanticAttributes.URL_FULL.getKey()`?
Currently you are testing the same property that is used in production code. I would prefer to use a pure `String` (or a value defined in test scope) in tests. Using the same as in production will never fail - even if you (or the dependency) change something that breaks the contract - in this case maybe a dashboard.

sda-commons-dependencies/build.gradle
```diff
@@ -197,12 +197,15 @@ dependencies {
...
+    api "io.opentelemetry.instrumentation:opentelemetry-apache-httpclient-5.2:2.1.0-alpha"
     api "io.opentelemetry:opentelemetry-exporter-otlp:$openTelemetryVersion"
     api "io.opentelemetry:opentelemetry-exporter-jaeger:$openTelemetryVersion"
     api "io.opentelemetry:opentelemetry-extension-trace-propagators:$openTelemetryVersion"
     api "io.opentelemetry:opentelemetry-sdk-testing:$openTelemetryVersion"
     api "io.opentelemetry:opentelemetry-exporter-logging:$openTelemetryVersion"
-    //api "io.opentelemetry:opentelemetry-semconv:${openTelemetryVersion}"
+    api "io.opentelemetry:opentelemetry-semconv:1.30.1-alpha"
+    api "io.opentelemetry.semconv:opentelemetry-semconv:1.23.1-alpha"
+    api "io.opentelemetry.instrumentation:opentelemetry-instrumentation-api:2.1.0"
```
JoergSiebahn, line 208: We should create a follow up to cleanup the OTEL dependencies. We enforce the OTEL-bom (why do we need two, do they conflict?) but still have to manage a lot on our own. I hope that can be simplified.

JoergSiebahn (approved):

PR #3243, opened 2024-03-11
Handle Trace-Token in Kafka Messaging

JoergSiebahn:


sda-commons-client-jersey/src/main/java/org/sdase/commons/client/jersey/filter/TraceTokenClientFilter.java
```diff
@@ -18,42 +13,22 @@
...
-  }
-
-  private void addNewTokenToMdcForCurrentRequest(
-      ClientRequestContext requestContext, String newTraceToken) {
-    if (MDC.getMDCAdapter() != null) {
-      MDC.put(RequestTracing.TOKEN_MDC_KEY, newTraceToken);
-      requestContext.setProperty(ADDED_TO_MDC_FOR_REQUEST, Boolean.TRUE);
+    if (requestContext.getProperty(TRACE_TOKEN_CONTEXT)
+        instanceof TraceTokenContext traceTokenContext) {
+      traceTokenContext.closeIfCreated();
```
sliesche, line 31: What if the client does not respond, is then the trace context is not cleared?
If that all in a separate thread i guess that would not be a problem

JoergSiebahn: > What if the client does not respond, is then the trace context is not cleared?

You are right. I did not expect that this kind of filter is only called when the server really responds. But that actually happened in the previous implementation as well.

> If that all in a separate thread i guess that would not be a problem

There could be a problem with a long lasting trace token in special cases. But I would assume that in most cases a thread will be terminated after a single action when it is not triggered with a trace token context that is around this request.

However, a quick research shows, that you can't cover the error case of timeout or connection issues. We could only change the behavior and avoid creating a trace token if we are not already in a context with a request token.

What do you think is the better option?

JoergSiebahn: Good idea. I added [this comment](https://github.com/SDA-SE/sda-dropwizard-commons/pull/3243/commits/2e224c3345b975dffdf31b6fc8ec81d9cca1e894).

sliesche (approved):

PR #3247, opened 2024-03-12
chore(deps): update opentelemetry bom version

eikenilsknopp:


JoergSiebahn (dismissed):
Very confusing version management in OTEL. Do you have any information, why `opentelemetry-exporter-jaeger` does not exist with the new version? If so, that information might be helpful in the PR.

eikenilsknopp:
> Very confusing version management in OTEL. Do you have any information, why `opentelemetry-exporter-jaeger` does not exist with the new version? If so, that information might be helpful in the PR.

Hmm they stopped releasing them in 1.35.0 because jaeger supports native otel: https://github.com/open-telemetry/opentelemetry-java/releases/tag/v1.35.0

Let me try if it works without the dependencies

JoergSiebahn (approved):

PR #3312, opened 2024-04-02
fix(kafka): treat continued handling errors like successful consumed message

JoergSiebahn:
Note: This fix only affects cases, when a subsequent message in the same poll causes the consumer to stop consuming. It may also cover scenarios when the application is shut down while the messages of this poll are handled.


sda-commons-server-kafka/src/main/java/org/sdase/commons/server/kafka/consumer/strategies/autocommit/AutocommitMLS.java
```diff
@@ -60,7 +60,9 @@ public void processRecords(ConsumerRecords<K, V> records, KafkaConsumer<K, V> co
               handler.getClass(),
               e);
           boolean shouldContinue = errorHandler.handleError(consumerRecord, e, consumer);
-          if (!shouldContinue) {
+          if (shouldContinue) {
+            addOffsetToCommitOnClose(consumerRecord);
+          } else {
```
AndreasHuebnerSda, line 65: else block isn't necessary

sliesche (approved):

PR #3499, opened 2024-06-20
authentication header is not used in request after redirect

JoergSiebahn:
This PR reproduces a reported issue that the Authentication header is lost when following 303 See Other redirects.

According the report, the behavior has changed from v5 to v6. This can be confirmed with a copy of [the new test](https://github.com/SDA-SE/sda-dropwizard-commons/pull/3499/commits/5ad6748df07eb1cc6be667b4ecbbc84e11383777) to the current state of `release/5.x.x`.


JoergSiebahn:
The change may be intended and required to comply with the spec. It may be caused by [this issue](https://github.com/eclipse-ee4j/jersey/issues/5414). I'm not 100% sure, if it is the Netty connector that is internally doing the requests of the clients created here. But that change was released with 2.41 and the latest Jersey version in DW Commons 5.x.x was 2.39.

PR #3700, opened 2024-10-08
fix: add uuid to IssuerClient oidc-client name for unique metric

sabkuehn:
We need to call "new OidcRequestFilter(jerseyClientBundle.getClientFactory(), oidcConfiguration, false);" multiple times in our service as there can be multiple Oidc. 

The code currently returns the following error: 
A metric named org.apache.hc.client5.http.io.HttpClientConnectionManager.oidc-client.available-connections already exists
java.lang.IllegalArgumentException: A metric named org.apache.hc.client5.http.io.HttpClientConnectionManager.oidc-client.available-connections already exists

We need a unique name for the oidc-client so that the error no longer occurs.

### DIFF
--- a/sda-commons-client-jersey/src/main/java/org/sdase/commons/client/jersey/oidc/OidcClient.java
+++ b/sda-commons-client-jersey/src/main/java/org/sdase/commons/client/jersey/oidc/OidcClient.java
@@ -40,6 +40,26 @@ public class OidcClient {
     this.cache = Caffeine.newBuilder().expireAfter(new AfterCreateExpiry()).build();
   }
 
+  public OidcClient(ClientFactory clientFactory, OidcConfiguration config, String clientName) {
+    this.config = config;
+
+    if (config.isDisabled()) {
+      LOGGER.warn("OIDC was disabled.");
+      this.cache = null;
+      this.issuerClient = null;
+      return;
+    }
+
+    this.issuerClient = new IssuerClient(clientFactory, config, clientName);
+
+    if (config.getCache().isDisabled()) {
+      this.cache = null;
+      return;
+    }
+
+    this.cache = Caffeine.newBuilder().expireAfter(new AfterCreateExpiry()).build();
+  }
+
   /**
    * Retrieves a new access token from the token endpoint using the given {@link OidcConfiguration}.
    *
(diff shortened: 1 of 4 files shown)


sabkuehn:
@JoergSiebahn 
I think passing a fully custom name in the constructor would be the most dynamic solution. I changed the implementation by copying the constructors of each class. Currently there is duplicate code because all parameters are final. This could be adjusted if you prefer.

JoergSiebahn:
Hi @sabkuehn,

thanks for updating. I think, we could live the code duplication, but it causes an issue with Sonar for code coverage.

Imho a simple new test case in [`OidcRequestFilterTest`](https://github.com/SDA-SE/sda-dropwizard-commons/blob/main/sda-commons-client-jersey/src/test/java/org/sdase/commons/client/jersey/oidc/OidcRequestFilterTest.java#L34) can fix that:

```java
  @Test
  void shouldUseCustomNameForMetrics() {
    // given
    clearAuthenticationContext();
    setupTokenEndpoint(true);

    // when new custom filter is used
    var filter =
        new OidcRequestFilter(
            app.getJerseyClientBundle().getClientFactory(),
            DW.getConfiguration().getOidc(),
            true,
            "custom-oidc-client");
    filter.filter(Mockito.mock(ClientRequestContext.class, Mockito.RETURNS_DEEP_STUBS));

    // then metrics show name
    var metrics = DW.getEnvironment().metrics().getMetrics().keySet();
    assertThat(metrics).anyMatch(m -> m.contains(".custom-oidc-client."));
  }
```

Also: Can you run the code formatting with `spotlessApply` and push again (see [spotless check](https://github.com/SDA-SE/sda-dropwizard-commons/actions/runs/11249823612/job/31279570918?pr=3700))?

And when you force push, can you adapt the commit message to the current implementation with a prefix mentioning the package. Personally I prefer to mention the purpose more than the implementation details and would see this as a feature, e.g. `feat(client-jersey): support multiple OIDC providers for client authentication`. But I can also squash merge in the end.

JoergSiebahn (approved):

JoergSiebahn:
Thanks for your contribution. The release will be available within the next hours.

PR #3753, opened 2024-11-20
chore: update CODEOWNERS to assign ownership to Team tachyon

mikhailfayez:


JoergSiebahn (dismissed):

JoergSiebahn:
This was already changed. I would prefer to use the full group identifier for Backstage: `group:default/tachyon`

PR #3922, opened 2025-04-30
SCM-1392: adapted RetryProcessingErrorMLS to support a max. number of…

mibutec:
This PR add the ability to do a defined amount of retries before giving up to `RetryProcessingErrorMLS`.

The idea is:
- Count the number of processing a specific offset for a given partition
- If that number matches the max number of retries, mark increase the offset and do some error handling
- For the error handling there are now 2 errorhandlers:
  - intermediateErrorHandler is called after each failed processing
  - finalErrorHandler is called when no more retries are left

### DIFF
--- a/sda-commons-server-kafka/src/main/java/org/sdase/commons/server/kafka/consumer/strategies/retryprocessingerror/RetryCounter.java
+++ b/sda-commons-server-kafka/src/main/java/org/sdase/commons/server/kafka/consumer/strategies/retryprocessingerror/RetryCounter.java
@@ -1,0 +1,65 @@
+package org.sdase.commons.server.kafka.consumer.strategies.retryprocessingerror;
+
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.Map;
+import org.apache.kafka.clients.consumer.ConsumerRecord;
+
+/**
+ * Helper class counting the number of executions for a given record.
+ *
+ * <p>Internally it increases a counter for the current offset of the record as long as the offset
+ * remains the same. If the offset changes, the counter is reset to 0.
+ */
+public class RetryCounter {
+  private final Map<Integer, OffsetCounter> offsetCountersByPartition =
+      Collections.synchronizedMap(new HashMap<>());
+
+  private final long maxRetryCount;
+
+  public RetryCounter(long maxRetryCount) {
+    this.maxRetryCount = maxRetryCount;
+  }
+
+  public void incErrorCount(ConsumerRecord<?, ?> consumerRecord) {
+    getOffsetCounterInternal(consumerRecord).inc();
+  }
+
+  public boolean isMaxRetryCountReached(ConsumerRecord<?, ?> consumerRecord) {
+    long counter = getOffsetCounter(consumerRecord);
+    return counter > maxRetryCount;
+  }
+
+  public long getOffsetCounter(ConsumerRecord<?, ?> consumerRecord) {
+    return getOffsetCounterInternal(consumerRecord).count;
+  }
+
+  public long getMaxRetryCount() {
+    return maxRetryCount;
+  }
+
+  private OffsetCounter getOffsetCounterInternal(ConsumerRecord<?, ?> consumerRecord) {
+    OffsetCounter counter =
+        offsetCountersByPartition.computeIfAbsent(
+            consumerRecord.partition(), k -> new OffsetCounter(0));
+    if (counter.offset != consumerRecord.offset()) {
+      counter.offset = consumerRecord.offset();
+      counter.count = 0;
+    }
+    return counter;
+  }
+
+  private static class OffsetCounter {
+    private long offset = 0;
+    private long count = 0;
+
+    public OffsetCounter(long offset) {
+      this.offset = offset;
+      this.count = 0;
+    }
+
+    public long inc() {
+      return ++count;
+    }
+  }
+}
--- a/sda-commons-server-kafka/src/main/java/org/sdase/commons/server/kafka/consumer/strategies/retryprocessingerror/RetryProcessingErrorMLS.java
+++ b/sda-commons-server-kafka/src/main/java/org/sdase/commons/server/kafka/consumer/strategies/retryprocessingerror/RetryProcessingErrorMLS.java
@@ -3,6 +3,7 @@ package org.sdase.commons.server.kafka.consumer.strategies.retryprocessingerror;
 import java.time.Duration;
 import java.time.Instant;
 import java.util.Collections;
+import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import org.apache.kafka.clients.consumer.ConsumerRecord;
@@ -27,12 +28,27 @@ public class RetryProcessingErrorMLS<K, V> extends MessageListenerStrategy<K, V>
 
   private static final Logger LOGGER = LoggerFactory.getLogger(RetryProcessingErrorMLS.class);
   private final MessageHandler<K, V> handler;
-  private final ErrorHandler<K, V> errorHandler;
+  private final ErrorHandler<K, V> intermediateErrorHandler;
+  private final ErrorHandler<K, V> finalErrorHandler;
+  private final long maxRetryCount;
   private String consumerName;
+  private Map<TopicPartition, OffsetCounter> offsetCounters =
+      Collections.synchronizedMap(new HashMap<>());
 
-  public RetryProcessingErrorMLS(MessageHandler<K, V> handler, ErrorHandler<K, V> errorHandler) {
+  public RetryProcessingErrorMLS(
+      MessageHandler<K, V> handler, ErrorHandler<K, V> intermediateErrorHandler) {
+    this(handler, intermediateErrorHandler, Long.MAX_VALUE, null);
+  }
+
+  public RetryProcessingErrorMLS(
+      MessageHandler<K, V> handler,
+      ErrorHandler<K, V> intermediateErrorHandler,
+      long maxRetryCount,
+      ErrorHandler<K, V> finalErrorHandler) {
     this.handler = handler;
-    this.errorHandler = errorHandler;
+    this.intermediateErrorHandler = intermediateErrorHandler;
+    this.maxRetryCount = maxRetryCount;
+    this.finalErrorHandler = finalErrorHandler;
   }
 
   @Override
@@ -57,9 +73,7 @@ public class RetryProcessingErrorMLS<K, V> extends MessageListenerStrategy<K, V>
         try {
           Instant timerStart = Instant.now();
           handler.handle(consumerRecord);
-          // mark last successful processed record for commit
-          lastCommitOffset = new OffsetAndMetadata(consumerRecord.offset() + 1);
-          addOffsetToCommitOnClose(consumerRecord);
+          lastCommitOffset = markConsumerRecordProcessed(consumerRecord);
 
           Instant timerEnd = Instant.now();
           if (LOGGER.isTraceEnabled()) {
@@ -71,20 +85,25 @@ public class RetryProcessingErrorMLS<K, V> extends MessageListenerStrategy<K, V>
           }
 
         } catch (RuntimeException e) {
-          LOGGER.error(
-              "Error while handling record {} in message handler {}",
-              consumerRecord.key(),
-              handler.getClass(),
-              e);
-          boolean shouldContinue = errorHandler.handleError(consumerRecord, e, consumer);
-          if (!shouldContinue) {
-            throw new StopListenerException(e);
-          } else {
-            LOGGER.warn(
-                "Error while handling record {} in message handler {}, will be retried",
+          long retryCount = getOffsetCounter(partition, consumerRecord.offset()).inc();
+          if (retryCount > maxRetryCount) {
+            LOGGER.error(
+                "Error while handling record {} in message handler {}, no more retries",
                 consumerRecord.key(),
                 handler.getClass(),
                 e);
+
+            callErrorHandler(finalErrorHandler, consumerRecord, e, consumer);
+            lastCommitOffset = markConsumerRecordProcessed(consumerRecord);
+          } else {
+            LOGGER.warn(
+                "Error while handling record {} in message handler {}, will be retried (%s / %s)..."
+                    .formatted(retryCount, maxRetryCount),
+                consumerRecord.key(),
+                handler.getClass(),
+                e);
+
+            callErrorHandler(intermediateErrorHandler, consumerRecord, e, consumer);
 
             // seek to the current offset of the failing record for retry
             consumer.seek(partition, consumerRecord.offset());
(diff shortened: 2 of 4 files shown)


sda-commons-server-kafka/src/main/java/org/sdase/commons/server/kafka/consumer/strategies/retryprocessingerror/RetryProcessingErrorMLS.java
```diff
@@ -71,21 +85,26 @@ private void processRecordsByPartition(
...
           } else {
             LOGGER.warn(
-                "Error while handling record {} in message handler {}, will be retried",
+                "Error while handling record {} in message handler {}, will be retried (%s / %s)..."
+                    .formatted(retryCount, maxRetryCount),
                 consumerRecord.key(),
                 handler.getClass(),
                 e);
 
+            callErrorHandler(intermediateErrorHandler, consumerRecord, e, consumer);
```
JoergSiebahn, line 106: If I understand this correctly, the error handler can not decide to stop consuming any more. Also the final [error handler](https://github.com/SDA-SE/sda-dropwizard-commons/pull/3922/files#diff-8fb386b215a69d009a70fdc9ae129fe6e8a350b2dc49dd46caeaaf286df31073R96) can only do something to mitigate the error but can't decide that something went very wrong and it's better to stop.

sda-commons-server-kafka/src/main/java/org/sdase/commons/server/kafka/consumer/strategies/retryprocessingerror/RetryProcessingErrorMLS.java
```diff
@@ -27,12 +28,27 @@ public class RetryProcessingErrorMLS<K, V> extends MessageListenerStrategy<K, V>
...
+  private final ErrorHandler<K, V> finalErrorHandler;
+  private final long maxRetryCount;
   private String consumerName;
+  private Map<TopicPartition, OffsetCounter> offsetCounters =
+      Collections.synchronizedMap(new HashMap<>());
 
-  public RetryProcessingErrorMLS(MessageHandler<K, V> handler, ErrorHandler<K, V> errorHandler) {
+  public RetryProcessingErrorMLS(
+      MessageHandler<K, V> handler, ErrorHandler<K, V> intermediateErrorHandler) {
+    this(handler, intermediateErrorHandler, Long.MAX_VALUE, null);
```
JoergSiebahn, line 40: Probably very unlike to reach max long, but what about using `-1` as _forever_?

Do you know if there is any increasing delay for the next try? Afaik the default polling is quite quick, so we could reach high numbers (not max long) early - maybe before any subsequent failing service has time to recover.

JoergSiebahn: Found it: https://github.com/SDA-SE/sda-dropwizard-commons/blob/52395ce450c4674d1cf9dbba0bc3a4ebce5e7a37/sda-commons-server-kafka/src/main/java/org/sdase/commons/server/kafka/consumer/MessageListener.java#L180

sda-commons-server-kafka/src/main/java/org/sdase/commons/server/kafka/consumer/strategies/retryprocessingerror/RetryProcessingErrorMLS.java
mibutec: I just moved the existing code into a method. Functionality stays the same:

```
private void callErrorHandler(...) {
    if (errorHandler != null) {
      boolean shouldContinue = errorHandler.handleError(consumerRecord, e, consumer);
      if (!shouldContinue) {
        throw new StopListenerException(e);
      }
    }
  }
```

sda-commons-server-kafka/src/main/java/org/sdase/commons/server/kafka/consumer/strategies/retryprocessingerror/RetryProcessingErrorMLS.java
```diff
@@ -27,12 +28,27 @@ public class RetryProcessingErrorMLS<K, V> extends MessageListenerStrategy<K, V>
...
+  }
+
+  public RetryProcessingErrorMLS(
+      MessageHandler<K, V> handler,
+      ErrorHandler<K, V> intermediateErrorHandler,
+      long maxRetryCount,
+      ErrorHandler<K, V> finalErrorHandler) {
```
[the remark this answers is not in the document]
mibutec: fine for me, I can change it

sda-commons-server-kafka/src/main/java/org/sdase/commons/server/kafka/consumer/strategies/retryprocessingerror/RetryProcessingErrorMLS.java
mibutec: When using `-1` I need a special handling for that value. Using MAX_VALUE I can just use the default logic

sda-commons-server-kafka/src/main/java/org/sdase/commons/server/kafka/consumer/strategies/retryprocessingerror/RetryProcessingErrorMLS.java
JoergSiebahn: Ah no. It's moved to the callErrorHandler function.

sda-commons-server-kafka/src/main/java/org/sdase/commons/server/kafka/consumer/strategies/retryprocessingerror/RetryProcessingErrorMLS.java
```diff
@@ -105,4 +124,46 @@ public void verifyConsumerConfig(Map<String, String> config) {
...
+    }
+  }
+
+  private OffsetAndMetadata markConsumerRecordProcessed(ConsumerRecord<K, V> consumerRecord) {
+    addOffsetToCommitOnClose(consumerRecord);
+    return new OffsetAndMetadata(consumerRecord.offset() + 1);
+  }
+
+  private OffsetCounter getOffsetCounter(TopicPartition partition, long offset) {
+    OffsetCounter counter = offsetCounters.computeIfAbsent(partition, k -> new OffsetCounter(0));
```
JoergSiebahn, line 147: My understanding: By tracking and comparing the offset, you effectively realize when a retried record is skipped after reaching the retry limit or processed successfully. I think for me it would be easier to understand, if we don't take care about the offset here and just remove the partition from the map on success or skip explicitly.

sda-commons-server-kafka/src/main/java/org/sdase/commons/server/kafka/consumer/strategies/retryprocessingerror/RetryProcessingErrorMLS.java
```diff
@@ -71,21 +85,26 @@ private void processRecordsByPartition(
...
+
+            callErrorHandler(finalErrorHandler, consumerRecord, e, consumer);
+            lastCommitOffset = markConsumerRecordProcessed(consumerRecord);
           } else {
             LOGGER.warn(
-                "Error while handling record {} in message handler {}, will be retried",
+                "Error while handling record {} in message handler {}, will be retried (%s / %s)..."
+                    .formatted(retryCount, maxRetryCount),
                 consumerRecord.key(),
                 handler.getClass(),
```
JoergSiebahn, line 103: Would it work with normal replacement of SLF4J as well?
```suggestion
                "Error while handling record {} in message handler {}, will be retried ({} / {})...",
                consumerRecord.key(),
                handler.getClass(),
                retryCount,
                maxRetryCount,
```

sda-commons-server-kafka/src/main/java/org/sdase/commons/server/kafka/consumer/strategies/retryprocessingerror/RetryProcessingErrorMLS.java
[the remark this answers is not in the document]
mibutec: done

sda-commons-server-kafka/src/main/java/org/sdase/commons/server/kafka/consumer/strategies/retryprocessingerror/RetryProcessingErrorMLS.java
```diff
@@ -27,12 +28,27 @@ public class RetryProcessingErrorMLS<K, V> extends MessageListenerStrategy<K, V>
 
   private static final Logger LOGGER = LoggerFactory.getLogger(RetryProcessingErrorMLS.class);
   private final MessageHandler<K, V> handler;
-  private final ErrorHandler<K, V> errorHandler;
+  private final ErrorHandler<K, V> intermediateErrorHandler;
+  private final ErrorHandler<K, V> finalErrorHandler;
+  private final long maxRetryCount;
   private String consumerName;
+  private Map<TopicPartition, OffsetCounter> offsetCounters =
```
JoergSiebahn, line 35: Does `TopicPartition` have an appropriate equals method, so that the partition is matched reliable? To be safe, we could just go with the number.

NIT: I prefer more precise names for such maps, e.g. `offsetCountersByPartition`.

sda-commons-server-kafka/src/main/java/org/sdase/commons/server/kafka/consumer/strategies/retryprocessingerror/RetryProcessingErrorMLS.java
mibutec: When taking the offset into account, I can get the current count just out of the context, without the help of any other code.

The other way I have to rely on the code finishing message processing to reset the current count. I think this way it is more robust

sda-commons-server-kafka/src/main/java/org/sdase/commons/server/kafka/consumer/strategies/retryprocessingerror/RetryProcessingErrorMLS.java
```diff
@@ -27,12 +28,27 @@ public class RetryProcessingErrorMLS<K, V> extends MessageListenerStrategy<K, V>
...
+      Collections.synchronizedMap(new HashMap<>());
 
-  public RetryProcessingErrorMLS(MessageHandler<K, V> handler, ErrorHandler<K, V> errorHandler) {
+  public RetryProcessingErrorMLS(
+      MessageHandler<K, V> handler, ErrorHandler<K, V> intermediateErrorHandler) {
+    this(handler, intermediateErrorHandler, Long.MAX_VALUE, null);
+  }
+
+  public RetryProcessingErrorMLS(
```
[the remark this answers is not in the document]
mibutec: Good point

sda-commons-server-kafka/src/main/java/org/sdase/commons/server/kafka/consumer/strategies/retryprocessingerror/RetryProcessingErrorMLS.java
mibutec: Good point, done

sda-commons-server-kafka/src/main/java/org/sdase/commons/server/kafka/consumer/strategies/retryprocessingerror/RetryProcessingErrorMLS.java
mibutec: Yes, it has

sda-commons-server-kafka/src/main/java/org/sdase/commons/server/kafka/consumer/strategies/retryprocessingerror/RetryProcessingErrorMLS.java
JoergSiebahn, line 40: Would it make sense to use the intermediateErrorHandler as finalErrorHandler here as well? You may run in (theoretical) NullPointerExceptions when passing null.

sda-commons-server-kafka/src/main/java/org/sdase/commons/server/kafka/consumer/strategies/retryprocessingerror/RetryProcessingErrorMLS.java
JoergSiebahn: Valid point. I wonder, if it is just me who needs too much time, to understand what this code is doing. Maybe we can help there a bit, e.g. JavaDoc or inline comments for the method and/or the map, more descriptive method name (what about `getRetryCounterForOffset`?) and class name (what about `OffsetAwareRetryCounter`? The current name counts offsets in my mind).

JoergSiebahn: It may also be an option to extract the whole tracking (this logic and the map) and maybe the retry limit exceeded decision in a dedicated class with a clean API. But I have no idea yet, how that could look like.

mibutec: I refactred the count-logic to another class and improved comments. Let's see if that helps to make things more readable.
I didn't change the name to keep things backward compatible.

sda-commons-server-kafka/src/main/java/org/sdase/commons/server/kafka/consumer/strategies/retryprocessingerror/RetryProcessingErrorMLS.java
```diff
@@ -21,18 +21,57 @@
...
+ *
+ * <p>After each error an error handler can decide if processing should be retried or stopped.
+ *
+ * <p>If the retry count exceeds:
+ *
+ * <ul>
+ *   <li>The failed message handling will be logged as ERROR
+ *   <li>a retryLimitExceededErrorHandler will be called allowing to do more error handling (i.e.
+ *       write to DLT)
+ *   <li>the offset will get increased, so the message handling gets ignored.
```
JoergSiebahn, line 34: Isn't there an exception (no offset increase) if the `retryLimitExceededErrorHandler` decides to stop processing?

sda-commons-server-kafka/src/main/java/org/sdase/commons/server/kafka/consumer/strategies/retryprocessingerror/RetryCounter.java
```diff
@@ -0,0 +1,65 @@
...
+import java.util.Map;
+import org.apache.kafka.clients.consumer.ConsumerRecord;
+
+/**
+ * Helper class counting the number of executions for a given record.
+ *
+ * <p>Internally it increases a counter for the current offset of the record as long as the offset
+ * remains the same. If the offset changes, the counter is reset to 0.
+ */
+public class RetryCounter {
```
JoergSiebahn, line 14: Thanks. For me it's easier to understand although the implementation is basically the same.

JoergSiebahn (commented):
Just some nitpicking comments.

sda-commons-server-kafka/src/main/java/org/sdase/commons/server/kafka/consumer/strategies/retryprocessingerror/RetryProcessingErrorMLS.java
mibutec: Technically you are right, the offset is conditional and will not be increased, if the `retryLimitExceededErrorHandler` throws an exception or returns `false`. For me it's part of the "a retryLimitExceededErrorHandler will be called allowing to do more error handling"

JoergSiebahn (approved):

PR #3947, opened 2025-05-23
ci: update info.version in openapi.yaml with semver-release

piotrkupiec:
Add new utility method so that the api version is ignored when comparing the files


hans4dev (dismissed):

piotrkupiec:
The need of this functionality will be discussed at a later time.

PR #3971, opened 2025-06-11
feat(kafka): Make the number of retries configurable in the ListenerConfig

nicowesterheide-sda:
Make the number of retries configurable in the ListenerConfig

### DIFF
--- a/sda-commons-server-kafka/src/main/java/org/sdase/commons/server/kafka/config/ListenerConfig.java
+++ b/sda-commons-server-kafka/src/main/java/org/sdase/commons/server/kafka/config/ListenerConfig.java
@@ -9,6 +9,7 @@ public class ListenerConfig {
   private long pollInterval = 100;
   private long pollIntervalFactorOnError = 4;
   private long maxPollInterval = 25_000;
+  private long maxRetries = Long.MAX_VALUE;
 
   private ListenerConfig() {
     // empty constructor for jackson
@@ -62,12 +63,21 @@ public class ListenerConfig {
     this.maxPollInterval = maxPollInterval;
   }
 
+  public long getMaxRetries() {
+    return maxRetries;
+  }
+
+  public void setMaxRetries(long maxRetries) {
+    this.maxRetries = maxRetries;
+  }
+
   public static class ListenerConfigBuilder {
 
     private long topicMissingRetryMs = 0;
     private long pollInterval = 100;
     private long pollIntervalFactorOnError = 4;
     private long maxPollInterval = 25_000;
+    private long maxRetries = Long.MAX_VALUE;
 
     public ListenerConfigBuilder withTopicMissingRetryMs(@NotNull long ms) {
       this.topicMissingRetryMs = ms;
@@ -89,6 +99,11 @@ public class ListenerConfig {
       return this;
     }
 
+    public ListenerConfigBuilder withMaxRetries(@NotNull int maxRetries) {
+      this.maxRetries = maxRetries;
+      return this;
+    }
+
     public ListenerConfig build(@NotNull int numberInstances) {
       ListenerConfig build = new ListenerConfig();
       build.setTopicMissingRetryMs(topicMissingRetryMs);
@@ -96,6 +111,7 @@ public class ListenerConfig {
       build.setMaxPollInterval(maxPollInterval);
       build.setPollIntervalFactorOnError(pollIntervalFactorOnError);
       build.setInstances(numberInstances);
+      build.setMaxRetries(maxRetries);
       return build;
     }
   }
--- a/sda-commons-server-kafka/src/main/java/org/sdase/commons/server/kafka/consumer/strategies/MessageListenerStrategy.java
+++ b/sda-commons-server-kafka/src/main/java/org/sdase/commons/server/kafka/consumer/strategies/MessageListenerStrategy.java
@@ -33,7 +33,7 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 /**
- * Base for MessageListener strategies. A strategy defies the way how records are handled and how
+ * Base for MessageListener strategies. A strategy defines the way how records are handled and how
  * the consumer interacts with the kafka broker, i.e. the commit handling.
  *
  * <p>A strategy might support the usage of a @{@link MessageHandler} that encapsulates the business
@@ -45,12 +45,13 @@ import org.slf4j.LoggerFactory;
 public abstract class MessageListenerStrategy<K, V> {
 
   private static final Logger LOGGER = LoggerFactory.getLogger(MessageListenerStrategy.class);
-
   private Map<TopicPartition, OffsetAndMetadata> offsetsToCommitOnClose = new HashMap<>();
   private Set<String> metadataFields = Set.of();
 
   /**
-   * @param metadataFields the configured {@link MetadataContext#metadataFields()}
+   * Initializes the MessageListenerStrategy with metadata fields and a maximum retry count 0.
+   *
+   * @param metadataFields a set of metadata field names to be considered when processing records
    */
   public void init(Set<String> metadataFields) {
     this.metadataFields = Optional.ofNullable(metadataFields).orElse(Set.of());
@@ -193,4 +194,11 @@ public abstract class MessageListenerStrategy<K, V> {
   public Map<String, String> forcedConfigToApply() {
     return Collections.emptyMap();
   }
+
+  /**
+   * Creates the RetryCounter with the given value, if applicable for the strategy.
+   *
+   * @param maxRetriesCount max retries value from config
+   */
+  public abstract void setRetryCounterIfApplicable(long maxRetriesCount);
 }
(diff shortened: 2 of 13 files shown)


sda-commons-server-kafka/src/main/java/org/sdase/commons/server/kafka/config/ListenerConfig.java
```diff
@@ -9,6 +9,7 @@ public class ListenerConfig {
   private long pollInterval = 100;
   private long pollIntervalFactorOnError = 4;
   private long maxPollInterval = 25_000;
+  private int maxRetries = 5;
```
mibutec, line 12: In the past, the behavior of RetryProcessingErrorMLS was to retry invinitly. Adding the retry count and abort retrying is a new feature.

To maintain backward compatibility, we should make the default Long.MAX_VALUE,

nicowesterheide-sda: fixed

sda-commons-server-kafka/src/main/java/org/sdase/commons/server/kafka/consumer/strategies/MessageListenerStrategy.java
```diff
@@ -45,17 +46,30 @@
...
    */
   public void init(Set<String> metadataFields) {
     this.metadataFields = Optional.ofNullable(metadataFields).orElse(Set.of());
   }
 
+  /**
+   * Initializes the MessageListenerStrategy with metadata fields and a maximum retry count.
+   *
+   * @param metadataFields a set of metadata field names to be considered when processing records
+   * @param maxRetriesCount the maximum number of retries allowed for failed operations
```
JoergSiebahn, line 66: NIT: 

```suggestion
@param maxRetriesCount the maximum number of retries allowed for failed operations. Only used by strategies that support retry.
```

sda-commons-server-kafka/src/main/java/org/sdase/commons/server/kafka/consumer/strategies/retryprocessingerror/RetryProcessingErrorMLS.java
```diff
@@ -51,27 +51,25 @@ public class RetryProcessingErrorMLS<K, V> extends MessageListenerStrategy<K, V>
...
+   * @param errorHandler the error handler called after each error, can be null
+   * @param retryLimitExceededErrorHandler the error handler called if the retry limit is exceeded,
+   *     can be null
    */
   public RetryProcessingErrorMLS(
       MessageHandler<K, V> handler,
-      ErrorHandler<K, V> errorHandler,
-      long maxRetryCount,
-      ErrorHandler<K, V> retryLimitExceededErrorHandler) {
+      @Nullable ErrorHandler<K, V> errorHandler,
```
JoergSiebahn, line 68: Technically in this implementation, the default error handler can be null. But looking at [the original implementation](https://github.com/SDA-SE/sda-dropwizard-commons/commit/32a2c88c0fd409fbc668e4befecaf2ee898a2cff#diff-8fb386b215a69d009a70fdc9ae129fe6e8a350b2dc49dd46caeaaf286df31073L79) it should not be. Users should think about their error handling. Imho the error handler for retry limit exceeded is just nullable for backward compatibility. As non-null was not enforced before, I would not add such a check, but I would also not document the technical implementation detail that is not a good choice.

sda-commons-server-kafka/src/main/java/org/sdase/commons/server/kafka/consumer/strategies/retryprocessingerror/RetryProcessingErrorMLS.java
```diff
@@ -51,27 +51,25 @@ public class RetryProcessingErrorMLS<K, V> extends MessageListenerStrategy<K, V>
...
+      @Nullable ErrorHandler<K, V> retryLimitExceededErrorHandler) {
```
JoergSiebahn, line 69: Changing the signature of the constructor is a breaking change. Can you introduce a new constructor and make this one deprecated? That will cause a bit of weird temporary code to prefer this maxRetryCount over the configured value from the init method for full backward compatibility. So it should be documented with the deprecation what should be removed when we finally remove this constructor.

nicowesterheide-sda: I was not sure how this should be handled for breaking changes, in the end I decided to remove it. I will restore the old constructor and make it deprecated then additionally to the new one.

sda-commons-server-kafka/src/main/java/org/sdase/commons/server/kafka/consumer/strategies/retryprocessingerror/RetryProcessingErrorMLS.java
nicowesterheide-sda: I added this because it was mentioned in the ticket description, but I can remove it again of course.

sda-commons-server-kafka/src/main/java/org/sdase/commons/server/kafka/consumer/strategies/MessageListenerStrategy.java
```diff
@@ -45,17 +46,30 @@
 public abstract class MessageListenerStrategy<K, V> {
 
   private static final Logger LOGGER = LoggerFactory.getLogger(MessageListenerStrategy.class);
-
+  protected RetryCounter retryCounter;
```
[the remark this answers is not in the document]
nicowesterheide-sda: The problem of having the RetryCounter in the RetryProcessingErrorMLS is that we cannot set it from the init method, or we would need some instanceof checks either in the KafkaBundle or the MessageListener for example, which is also not really nice. Because the strategy is instantiated before we access the ListenerConfig, we have to create the RetryCounter afterwards and set it in the RetryProcessingErrorMLS object somehow.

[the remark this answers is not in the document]
JoergSiebahn: What do you think of a protected no-op method in `MessageListenerStrategy` to overwrite in implementations that support it: `setRequestCounterIfApplicable(...)`? Instead of no-op it can also log _"retry not supported by {this.getClass}: not using maxRetryCount" (info only as it is a common case).

In the retrying strategy that can also check and ignore and maybe log an info about the new configuration if retry was set via the now deprecated constructor.

sda-commons-server-kafka/src/main/java/org/sdase/commons/server/kafka/consumer/strategies/retryprocessingerror/RetryProcessingErrorMLS.java
JoergSiebahn: In general we want to avoid breaking changes. A new major always scares the consumers and even if you are not affected, you need more time to double check that you are not affected. Also we prefer to not break something without a deprecation and an alternative in a version before the breaking change. That supports a graceful migration for the consumers. At one point, we can't avoid something breaking - mostly du to major upgrades of one of the frameworks. Then It's time to also resolve all the deprecations.

sda-commons-server-kafka/src/main/java/org/sdase/commons/server/kafka/consumer/strategies/retryprocessingerror/RetryProcessingErrorMLS.java
mibutec: To close the circle :D: I added it because of your comment in https://github.com/SDA-SE/malware-scanner-facade/pull/1211 (MalwareScannerFacadeApp.java)

sda-commons-server-kafka/src/main/java/org/sdase/commons/server/kafka/consumer/strategies/MessageListenerStrategy.java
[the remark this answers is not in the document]
nicowesterheide-sda: I updated the implementation, the RetryCounter is back in the RetryProcessingMLS and there is a new abstract method for setting the retry counter if applicable which can be overridden by the concrete implementations to set the retry counter if the strategy supports it.

sda-commons-server-kafka/src/main/java/org/sdase/commons/server/kafka/consumer/strategies/retryprocessingerror/RetryProcessingErrorMLS.java
JoergSiebahn: 😅 Ok, hmmm, I didn't remember. Seems like I vote for nullable as well then. 😂

sda-commons-server-kafka/src/main/java/org/sdase/commons/server/kafka/consumer/strategies/retryprocessingerror/RetryProcessingErrorMLS.java
```diff
@@ -51,29 +52,51 @@ public class RetryProcessingErrorMLS<K, V> extends MessageListenerStrategy<K, V>
    * @param errorHandler the error handler called after each error
    */
   public RetryProcessingErrorMLS(MessageHandler<K, V> handler, ErrorHandler<K, V> errorHandler) {
-    this(handler, errorHandler, Long.MAX_VALUE, null);
+    this(handler, errorHandler, null);
```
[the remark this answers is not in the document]
nicowesterheide-sda: added

PR #4095, opened 2025-09-22
fix(deps): bump org.springframework.data:spring-data-mongodb from 4.1…

Solemn1:
….12 to 4.5.4

Hello SDA SE,

Mend whitesource scan detected a vulnerability of high severity in the spring libraries (spring-core-6.0.20.jar), therefore we need to bump the spring data mongo dependency.

### DIFF
--- a/sda-commons-dependencies/build.gradle
+++ b/sda-commons-dependencies/build.gradle
@@ -120,7 +120,7 @@ dependencies {
     api "it.unibo.tuprolog:parser-theory-jvm:$tuprologVersion"
 
     // sda-commons-server-mongo-testing, sda-commons-server-spring-data-mongo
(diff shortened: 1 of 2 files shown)


YellowFlora:
> > @Solemn1 i had to recreate the branch because our workflows had an error: #4096
> 
> Hi, thank you @YellowFlora . When do you think you can merge? Unfortunately, the vulnerability is blocking our work. If you’re unable to merge, we’ll need to find another solution.

Hi @Solemn1, we had a few intern discussions yesterday and we will not be able to update spring data mongodb until the next major is released. We tested the dropwizard update yesterday in other services of us, and some broke because of some changes that were made in spring data mongodb.

Instead we will update either spring security or directly spring-core. We will be able to provide an update tomorrow morning probably. So we can fix the CVE atleast. 

Sorry the delay.

PR #4340, opened 2026-03-25
ci: prepare ci pipeline for r8 branch flow

YellowFlora:


.github/workflows/cve-check.yml
christoph-mueller-sda-se, line 1: Should we not keep an CVE check for R6?
(Similar to sda-spring-boot-commons/blob/main/.github/workflows/cve-check-r6.yml)

christoph-mueller-sda-se (approved):

.github/workflows/cve-check.yml
christoph-mueller-sda-se: ok, found `.github/workflows/cve-check-r8.yml` in `feat/drop_java_17` (PR-4341)

PR #4383, opened 2026-04-20
test: verify authorization is not affected by CVE-2026-2332

JoergSiebahn:


YellowFlora:
see https://github.com/SDA-SE/sda-dropwizard-commons/pull/4385 instead

PR #4385, opened 2026-04-20
test: verify authorization is not affected by CVE-2026-2332

YellowFlora:


YellowFlora:
[ignore-policy.rego](https://github.com/SDA-SE/sda-dropwizard-commons/pull/4385/changes#diff-b072ac81c042ceda99ba4a9e588dc9254ea9c37bed1727df018ed009ace33216) was deleted accidentally in another pr

sda-commons-starter-example/src/test/java/org/sdase/commons/starter/example/VerifyEffectOfCve20262332IT.java
```diff
@@ -0,0 +1,123 @@
...
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Order;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.extension.RegisterExtension;
+import org.sdase.commons.server.auth.testing.AuthClassExtension;
+import org.sdase.commons.server.opa.testing.OpaClassExtension;
+import org.sdase.commons.starter.SdaPlatformConfiguration;
+import org.sdase.commons.starter.example.people.db.TestDataUtil;
+
+class VerifyEffectOfCve20262332IT {
```
christoph-mueller-sda-se, line 36: Add a comment mentioning the CVE verbatim, so it can be found when searching, e.g.
```suggestion
/** to make sure we are not vulnerable to CVE-2026-2332 */
class VerifyEffectOfCve20262332IT {
```

sda-commons-starter-example/src/test/java/org/sdase/commons/starter/example/VerifyEffectOfCve20262332IT.java
```diff
@@ -0,0 +1,123 @@
...
+
+  @BeforeEach
+  void setupTestData() {
+    TestDataUtil.clearTestData();
+    TestDataUtil.addPersonEntity("john-doe", "John", "Doe");
+    OPA.reset();
+  }
+
+  @Test
+  @SuppressWarnings("java:S3457")
```
christoph-mueller-sda-se, line 63: This suppression seems unnecessary here.
There is no logging and no string concatenation.
(2 occurrences in this file)

sda-commons-starter-example/src/test/java/org/sdase/commons/starter/example/VerifyEffectOfCve20262332IT.java
```diff
@@ -0,0 +1,123 @@
...
+      String line;
+      while ((line = br.readLine()) != null) {
+        System.out.println(line);
+        response.append(line);
+      }
+      assertThat(response).contains("Location", "people");
+    }
+  }
+
+  @Test
```
christoph-mueller-sda-se, line 94: add some explanation for the future reader, e.g.

```suggestion
/**
 * Verifies the practical effect of CVE-2026-2332 (Jetty chunk-extension quoted-string parsing)
 * in this application setup.
 *
 * <p>The crafted payload can cause Jetty 11.0.26 to split one TCP payload into two HTTP requests
 * (a malformed POST plus a smuggled GET), but the smuggled request is still processed through the
 * normal auth/OPA checks. In this test, GET /people remains denied, so no authorization bypass or
 * data leak is observed even though the underlying Jetty parser version is in the advisory range.
 */
  @Test
```

sda-commons-starter-example/src/test/java/org/sdase/commons/starter/example/VerifyEffectOfCve20262332IT.java
```diff
@@ -0,0 +1,123 @@
...
+      pw.print("\r\n");
+      pw.flush();
+      BufferedReader br = new BufferedReader(new InputStreamReader(s.getInputStream()));
+      StringBuilder response = new StringBuilder();
+      String line;
+      while ((line = br.readLine()) != null) {
+        System.out.println(line);
+        response.append(line);
+      }
+      assertThat(response).doesNotContain("John").contains("Not authorized");
```
christoph-mueller-sda-se: we should also verify we really got two responses

```suggestion
      assertThat(countOccurrences(response, "HTTP/1.1 ")).isEqualTo(2);
      assertThat(response)
          .contains("HTTP/1.1 400 Bad Request")
          .contains("HTTP/1.1 403 Forbidden")
          .doesNotContain("John")
          .contains("Not authorized");
```

sda-commons-starter-example/src/test/java/org/sdase/commons/starter/example/VerifyEffectOfCve20262332IT.java
JoergSiebahn: Sonar raised the issue for using `\n` in `printf`. It requires to use `%n` which converts to platform specific line break with `printf`. That is not suitable here because HTTP requires `\r\n` on all platforms.

sda-commons-starter-example/src/test/java/org/sdase/commons/starter/example/VerifyEffectOfCve20262332IT.java
JoergSiebahn: With so many comments for the test that was just created as a reproducer, I wonder if it's good enough to remove the test from this PR, add a bit more context to the ignore policy and include a link to the original PR in the docs.
WDYT?

christoph-mueller-sda-se: I would keep the test. It ensures our assumption and understanding on every CI run without measurable extra cost.

christoph-mueller-sda-se (approved):

PR #4466, opened 2026-06-18
fix(deps): bump com.fasterxml.jackson:jackson-bom to 2.21.4 and tools.jackson:jackson-bom to 3.1.4

piotrkupiec:


YellowFlora (changes requested):
Could you fix the commit messages please? Drop the first commit completly and split the other commit into 2 commits? Then we can rebase merge with a clearer history

YellowFlora (approved):

PR #4477, opened 2026-06-22
feat: add beforePoll hook to MessageListenerStrategy

deus0ne:
**What is the problem?**
Commit ee80de4 changed the MessageListener to only invoke `strategy.processRecords()` when `records.count() > 0`.
This prevents advanced strategies from performing periodic tasks on every poll cycle, such as reacting to external control signals (e.g., `pause()` or `seek()`) or managing internal state.

**How does this PR solve the problem?**
This PR introduces a new `beforePoll(KafkaConsumer<K, V> consumer)` hook to the `MessageListenerStrategy`. It is invoked in every `MessageListener` loop, right before `consumer.poll()`. The base implementation is empty, making this change fully backward-compatible.

This provides a clean, dedicated API for such pre-poll actions, removing the need for strategies to implement complex or inefficient workarounds.

### DIFF
--- a/sda-commons-server-kafka/src/main/java/org/sdase/commons/server/kafka/consumer/MessageListener.java
+++ b/sda-commons-server-kafka/src/main/java/org/sdase/commons/server/kafka/consumer/MessageListener.java
@@ -79,6 +79,9 @@ public class MessageListener<K, V> implements Runnable {
         // wait for the next poll interval
         waitForNextPoll();
 
+        // allow to run optional pre-poll logic
+        strategy.beforePoll(consumer);
+
         ConsumerRecords<K, V> records =
             consumer.poll(Duration.ofMillis(configuredPollIntervalMillis));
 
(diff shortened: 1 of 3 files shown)


JoergSiebahn:
Hi @deus0ne, thanks for you PR. It's an interesting case that I did not think about before.

Could you please run `spotless:apply` for code formatting and amend the changes?

Out of curiosity: I think, there was no way before the mentioned change to do anything before the poll, just right after the poll when message handling starts. In need of pause or seek, what did you do? Discard the polled messages and then work with the consumer? Could it be that a custom strategy with such needs becomes much simpler with the new hook?

deus0ne:
We have our own 'offset reset' and 'downtime' features, and for both, we used a seek()-based strategy to discard and re-poll the batch. This worked even when no records were polled because processRecords() always ran.

With the beforePoll hook, we can use both features again as intended. An offset reset with seek can now be applied immediately to the very next poll. It also lets us simplify the downtime logic: we can now use pause() before the poll, which causes it to return an empty list. This is far more efficient than repeatedly seeking back the same records while a pause is active.

JoergSiebahn (approved):

PR #4478, opened 2026-06-25
feat: support nested field paths in field filtering

lisahaase-sdase:


YellowFlora (changes requested):
NTH: add regression test for nested object without @EnableFieldFilter to pin compatibility behavior.

MUST: document that nested field paths are only applied inside nested object types annotated with @EnableFieldFilter; otherwise selecting a nested field keeps the whole nested object.

MUST: fix Sonar issues.

YellowFlora (approved):

PR #4491, opened 2026-07-02
feat(sda-commons-client-jersey): add oidc startup validation toggle

sabkuehn:
Add OIDC_STARTUP_VALIDATION_DISABLED to control startup-time validation of OIDC token retrieval. The default keeps validation disabled for backward compatibility.

Motivation: error message was previously lost in the business request
Before this change, there was no startup validation for the OIDC configuration. When the token endpoint was misconfigured or unreachable, the failure only surfaced during an actual business request and even then, only as a missing Authorization header with a warning log, making the root cause hard to diagnose.
This PR adds startup validation (startupValidationDisabled = false) so that a misconfigured OIDC setup fails fast and loudly at application startup.

### DIFF
--- a/docs/client-jersey.md
+++ b/docs/client-jersey.md
@@ -256,6 +256,7 @@ Example config for **production** to be used with environment variables of the c
 ```yaml
 oidc:
   disabled: ${OIDC_DISABLED:-false}
+  enableStartupValidation: ${OIDC_ENABLE_STARTUP_VALIDATION:-false}
   grantType: ${OIDC_GRANT_TYPE:-client_credentials}
   clientId: ${OIDC_CLIENT_ID}
   clientSecret: ${OIDC_CLIENT_SECRET}
@@ -269,6 +270,13 @@ oidc:
 - _OIDC_DISABLED_
   * Disable retrieving a new token completely. Not meant for production! 
   * Example: `false`
+
+- _OIDC_ENABLE_STARTUP_VALIDATION_
+  * Enables startup validation of the OIDC configuration by performing an initial access-token retrieval.
+    Set to `true` to fail fast during startup if the OIDC provider is unreachable or misconfigured.
+  * Note: this only verifies that a token can be retrieved — it does **not** check whether the token
+    has the required scopes or permissions to access the target resource server.
+  * Default: `false`
 
 - _OIDC_GRANT_TYPE_
   * Sets the OIDC grant type.
--- a/sda-commons-client-jersey/src/main/java/org/sdase/commons/client/jersey/oidc/OidcConfiguration.java
+++ b/sda-commons-client-jersey/src/main/java/org/sdase/commons/client/jersey/oidc/OidcConfiguration.java
@@ -48,6 +48,8 @@ public class OidcConfiguration {
 
   private String scope;
 
+  private boolean enableStartupValidation;
+
   private CacheConfiguration cache = new CacheConfiguration();
 
   /**
@@ -127,6 +129,15 @@ public class OidcConfiguration {
 
   public OidcConfiguration setScope(String scope) {
     this.scope = scope;
+    return this;
+  }
+
+  public boolean isEnableStartupValidation() {
+    return enableStartupValidation;
+  }
+
+  public OidcConfiguration setEnableStartupValidation(boolean enableStartupValidation) {
+    this.enableStartupValidation = enableStartupValidation;
     return this;
   }
 
--- a/sda-commons-client-jersey/src/main/java/org/sdase/commons/client/jersey/oidc/filter/OidcRequestFilter.java
+++ b/sda-commons-client-jersey/src/main/java/org/sdase/commons/client/jersey/oidc/filter/OidcRequestFilter.java
@@ -22,17 +22,21 @@ public class OidcRequestFilter implements ClientRequestFilter {
 
   private static final Logger LOGGER = LoggerFactory.getLogger(OidcRequestFilter.class);
 
+  private final OidcConfiguration oidcConfiguration;
+
   private OidcClient oidcClient;
 
   private AuthHeaderClientFilter authHeaderClientFilter;
 
   public OidcRequestFilter(
       ClientFactory clientFactory, OidcConfiguration oidc, boolean authenticationPassthrough) {
+    this.oidcConfiguration = oidc;
     if (authenticationPassthrough) {
       authHeaderClientFilter = new AuthHeaderClientFilter();
     }
     if (!oidc.isDisabled()) {
       this.oidcClient = new OidcClient(clientFactory, oidc);
+      this.validateOidcConfig();
     }
   }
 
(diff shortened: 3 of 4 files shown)


docs/client-jersey.md
```diff
@@ -270,6 +271,11 @@ oidc:
   * Disable retrieving a new token completely. Not meant for production! 
   * Example: `false`
 
+- _OIDC_STARTUP_VALIDATION_DISABLED_
```
JoergSiebahn: In my opinion it's more intuitive when booleans have a default of `false`. What do you think about `enableStartupValidation`? I'm open for other suggestions.

sda-commons-client-jersey/src/main/java/org/sdase/commons/client/jersey/oidc/filter/OidcRequestFilter.java
```diff
@@ -41,12 +45,59 @@ public OidcRequestFilter(
...
+   */
+  public void validateOidcConfig() {
+    if (oidcConfiguration.isStartupValidationDisabled()) {
+      return;
     }
+
+    OidcResult oidcResult;
+    try {
+      oidcResult = oidcClient.createAccessToken();
+    } catch (Exception e) {
```
JoergSiebahn: `createAccessToken` does not declare any exception. So we could catch `RuntimeException` here.


```suggestion
    } catch (RuntimeException e) {
```

sda-commons-client-jersey/src/main/java/org/sdase/commons/client/jersey/oidc/OidcConfiguration.java
```diff
@@ -48,6 +48,8 @@ public class OidcConfiguration {
 
   private String scope;
 
+  private boolean startupValidationDisabled = true;
```
JoergSiebahn: So above:

```suggestion
  private boolean enableStartupValidation;
```

docs/client-jersey.md
```diff
@@ -270,6 +271,11 @@ oidc:
...
+  * Disables startup validation of the OIDC configuration by skipping an initial access-token retrieval.
+  * Set to `false` to fail fast during startup if the OIDC provider is unreachable or misconfigured.
```
JoergSiebahn: Is it worth to mention, that this validation only validates token retrieval but not if the token is suitable to access the resource server?

sda-commons-client-jersey/src/main/java/org/sdase/commons/client/jersey/oidc/filter/OidcRequestFilter.java
```diff
@@ -41,12 +45,59 @@ public OidcRequestFilter(
...
+   * retrieve a token.
+   *
+   * <p>Validation is skipped when {@link OidcConfiguration#isStartupValidationDisabled()} returns
+   * {@code true}. The constructors also skip this validation when OIDC is disabled.
+   *
+   * @throws IllegalStateException if startup validation is enabled and the OIDC client is not
+   *     initialized, the token endpoint cannot be reached, or no valid access token can be
+   *     retrieved
+   */
+  public void validateOidcConfig() {
```
JoergSiebahn: Does this method need to be public? I only can see it's called within the class. If it's private we don't need to keep the configuration as a member and we would not have another method that is bound to SemVer rules regarding breaking changes.

sda-commons-client-jersey/src/main/java/org/sdase/commons/client/jersey/oidc/filter/OidcRequestFilter.java
```diff
@@ -41,12 +45,59 @@ public OidcRequestFilter(
...
+  }
+
+  private String buildOidcConfigValidationErrorMessage(String detail) {
+    return "OIDC startup validation failed. "
+        + detail
+        + " Configuration: issuerUrl='"
+        + oidcConfiguration.getIssuerUrl()
+        + "', grantType='"
+        + oidcConfiguration.getGrantType()
+        + "'. Check OIDC provider settings.";
```
JoergSiebahn: NIT: This may use String format:


```suggestion
    return "OIDC startup validation failed. %s Configuration: issuerUrl='%s', grantType='%s'. Check OIDC provider settings."
        .format(detail, oidcConfiguration.getIssuerUrl(), oidcConfiguration.getGrantType());
```
(The suggestion may not pass the code format requirement)

sda-commons-client-jersey/src/main/java/org/sdase/commons/client/jersey/oidc/filter/OidcRequestFilter.java
```diff
@@ -41,12 +45,59 @@ public OidcRequestFilter(
...
+
+    OidcResult oidcResult;
+    try {
+      oidcResult = oidcClient.createAccessToken();
+    } catch (Exception e) {
+      throw new IllegalStateException(
+          buildOidcConfigValidationErrorMessage("Token retrieval failed."), e);
+    }
+
+    if (oidcResult == null || oidcResult.getState() != OidcState.OK) {
```
JoergSiebahn: Shouldn't this check only for `ERROR`? The state `SKIPPED` indicates that OIDC is disabled. That should be a valid case for validation.


```suggestion
    if (oidcResult == null || oidcResult.getState() == OidcState.ERROR) {
```

JoergSiebahn (commented):
Hi @sabkuehn,

thanks for the contribution. It makes sense to have the option to validate the token retrieval on startup. There are some minor things I'm struggling with. Could you have a look at the comments below?

docs/client-jersey.md
sabkuehn: Sounds good. Changed to OIDC_ENABLE_STARTUP_VALIDATION.

sda-commons-client-jersey/src/main/java/org/sdase/commons/client/jersey/oidc/filter/OidcRequestFilter.java
sabkuehn: updated

sda-commons-client-jersey/src/main/java/org/sdase/commons/client/jersey/oidc/OidcConfiguration.java
sabkuehn: updated

docs/client-jersey.md
sabkuehn: Added

sda-commons-client-jersey/src/main/java/org/sdase/commons/client/jersey/oidc/filter/OidcRequestFilter.java
sabkuehn: changed to private.

sda-commons-client-jersey/src/main/java/org/sdase/commons/client/jersey/oidc/filter/OidcRequestFilter.java
sabkuehn: updated

sda-commons-client-jersey/src/main/java/org/sdase/commons/client/jersey/oidc/filter/OidcRequestFilter.java
sabkuehn: updated

JoergSiebahn (approved):
Thank you, @sabkuehn

PR #4501, opened 2026-07-10
Prepare module downsizing for next major version

JoergSiebahn:
Due to announced [publishing limits for Maven Central](https://central.sonatype.org/publish/maven-central-publishing-limits/) we need to reduce the affected metrics to continue reasonable release cycles and be able to publish security fixes in time.

The most critical metric at the moment is the amount of published files. Each module consumes 16 files per release.

This PR prepares deprecation of modules so they can be removed with the next major update. The goal is to move functionality from small modules into modules that are required anyways. In most (maybe all) cases, classes will be migrated to sda-commons-server-dropwizard. This PR aims for backward compatible changes, but there may be edge cases that are not covered without migrating away from deprecated modules.

The migration to prepare for the next major release of SDA Dropwizard Commons should require to change imports and remove the dependency to the affected modules only.

All tasks beyond removing the deprecated classes and modules are marked with `TODO next-major`.

The following modules are deprecated and their functionality is available in sda-commons-server-dropwizard:

- [x] sda-commons-shared-tracing
- [x] sda-commons-shared-error (to be discussed, before this change shared-error offered the error classes without DW dependency, this may be beneficial/required for pure API libraries (which we don't recommend!))
- [x] sda-commons-server-consumer
- [x] sda-commons-server-trace
- [x] sda-commons-server-cors
- [x] sda-commons-server-healthcheck
- [ ] sda-commons-server-security
- [ ] sda-commons-server-jackson (maybe)
- [ ] sda-commons-server-auth (maybe)
- [ ] sda-commons-server-auth-testing (maybe move to sda-commons-shared-wiremock-testing)
- [ ] sda-commons-shared-certificates (if we can do the parsing without Bouncycastle and can drop the dependency)
- [ ] sda-commons-server-prometheus (maybe, comes with a lot of dependencies)
- [ ] sda-commons-shared-yaml (snakeyaml comes with dropwizard-configuration anyways, but sda-commons-shared-asyncapi would depend on sda-commons-server-dropwizard instead of only shared-yaml)


YellowFlora:
It should be possible to replace Bouncycastle in shared certificates with [Java CertificateFactory](https://docs.oracle.com/javase/8/docs/api/java/security/cert/CertificateFactory.html)

YellowFlora:
dropwizard_modules drawio

PR #4533, opened 2026-07-27
Final dependency updates and CVE mitigation for Release 8.x.x

JoergSiebahn:
After merge:
- [ ] Update release notes to clarify this is the last release for SDA Dropwizard Commons Version 8.
- [ ] Make branch release/8.x.x readonly


YellowFlora (commented):
I think we should avoid the micrometer update 
The 1.17.0 update had a slight behavior change. Nothing major, but we said we only provide security updates and that may be a problem in rare cases https://github.com/micrometer-metrics/micrometer/wiki/1.17-Migration-Guide

YellowFlora (approved):

PR #4534, opened 2026-07-28
Avoid duplicate embed parameters and inherit field filter configuration

sabkuehn:
## Problem

`EmbedParameterModifier` adds an `embed` query parameter even when one already exists, resulting in duplicate parameters in the generated OpenAPI document.

## Change

Check for an existing query parameter with `name = embed` before adding the generated parameter.

## Result

The modifier becomes idempotent and preserves explicitly configured parameter settings.

## Problem

`@EnableFieldFilter` is not inherited by subclasses. Abstract base resource classes must therefore repeat the annotation on every concrete implementation.

## Change

Add `@Inherited` to `EnableFieldFilter`.

## Result

Concrete subclasses automatically use the field-filter configuration declared on their base class and may still override it locally.

### DIFF
--- a/sda-commons-server-openapi/src/main/java/org/sdase/commons/optional/server/openapi/parameter/embed/EmbedParameterModifier.java
+++ b/sda-commons-server-openapi/src/main/java/org/sdase/commons/optional/server/openapi/parameter/embed/EmbedParameterModifier.java
@@ -7,24 +7,24 @@ import io.swagger.v3.oas.annotations.OpenAPIDefinition;
 import io.swagger.v3.oas.integration.api.OpenApiReader;
 import io.swagger.v3.oas.models.OpenAPI;
 import io.swagger.v3.oas.models.Operation;
-import io.swagger.v3.oas.models.media.ArraySchema;
-import io.swagger.v3.oas.models.media.ComposedSchema;
-import io.swagger.v3.oas.models.media.MediaType;
-import io.swagger.v3.oas.models.media.Schema;
-import io.swagger.v3.oas.models.media.StringSchema;
+import io.swagger.v3.oas.models.media.*;
 import io.swagger.v3.oas.models.parameters.Parameter;
 import io.swagger.v3.oas.models.parameters.QueryParameter;
-import java.util.HashMap;
-import java.util.List;
-import java.util.Map;
-import java.util.Objects;
+import java.util.*;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 /** Adds the embeddable resources as query parameter, so they can be selected in the swagger ui. */
 @OpenAPIDefinition
 @SuppressWarnings({"java:S3740", "rawtypes"})
 // ignore "Raw types should not be used" introduced by swagger-core
 public class EmbedParameterModifier implements ReaderListener {
+
+  private static final Logger LOGGER = LoggerFactory.getLogger(EmbedParameterModifier.class);
+
   private static final String EMBEDDED_PROPERTY = "_embedded";
+  private static final String EMBED_PARAMETER_NAME = "embed";
+  private static final String QUERY_PARAMETER_LOCATION = "query";
 
   @Override
   public void beforeScan(OpenApiReader reader, OpenAPI openAPI) {
@@ -80,8 +80,7 @@ public class EmbedParameterModifier implements ReaderListener {
               if (embedQueryParameter == null) {
                 return;
               }
-
-              operation.addParametersItem(embedQueryParameter);
+              addEmbedParameter(operation, embedQueryParameter);
             });
   }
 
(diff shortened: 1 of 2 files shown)


YellowFlora (commented):
Hello @sabkuehn 

Thanks a lot for the PR. 

I have a few notes on this PR.

First regarding the change for `EnableFieldFilter`.

We changed it behavior recently and there I already thought about adding the inherit annotation. But I am hesitant about it. The problem in my eyes is, that it would let to behavior changes downstream.

Let's say you have `Car` and `Plane` both extend `Vehicle`.
Now you have three endpoints, each returning one of these types.
You define that filtering should be enabled for `Vehicle` and `Car` but not for `Plane`. Maybe there are reasons, why it `Planes` should not get filtered.

Adding the annotation would lead to a change that planes are automatically filterable now too. 
Additionally it would go against the explicit opt in principle of the functionality we hold so far.

Second regarding `EmbedParameterModifier`. 

I'm worry a bit about what will happen when someone defines their own `embed` query parameter.  
Let's say someone defines a boolean `embed` parameter. How will the application deal with it when the embed functionality is enabled?
The current behavior is an indicator at least for the developer to double check the property and may change the name to something else to not conflict with this functionality.

sabkuehn:
Thanks for the detailed feedback.

Regarding `EnableFieldFilter`, I agree that adding `@Inherited` changes the current explicit opt-in semantics. A subtype could become filterable unintentionally just because its base class is annotated. I will remove this change from the PR. Concrete resource types should continue to opt in explicitly.

Regarding `EmbedParameterModifier`, the intention was specifically to avoid duplicate parameters when an imported OpenAPI specification already contains the generated `embed` parameter.

Your boolean parameter example is a valid concern. An existing parameter with the same name should not always be accepted silently. I suggest handling the cases as follows:

* If no `embed` query parameter exists, generate it as before.
* If a compatible `embed` parameter already exists, keep the existing definition.
* If an incompatible parameter exists, such as a boolean `embed` parameter, fail with a clear error explaining the conflict.

This would preserve explicitly configured properties such as `style` and `explode`, avoid producing an invalid OpenAPI document with duplicate parameters, and still make naming conflicts visible to the developer.

I adjusted the PR and added tests for both compatible and incompatible existing parameters.

sda-commons-server-openapi/src/main/java/org/sdase/commons/optional/server/openapi/parameter/embed/EmbedParameterModifier.java
```diff
@@ -93,6 +87,52 @@ String getOriginalRef(Schema<?> schema) {
...
+
+    if (operation.getParameters() != null) {
+      for (Parameter parameter : operation.getParameters()) {
+        if (parameter == null || !isEmbedQueryParameter(parameter)) {
+          continue;
+        }
+
+        if (existingEmbedParameter != null) {
+          throw new IllegalStateException(
+              "The operation contains multiple 'embed' query parameters.");
```
YellowFlora:
```suggestion
              "The operation contains multiple 'embed' query parameters already.");
```

sda-commons-server-openapi/src/main/java/org/sdase/commons/optional/server/openapi/parameter/embed/EmbedParameterModifier.java
```diff
@@ -93,6 +87,52 @@ String getOriginalRef(Schema<?> schema) {
...
+    if (existingEmbedParameter == null) {
+      operation.addParametersItem(generatedParameter);
+      return;
+    }
+
+    if (!isCompatibleEmbedParameter(existingEmbedParameter)) {
+      throw new IllegalStateException(
+          "The query parameter 'embed' conflicts with the generated embed parameter. "
+              + "It must be an array of strings or use a different name.");
+    }
```
YellowFlora, line 122: Could you add an info log message, which displays, that a compatible embed element was found and adding another one was skipped?

sda-commons-server-openapi/src/main/java/org/sdase/commons/optional/server/openapi/parameter/embed/EmbedParameterModifier.java
```diff
@@ -93,6 +87,52 @@ String getOriginalRef(Schema<?> schema) {
...
+  private void addEmbedParameter(Operation operation, Parameter generatedParameter) {
+    Parameter existingEmbedParameter = null;
+
+    if (operation.getParameters() != null) {
+      for (Parameter parameter : operation.getParameters()) {
+        if (parameter == null || !isEmbedQueryParameter(parameter)) {
+          continue;
+        }
+
+        if (existingEmbedParameter != null) {
```
YellowFlora, line 104: Could you add a test for this scenario too?

YellowFlora (commented):
Hello @sabkuehn 

I go along with your counterproposal. I quite like that idea. 
So the user of `EmbedParameterModifier` knows exactly, when something will not work correctly.

I only found a few minor things, the rest looks good to me. 

It could also be thought about using EMBED_PARAMETER_NAME in the exception messages, to ensure everywhere the same parameter name is referenced, but I don't see it as necessary.

sda-commons-server-openapi/src/main/java/org/sdase/commons/optional/server/openapi/parameter/embed/EmbedParameterModifier.java
sabkuehn: shouldKeepCompatibleExistingEmbedParameter already covers this.

sda-commons-server-openapi/src/main/java/org/sdase/commons/optional/server/openapi/parameter/embed/EmbedParameterModifier.java
sabkuehn: added

sda-commons-server-openapi/src/main/java/org/sdase/commons/optional/server/openapi/parameter/embed/EmbedParameterModifier.java
sabkuehn: added

sda-commons-server-openapi/src/main/java/org/sdase/commons/optional/server/openapi/parameter/embed/EmbedParameterModifier.java
YellowFlora: `shouldKeepCompatibleExistingEmbedParameter` doesn't cover the throwing of `new IllegalStateException(
              "The operation contains multiple 'embed' query parameters already.");`

sabkuehn: added shouldRejectMultipleExistingEmbedParameters

YellowFlora (approved):

YellowFlora:
Just as short note. I'm discussing the Sonar check internally right now how we will handle it.

PR #4561, opened 2026-08-19
fix: remove stale entity transport headers for decompressed responses

scriam:


sda-commons-client-jersey/src/test/java/org/sdase/commons/client/jersey/ApiClientTest.java
```diff
@@ -322,6 +325,33 @@ void addReceivedTraceTokenToHeadersToPlatformCall() {
...
+                    .withStatus(200)
+                    .withHeader(HttpHeaders.CONTENT_TYPE, MediaType.APPLICATION_JSON)
+                    .withHeader(HttpHeaders.CONTENT_ENCODING, "gzip")
+                    .withHeader(HttpHeaders.CONTENT_LENGTH, String.valueOf(compressedBody.length))
+                    .withBody(compressedBody)));
+
+    try (Response response =
+        dwClient().path("api").path("cars").request(MediaType.APPLICATION_JSON_TYPE).get()) {
+      assertThat(response.getStatus()).isEqualTo(200);
+      assertThat(response.getHeaderString(HttpHeaders.CONTENT_ENCODING)).isNull();
```
YellowFlora, line 345: Must: could you add another check that the content-length header was also removed?

sda-commons-client-jersey/src/main/java/org/sdase/commons/client/jersey/filter/CompressedResponseHeaderFilter.java
```diff
@@ -0,0 +1,43 @@
...
+import jakarta.ws.rs.client.ClientResponseFilter;
+import java.util.Arrays;
+import java.util.Locale;
+import java.util.Set;
+
+public final class CompressedResponseHeaderFilter implements ClientResponseFilter {
+
+  private static final String CONTENT_MD5 = "Content-MD5";
+
+  private static final Set<String> COMPRESSED_ENCODINGS = Set.of("br", "deflate", "gzip", "x-gzip");
```
YellowFlora: NTH: Should zstd be included here as well? 
since 5.6 httpclient also supports zstd 

maybe there is a way to get the list directly from the lib?

sda-commons-client-jersey/src/main/java/org/sdase/commons/client/jersey/filter/CompressedResponseHeaderFilter.java
```diff
@@ -0,0 +1,43 @@
...
+
+  private static final Set<String> ENTITY_TRANSPORT_HEADERS =
+      Set.of(CONTENT_ENCODING, CONTENT_LENGTH, CONTENT_MD5);
+
+  @Override
+  public void filter(ClientRequestContext requestContext, ClientResponseContext responseContext) {
+    String contentEncoding = responseContext.getHeaderString(CONTENT_ENCODING);
```
YellowFlora, line 31: NTH: Could you add a comment here explaining that this is a workaround for dropwizard/dropwizard#11253?

sda-commons-client-jersey/src/test/java/org/sdase/commons/client/jersey/ApiClientTest.java
scriam: this is tested here https://github.com/SDA-SE/sda-dropwizard-commons/pull/4561/changes#diff-a6d128b065c9b8f886361cabba0cbbb99704b62699fcca0e68962872443a46e6R21. 

In this case we cannot assert that the content-length compressed header is null, because at this point in return there is a content-length that already corresponds to the decompressed json body.

sda-commons-client-jersey/src/main/java/org/sdase/commons/client/jersey/filter/CompressedResponseHeaderFilter.java
scriam: done

sda-commons-client-jersey/src/main/java/org/sdase/commons/client/jersey/filter/CompressedResponseHeaderFilter.java
scriam: done

sda-commons-client-jersey/src/test/java/org/sdase/commons/client/jersey/ApiClientTest.java
YellowFlora: Right, i overlooked that

YellowFlora (approved):

